PR #4, opened 2015-03-02
docs moved from manifests wiki

ozzyjohnson:
Some docs which I added to the manifests wiki are a better fit here.


dlapiduz:
Thanks @ozzyjohnson !

PR #9, opened 2015-03-09
create a working self-contained BOSH / CF on AWS

ozzyjohnson:
This doc describes the process to create a working CF environment within a single instances on AWS via Vagrant and recover it after restart.  I put this together while creating my own environment to test logging options and the services-contrib collection.


dlapiduz:
@ozzyjohnson most of these instructions also apply for a local copy of CF, right?

ozzyjohnson:
@dlapiduz - Yes, the vast majority would apply to a fully local environment as well.  I'm thinking that if I were to put together a guide to going fully local with virtualbox etc. I would write it up in a very similar, but separate doc.

PR #21, opened 2015-03-18
Add Updating CF doc

dlapiduz:


content/ops/updating-cf.md
```diff
@@ -0,0 +1,18 @@
...
+menu:
+  main:
+    parent: ops
+title: Updating Cloud Foundry
+weight: 10
+---
+
+Note: Grab something to eat and drink, this might take a while and an update shouldn't be left unattended.
+
+1. SSH to jumpbox
```
afeld, line 11: How would someone find out what the jumpbox user/IP is?

dlapiduz, line 11: Do you know the secret handshake?

Will add how to find it...

dlapiduz, line 11: Actually, we need to add like an "AWS onbaording for Cloud Foundry operators" doc. Because the only way to SSH to the jumpbox is if you have the key to it...

afeld:
You would know the most about this, but looks fine to me :+1: 
- Are there general CF docs around this that can be linked to?
- Seems like a lot of this could be scriptable, though not a blocker.

dlapiduz:
@afeld unfortunately there aren't that many good docs about it. I should contribute some.

afeld:
Ideally we will move as much of this repo as possible to http://docs.cloudfoundry.org/. That's sort of a long-term meta-task, I guess.

dlapiduz:
Actually I think that we need to strike a balance. We should keep here a small set of docs that are pertinent for the developers day to day operations with links for deeper knowledge or edge cases.

dlapiduz:
One very important thing is that the CF community is not there to provide support for our users. We need to do that.

dlapiduz:
@afeld @ozzyjohnson can we merge this?

PR #22, opened 2015-03-18
simple user creation script

ozzyjohnson:
Something quick to simplify new user requests.


dlapiduz:
@ozzyjohnson this is really cool!

I wonder if that should be a CLI plugin though. Did you think about that?

ozzyjohnson:
@dlapiduz - Yeah, definitely. I have several ideas about plugins and tools, but I can write shell quickly - without having to think about it.

afeld:
Merge early, iterate often!

PR #34, opened 2015-03-26
most organizations should only need a single space

afeld:
Switching between spaces is a bit of a hassle, and for our use case, I'm realizing that there's not really any benefit to splitting environments between them. Applications with multiple environments will have the environment name in the CF app name anyway (e.g. `c2-staging`).

### DIFF
--- a/content/getting-started/index.md
+++ b/content/getting-started/index.md
@@ -92,13 +92,9 @@ cf target -o ORGNAME
 
 Every application is scoped to a "space". Applications in the same space share a location for app development, deployment, and maintenance.
 
-### Naming convention
-
-For simple organizations where there's only going to be a single application per environment, just use a single space called `general`. If the org is more general (e.g. `devops`), the space may correspond to the name of the project (e.g. `hubot`).
-
 ### Management
 
-To create a space:
+For most organizations, having a single space called `general` will work. To create a space:
 
 ```bash
 cf create-space SPACENAME
--- a/content/getting-started/index.md
+++ b/content/getting-started/index.md
@@ -94,7 +94,7 @@ Every application is scoped to a "space". Applications in the same space share a
 
 ### Naming convention
 
-Within 18F, the `SPACENAME` will correspond to an environment, e.g. `dev` or `prod`. If the org is more general (e.g. `devops`), the space may correspond to the name of the project (e.g. `hubot`).
+For simple organizations where there's only going to be a single application per environment, just use a single space called `general`. If the org is more general (e.g. `devops`), the space may correspond to the name of the project (e.g. `hubot`).
 
 ### Management
 


ozzyjohnson:
I'm fine with this. But I'm thinking it would make sense to add something about the tradeoffs. 

If `c2-staging` and `c2-prod` are both in a single space with services `c2-prod-db` and `c2-staging-db` respectively, `c2-staging` can bind to the `c2-prod-db`. If they're in separate spaces, they have no visibility to each others services.

Also, quotas can now be applied to spaces. Potentially valuable to ensure that `*-prod` always has enough room in the overall org quota to launch or relaunch.

dlapiduz:
I honestly can't endorse this. I think that spaces exist for a reason and I know it is not ideal how we switch from one to the other but _hopefully_ you wouldn't have to do it so much...

Spaces provide very good logical separation and as @ozzyjohnson says you can split services, quotas and permissions.

Ideally a dev would only push to a `dev` space and `staging` and `production` would be automated. I understand this is not the case yet but eventually we will get there.

dlapiduz:
This is a first but :-1:

afeld:
> If they're in separate spaces, they have no visibility to each others services.

If we're using RDS instances in a single VPC, they can all theoretically connect to each others' databases anyway, so I don't see this as being any worse.

> quotas can now be applied to spaces. Potentially valuable to ensure that `*-prod` always has enough room in the overall org quota to launch or relaunch.

Do you think that's something we should be doing? That's the only valuable thing (to us) I'm hearing in favor of keeping each environment in a separate space.

> it is not ideal how we switch from one to the other

Just as background, this idea came up because I accidentally created the three environments for CAP in the same space...and I (almost) know what I'm doing. From what I can tell, there isn't an easy way to transfer an app from one space to another, other than deleting and re-pushing it. If the same app were called the same thing in each space, it would be _very_ easy to accidentally run `cf` commands against the wrong space.

> _hopefully_ you wouldn't have to do it so much

I'm approaching the documentation on this site as the best/simplest practices we have for our current setup. If in the future we have a custom CLI or a web interface to manage these things, it may be much easier to deal with multiple spaces, at which time I would be totally on board with the multi-space recommendation.

> This is a first but :-1:

If you both still feel that way, happy to close the issue. Just wanted to make sure we thought it all the way through.

Thanks!

dlapiduz:
@afeld to be honest I have pushed apps to the wrong space (and org) more than once. I don't think that saying "lets use only one space" is the solution though... 

I think @kaitlin had a good idea with a kind of ".ruby-version" solution... but I still think we shouldn't _recommend_ having only one space.

Also, at the end of the day, each org can do whatever they want.

kaitlin:
I don't think the title of this issue makes sense. I have one organization for our client (OASIS) and multiple spaces for the different projects we're building for that client. I find the spaces useful in that context.

dlapiduz:
Keeping the branch around to see if we want to revisit this.

PR #36, opened 2015-03-26
add instructions around running Rails migrations

afeld:


ozzyjohnson:
@afeld - What do you think about the options described here:

http://docs.cloudfoundry.org/devguide/services/migrate-db.html

Particularly the third option [migrate frequently](http://docs.cloudfoundry.org/devguide/services/migrate-db.html#frequent-migration).

What I don't like with each of those options is baking things into the `command:` key. Not only does that seem like something that might end deprecated, but I've seen inconsistent or at least confusing behavior from the combination of possibilities presented by `-c`, `command:` and the buildpack default start command.

I'd think a variation on that last option would be ideal. Baking some logic into the start-up of the app that checks and environment variable and runs the migration.

I like this because:
- It keeps all of the app function observable in source control.
- It keeps the toggle observable in the `cf env`.

An automated, observable process with a manual latch as opposed to a fully manual process with no trail.

cmc333333:
I agree with @ozzyjohnson, every deploy should include a migration step. The automation nicely removes user error from the equation. We handled this on peacecorps by assigning one node to play a "mutex leader"; it was the only one which could update the db. This wasn't ideal, as we needed to assign that role ahead of time.

The `VCAP_APPLICATION` solution is better. I'm not a fan of shoving it into a ruby script (rather than that being built into cloud foundry) but meh.

afeld:
A bit torn about migrations happening automatically. On the one hand, having the database schema keep up with deployed changes the code makes perfect sense, and takes away the room for user error. That being said, I've already been responsible for a few database migration failures at 18F, and if those happened without me watching them run and poking at the data after (e.g. an app with continuous deployment + migration), it would have been worse.

## Preaching to the choir

In my experience, migrations are one of the most fragile parts of codebases. Reasons include:
- They aren't generally covered by automated tests.
- Even if they're tested manually on a snapshot of production data, new data changes could cause the migration to fail anyway.
- They don't usually include any verification step.
- Unless the migration runs within a single transaction, they are hard to roll back.
  - ["There are certain queries [in Rails migrations] that can't run inside a transaction."](http://edgeguides.rubyonrails.org/active_record_migrations.html#migration-overview)

I think most of these issues can be mitigated with tooling and process (we should talk through ideas in separate thread(s)), but without strict adherence to those, I would be very nervous to have a migration run on a production DB without someone babysitting it. I don't know of a single organization that has really "figured this out", though I would love to see what that looks like.

afeld:
TL;DR I still prefer manually-triggered migrations.

arowla:
@afeld I agree wholeheartedly that manually-triggered migrations are the gold standard for production sites. I have watched nervously as the world switched to scripted migrations led by developers, rather than hand-coded SQL, written by developers, but looked over and applied by DBA's. The more abstraction we get between us and the DB, the more muddy the thinking becomes about how DB changes are going to affect a live site. There is a tendency to try to do too many things at once, as opposed to doing things step-wise, additive first, (then code deployed), subtractive/mutative last. Manually applied data migrations also give one the option to do some queries within a transaction to make sure things look OK (manual verification) before committing the transaction. That can be huge. Scripting migrations with an ORM also leads to weird chicken-and-egg problems with models, etc.

That said, there must be some sort of middle-ground. Migrations are baked in to all the modern web frameworks, so they must be working for someone. (Right? Right?? The cynical voice says that the push toward scripted/ORM migrations comes from the NoSQL/LowSQL crowd who just doesn't want to learn how to use a SQL database.) I am not sure what the best solution is here-- I have actually never worked on a production site with scripted migrations, only manual ones. But I would love to have a conversation about it. The best I can suggest right now is this:
- Every mission-critical production site needs a staging site with an up to date snapshot of production data. Migrations would be applied there first.
- The staging site, during the migration test, _must_ be under load from a site spider to catch whether transient errors from table locking, etc. during a migration are significant enough to warrant a more careful redesign of the migration.
- There must be a review process for migrations involving people highly experienced in RDBMS's, to help plan production changes effectively.

I am happy to help with all these points, but especially the last one, were we to go that route. I'm actually going to propose here that we spin up a Databases Working Group to hash some of this stuff out.

cmc333333:
These seem more like arguments for manual deployments than for including scripted migrations, no? We still want control of the big red button in case anything goes wrong. Requiring someone type `somecommand migrate` after updating code doesn't give us any additional benefits, aside from being ssh'd in. Instead, requiring manual steps makes it much easier for the developing team to get out of sync.

I agree whole heartedly about keeping an up-to-date staging site to verify before pushing to production.

Requiring a "people highly experienced in RDBMS's" seems a step too far (and is slightly insulting). Most ORMs remove the error surface area significantly, particularly for most common types of migrations. Further, even if we assumed that the migration framework only allowed raw SQL, every web dev should have enough experience to perform this themselves. We should all know how to order migrations so that data isn't lost. Honestly, the only migration issues I've encountered have been with rails, which linearizes migration order by timestamp (rather than matching the fork-merge model), and which doesn't encourage data validation at the db level. If you don't feel confident in your work, you should ask for help. Adding more layers of review will just bog development down, and I highly doubt would catch a significant number of errors.

cmc333333:
Also, to make an analogy, running the library-updating step could also fail (e.g. https://github.com/18F/peacecorps-site/pull/355). Requiring deployers run this manually would introduce lots of version inconsistency errors, though, so we include it in deployment scripts.

jgrevich:
The concept of  ['Hot Compatibility'](http://pedro.herokuapp.com/past/2011/7/13/rails_migrations_with_no_downtime/) may be useful in this case. It usually comes up when dealing with [zero-downtime deploys](http://blog.honeybadger.io/zero-downtime-migrations-of-large-databases-using-rails-postgres-and-redis/) but I think it also applies to many of the problems cited above.

> So here's the basic principle that allows you to avoid downtime: any migration being deployed should be compatible with the code that is already running.
> 
> In order to do so, you'll usually split your deploy process in two steps:
> 1. Make the code compatible with the migration you need to run
> 2. Run the migration, and remove any code written specifically for it
> 
> Going back to our example: if you want to remove a column, you'll need to deploy a patch telling ActiveRecord to ignore it first. Only then you can deploy the migration, and clean up that patch.

yozlet:
I'm also interested in Hot Compatibility. Does anyone know of an easy way to auto-test new migrations in CI?

jessieay:
In the Rails world, it's pretty typical for production applications to have a deploy script that runs migrations automatically. 

I agree with @cmc333333 that having to `ssh` in to run migrations isn't making things any more foolproof than have a deploy script that runs them right away. In addition, having migrations run automatically reduces downtime because there isn't a lag between deploy and `ssh` time. 

The points that @arowla brings up around the challenges of scripted migrations are interesting, but given that we don't have any DBAs (as far as I know!) here, it doesn't seem like we should move in the direction of manually triggered migrations administered by DBAs.

Agreed that we should be testing deploys against a staging that has prod data, though. Great idea!

sharms:
Looking to make a choice one way or the other to clean out the pull queue - after reviewing the thread:

If the app is not deployed via CI, and has no plans / resources for CI or staffing issues, deploying unscripted make sense.

If the app is deployed via CI then migrate often should be the default, as a human cannot reasonable be 'continuous'.

I feel like an easily solvable bridge would be to make a doc which says 'Here is a link to deploy continuous' w/ example, and then a 'Troubleshooting' header which includes @afeld steps.  If I don't hear back next week I will submit the pull for it.

dlapiduz:
@afeld should we close this?

afeld:
Yeah, definitely not the recommendation anymore.

PR #38, opened 2015-03-31
Change wercker setup to support new version of buildpack

dlapiduz:
There is a collision with the "public" folder naming...


dlapiduz:
@afeld thanks for fixing this...

PR #47, opened 2015-04-22
Remove warning about SSL validation

konklone:
This is fixed now that we're on `api.18f.gov` right?


afeld:
Anyone have a repo to try it with?

dlapiduz:
We should test it but I am going ahead with the merge since this _should_ work.

PR #49, opened 2015-04-29
Document sudo requirement for continuous deployment on Travis

konklone:
On Travis, you need to add `sudo: required` for the build to work.


dlapiduz:
That is interesting, thanks!

PR #51, opened 2015-05-04
add guide for moving apps

afeld:
I've needed to do this a few times...a couple things I realized:
- I don't understand why there's a difference between a "domain" and a "(non-shared) route" in CF.
- This is waaaay too many steps, and we should script it at some point.


dlapiduz:
Cool! thanks!

PR #60, opened 2015-05-27
Multiple instances

geramirez:
Adding draft documentation for working with multiple instances. Feel free to make changes.

### DIFF
--- a/content/ops/multiple-instances.md
+++ b/content/ops/multiple-instances.md
@@ -1,0 +1,54 @@
+---
+date: 2015-05-27T13:32:00-04:00
+menu:
+  main:
+    parent: ops
+title: Running Multiple Instances
+weight: 10
+---
+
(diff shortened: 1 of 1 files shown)


dlapiduz:
I would also add a note about `cf scale`. What do you think?

geramirez:
@dlapiduz good point. I just learned something new.

content/ops/multiple-instances.md
```diff
@@ -0,0 +1,54 @@
...
+The number of instances can also be defined using the `instances` key in the `manifest.yml`.
+```yaml
+memory: 512mb
+instances: 2
+applications:
+  - name: <<APP_NAME>>
+    path: .
+```
+
+### Scale Command Method
```
dlapiduz: Should this be `####` instead of `###`?

PR #64, opened 2015-06-03
fix broken link and add sandbox note

ultrasaurus:
just getting started learning CF myself
thought this would have been good to start with


afeld:
w00t!

PR #68, opened 2015-06-15
update link to staticfile buildpack

khandelwal:
This updates the link in the manifest.yml example to the new static buildpack link (note that it has graduated from the incubator).


afeld:
Thanks!

PR #71, opened 2015-07-09
Add a tip on continuously deploying to certain apps

theresaanna:
I just set up continuous deployment for CALC's dev server and, after trying the (http://docs.travis-ci.com/user/deployment/#Conditional-Releases-with-on%3A)[on setting] to accomplish this, I ended up reconfiguring the apps. The one I want to deploy to is in its own space now, making configuration easy.

It took me awhile to get this working properly, so I'm hoping to spare others the effort.

### DIFF
--- a/content/apps/continuous-deployment.md
+++ b/content/apps/continuous-deployment.md
@@ -43,6 +43,8 @@ sudo: required
 
 Replace `DEPLOYER_USER`, `ORG`, and `SPACE` accordingly and run `travis encrypt --add deploy.password --skip-version-check` to enter the deployer's password.
 
+If you want to have continuous deployment on only a specific app or group of apps, the easiest solution is to isolate them in their own space.
+
 #### Jekyll Site
 
 Deploying a Jekyll site requires a few changes to your `.travis.yml` and `manifest.yml` as well as the addition of a `Staticfile` and `Gemfile`.


dlapiduz:
@theresaanna I am interested in knowing why one space is easier than many? Is it because travis does a plain `cf push`?

theresaanna:
@dlapiduz I'm not sure what Travis does under the surface. What I can tell you is that it will deploy to each app in the space that you specify. I spent awhile trying to make a custom condition work with an env var, but I had no luck. See http://docs.travis-ci.com/user/deployment/#Conditional-Releases-with-on%3A

Ideally, you could specify the app in `.travis.yml` like:

```
deploy:
  organization: org
  space: space
  app: app
```

If I were a Rubyist, I'd attempt to write that feature, but I am not. https://github.com/travis-ci/dpl

If it helps, here's a CALC Travis build that deploys: https://travis-ci.org/18F/calc/builds/70287251#L368
(Ignore any mention of the `calc-prod` app. That shouldn't have existed in the target space.)

dlapiduz:
@theresaanna to be completely honest, I am not sure this is good advice. You shouldn't have to have only one app per space to do CD. Travis support for CF is still in the works and right now it only does `cf push` basically pushing all apps in your `manifest.yml` file.

I understand that having only one app per space works for you right now but I don't want to encourage other people to do the same.

What I do recommend is that, if the travis integration doesn't work as expected, you can create your own shell script. We have a couple examples going around and we can help with that...

dlapiduz:
@ozzyjohnson @afeld maybe you guys can chime in...

ozzyjohnson:
I'm thinking there are some mixed signals here. 

As I understand it the complication here was not about the number of apps per space, but the number of apps per manifest - that calc was defining multiple apps in a manifest and travis doesn't expose the ability to specify arguments to `cf push`.

Breaking each app out into it's own manifest while sharing a space should address the travis / app organization issues while slightly complicating the manual workflow (multiple pushes).

Does that make sense?

dlapiduz:
@ozzyjohnson I don't think that Travis allows you to specify which manifest file to use... only manifest.yml works.

ozzyjohnson:
@dlapiduz - Right, I should have been specific. Each app would have its own manifest at the root of its own repo / branch.

dlapiduz:
@ozzyjohnson that makes sense. One manifest per app works...

Issue #69, opened 2015-06-17
revisit recommendation around shared vs. independent databases


afeld:
- https://docs.18f.gov/apps/databases/
- https://18f.slack.com/archives/devops/p1434553182000569

afeld:
Basically that whole page needs updating.

PR #84, opened 2015-08-05
Describe the rds broker for creating DBs

cmc333333:
Per conversation with @ozzyjohnson and referencing comments from @dlapiduz


afeld:
Closes #69.

PR #85, opened 2015-08-05
expand information around database services

afeld:
Had been re-writing this page concurrently with @cmc333333 in #84 :blush: Rebased on top of `master`.

### DIFF
--- a/content/apps/databases.md
+++ b/content/apps/databases.md
@@ -6,21 +6,33 @@ title: Databases
 weight: 10
 ---
 
-The current recommendation is to use our RDS service broker to spin up and manage a database for you. If you need more control, however, you can use AWS-managed resources directly.
+There are multiple options for how to set up databases for use with applications in Cloud Foundry.
 
-## Service Broker
+## Services
+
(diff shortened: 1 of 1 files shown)


content/apps/databases.md
```diff
@@ -6,22 +6,34 @@ title: Databases
...
 
 ```bash
-cf create-service rds <PLAN> <DB_NAME>
+cf create-service <SERVICE> <PLAN> <DB_NAME>
 cf bind-service <APP_NAME> <DB_NAME>
 cf restage <APP_NAME>
 ```
 
-Where `PLAN` is one of `shared-psql`, `micro-psql`, or `medium-psql`. The shared instance is preferred as an existing, shared database will be used. The other two options will take longer to spin up, but may be required by your application.
+If you want to use PostgreSQL, use the `rds` service rather than `postgresql*`. This gives you things like database snapshots automatically. The `shared-psql` plan(s) uses an existing instance, which is preferred for staging environments. For a production application, use `micro-psql` or `medium-psql`. To use the `rds` service with `myapp` in your staging environment, for example, you would do something like:
```
dlapiduz: I am not sure we should discourage people from using the shared service for production stuff.... If it is just a simple open data app it can totally go in the shared service.

afeld:
Updated!

PR #86, opened 2015-08-07
draft quotas doc

ozzyjohnson:
Draft doc.

@dlapiduz @ramirezg @jcscottiii @mogul - Comments?

### DIFF
--- a/content/apps/quotas.md
+++ b/content/apps/quotas.md
@@ -1,0 +1,123 @@
+---
+menu:
+  main:
+    parent: apps
+title: Quotas
+weight: -10
+---
+
+#### Quotas:
+
+Cloud Foundry capacity is billed by quota. Quotas provide a not to exceed reservation of memory, compute, application routes and service instances. Quotas are associated with and billed to a project IAA number or equivalent billing tag. Quota capacity is a hard limit, unused capacity does not ‘roll over’ month to month.
+
+##### Quota Billing:
+
+Quotas are billed by the unit and tallied daily. A monthly bill might include 9 days at 1 unit, 11 days at 2 units and 10 days at 4 units.
+
+##### Quota Units:
+
+Each unit provides.
+
+- 10 application routes.
+- 3.75GB of RAM.
+- 1 vCPU of compute.
+- 6 service instances.
+
+###### Routes:
+
+Hostname and domain pairs where an application that exposes a listening port can be reached.
+
+###### RAM:
+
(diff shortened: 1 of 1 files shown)


content/apps/quotas.md
```diff
@@ -0,0 +1,123 @@
...
+
+Quotas are billed by the unit and tallied daily. A monthly bill might include 9 days at 1 unit, 11 days at 2 units and 10 days at 4 units.
+
+##### Quota Units:
+
+Each unit provides.
+
+- 10 application routes.
+- 3.75GB of RAM.
+- 1 vCPU of compute.
```
dlapiduz, line 23: I don't think we can guarantee compute power....

content/apps/quotas.md
```diff
@@ -0,0 +1,123 @@
...
+
+##### Quota Billing:
+
+Quotas are billed by the unit and tallied daily. A monthly bill might include 9 days at 1 unit, 11 days at 2 units and 10 days at 4 units.
+
+##### Quota Units:
+
+Each unit provides.
+
+- 10 application routes.
```
dlapiduz, line 21: Is there a reason we limit the amount of routes?

ozzyjohnson, line 21: As I see it, the number of routes should be < ∞ while > whatever the typical org will use. The current default is probably a bit low (perhaps double it), but I don't see a lot to be gained by removing the limit entirely.

content/apps/quotas.md
ozzyjohnson, line 23: Depends.

The RAM reservation extrapolates to a relative CPU share on the DEA. So there is a guarantee in that sense, but how articulate what means is something I deliberated on for a while. If we stick with M-type instances and the same unit size it remains roughly true.

We could refer to CPU in terms of share more like [Heroku](https://devcenter.heroku.com/articles/dyno-types#available-dyno-types).

ozzyjohnson, line 23: Following up after Slack conversations this afternoon. Liking the idea of relating this as a relative share with 1 share equal to 512MB in DEA RAM/CPU and the current quota unit being a bit over 7 shares.

content/apps/quotas.md
dlapiduz, line 21: Ok we can switch it when we actually change the amount of routes... Is it ok if I merge this?

ozzyjohnson, line 21: Good to merge by me.

PR #110, opened 2015-10-06
use the CLI plugin to create new users

afeld:
@jcscottiii Does the plugin have full parity with the script? Haven't actually gotten around to trying it yet.
## After merging
- [ ] Delete [cf-create-user](https://github.com/18F/cloud-foundry-scripts/blob/master/cf-create-user.sh)

### DIFF
--- a/content/ops/create-user.md
+++ b/content/ops/create-user.md
@@ -7,20 +7,8 @@ weight: 10
 ---
 
 ## Creating users
-The helper script [cf-create-user](https://github.com/18F/cloud-foundry-scripts/blob/master/cf-create-user.sh) should always be used to create users so that account creation procedures happen consistently no matter who creates the account.
 
-**Usage:**
-```
-cf-create-user.sh user@domain.tld [org]
-```
-This script will:
-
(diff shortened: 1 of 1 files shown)


dlapiduz:
We should now direct users to do self registration or invitations rather than creating users manually. https://login.18f.gov/create_account is online now

afeld:
I thought that just got locked down?

https://18f.slack.com/archives/cloud-gov/p1444169883000366

Sorry if I'm misunderstanding.

dlapiduz:
Changing directions as we go :) Lets merge this for now

mogul:
objection

The script adds people to the statuspage.io notifications right after it creates them and the plugin doesn't. I was waiting until it did to kill the reference to the script in the docs, but it depends on someone with Go chops (eg not me) to replicate what I did in the script here: 
https://github.com/18F/cloud-foundry-scripts/commit/a8581d33bdf1861ddb416f71f7699369b666ddd4

So by doing this we just directed people to skip adding new accounts to the statuspage notifications. Can we bring it back, or else get someone to update the plugin?

PR #111, opened 2015-10-07
Add section about buildpacks

arowla:
Including the buildpack release schedule that @dlapiduz so aptly described just now

cc @pkarman


pkarman:
lovely. +1

PR #116, opened 2015-10-16
Added -p flag to example make-proxy command

arowla:
... since it seems to be required to get the same results as shown


ozzyjohnson:
Thanks.

PR #118, opened 2015-10-19
Typo fix for instances.

bridgetkromhout:
A setting of `intances: 2` isn't going to produce the desired results. Correcting the typo in this doc.


pkarman:
@bridgetkromhout ! thanks

PR #124, opened 2015-10-26
CF local build guide on Mac

mzia:
A simple guide to setup CloudFoundry on a Mac to match as close to the 18F production/aws build.  

We still need to setup a RDB for any app requiring DB backend.


dlapiduz:
Nice write up!

PR #130, opened 2015-11-18
Durable logging patches

pkarman:
Some doc patches to clarify ELK set up.

This includes a reference to the `make-elk-service.sh` script, which is a pending PR: https://github.com/18F/cf-service-proxy/pull/9

If that PR changes in some way, I can change/revert the commit 4238051 that references it.


pkarman:
@ozzyjohnson this is ready for your review

PR #142, opened 2015-12-23
Conditional deployments with Travis

LinuxBozo:
@mogul Captures the conditional builds with travis.

@adelevie Now that you have followed this process, is there anything else you would add/change here?


adelevie:
This is great. It would go well with a section on using `inherits` in the manifest. (I still need to actually do that in my app!)

PR #144, opened 2015-12-24
rearrange menu sections

afeld:
The biggest change in this pull request moves all the navigation items previously under Docs -> Operations to a Contributing section of the menu. The Contributing section is really intended for cloud.gov staff, but @mogul suggested the word "Contributing", which I think makes it more welcoming than something like "Internal Docs". Totally open to other ideas for the menu/sub-menu titles.

There are no content changes – just navigational. See commits for more specifics.

**Before**

screen shot 2015-12-27 at 9 55 17 am

**After**

screen shot 2015-12-24 at 11 27 50 am

/cc @jameshupp

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -56,3 +56,15 @@ title = "docs.cloud.gov"
 [[menu.main]]
     name = "Contributing"
     identifier = "ops"
+[[menu.main]]
+    name = "Deployment"
+    identifier = "deployment"
+    parent = "ops"
+[[menu.main]]
+    name = "Users"
+    identifier = "users"
+    parent = "ops"
+[[menu.main]]
+    name = "Operations"
+    identifier = "operations"
+    parent = "ops"
--- a/config.toml
+++ b/config.toml
@@ -52,12 +52,6 @@ title = "docs.cloud.gov"
     identifier = "apps"
     pre = "<i class='fa fa-file-text'></i>"
     weight = -110
-[[menu.main]]
-    name = "Standards"
-    parent = "usage"
-    identifier = "standards"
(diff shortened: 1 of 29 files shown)


mogul:
@jameshupp What do you think? I'm fine with this if you are... Go ahead and merge if so.

jameshupp:
This looks cool to me, with the one caveat that the distinction `Contributing > Deployment` vs. `Usage > Deploying Apps` is not obvious to me. The distinction is probably better understood to developers, so I think it's fine. But if you _do_ think this is potentially confusing, consider further modifying "Deployment" under Contributing. Are you contributing to what we deploy, how we deploy, how users deploy, etc.?

To reiterate, if this is already clear to people with the skills required to actually manage apps in cloud.gov, the existing PR is fine.

afeld:
Hmm. The distinction is between people using the platform vs people running (or generally wanting to help with) the platform. The latter could arguably be a whole different (section of the) site, but I don't think we want to make that split right now. "Usage" and "Contributing" are fairly common verbs(?) corresponding to two groups in the open source world, but I am totally open to other ideas.

Will merge later today if there are no suggestions...easy to change it later if we have an epiphany.

dlapiduz:
what do you all think about moving the "contributing" outside of the main "docs" site?

jameshupp:
@afeld I get that, and I think it's perfectly good. Just wanted to flag because of proximity.

afeld:
@dlapiduz I don't think it's a bad idea, but I say we keep that discussion separate. If the plan is still to merge the landing page and the docs site, I think it could just be a subsection of the site, just like the user docs would be.

PR #148, opened 2015-12-29
add more information around user accounts

afeld:
See commits.


content/getting-started/accounts.md
```diff
@@ -0,0 +1,15 @@
...
+menu:
+  main:
+    parent: getting-started
+title: Accounts
+weight: -100
+---
+
+## Setting up your account
+
+You will need a Cloud Foundry account before continuing. If you don't already have an account, please [create an issue in the DevOps issue tracker](https://github.com/18F/DevOps/issues/new) and assign it to [@dlapiduz](https://github.com/dlapiduz) or [@ozzyjohnson](https://github.com/ozzyjohnson).
```
afeld: @dlapiduz @ozzyjohnson Is [the page around creating users](https://docs.cloud.gov/ops/create-user/) still up-to-date, now that the platform allows us to [invite users](https://login.cloud.gov/invitations/new)?

mogul: Absolutely not. The process right now should be to ask someone who already has access to invite you using that page, even for 18F'ers.

mogul:
While you're at it, can you please direct users to [the Cloud Foundry community docs for installing the CLI](https://docs.cloudfoundry.org/devguide/installcf/install-go-cli.html)? (Reason: We really shouldn't be directing people to the Pivotal version, and we shouldn't be duplicating docs that exist and are maintained elsewhere.)

content/getting-started/setup.md
```diff
@@ -35,19 +35,15 @@ cf -v
...
 ### Log in
 
+[Once you have an account]({{< relref "accounts.md" >}}), run
+
 ```bash
 cf login -a api.cloud.gov
 ```
 
-Once you log in for the first time, you'll probably want to change your password with:
+If you were given a password, you'll probably want to change it with:
```
dlapiduz: This process also might need to be changed. Now that we send invites rather than creating users the process requires you to set a strong password.

afeld:
Updated! Ended up collapsing a bunch of pages by relying on the community documentation, which seemed to cover everything that we had in ours.

content/getting-started/accounts.md
```diff
@@ -0,0 +1,19 @@
+---
+menu:
+  main:
+    parent: getting-started
+title: Accounts
+weight: -100
```
dlapiduz, line 6: Why the -100?

dlapiduz:
General :+1:, not sure about lowering the priority on the accounts section given that it is literally the first thing the user needs...

LinuxBozo:
@dlapiduz pretty sure Hugo uses "golf score" like weight numbers.. the lower the number, the higher in the ordering it appears..

mogul:
:clap: Wow, this is great. Thanks for all the work, @afeld!

dlapiduz:
Thanks @LinuxBozo... yeah I missed that!

content/getting-started/accounts.md
afeld, line 6: Pretty sure I just copied this from another page. I've only been adjusting the `weight`s as necessary.

PR #152, opened 2015-12-30
Need to connect to the bosh-lite instance prior to running cf

sharms:
These commands will not work without first ssh'ing back into the bosh-lite vagrant VM

### DIFF
--- a/content/ops/creating-a-local-dev-environment-in-Virtual-Box.md
+++ b/content/ops/creating-a-local-dev-environment-in-Virtual-Box.md
@@ -196,6 +196,8 @@ Secure copy the cflinuxfs2 container to the CF VM.
 
 Connect using the Cloud Foundry CLI.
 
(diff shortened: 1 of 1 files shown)


content/ops/creating-a-local-dev-environment-in-Virtual-Box.md
```diff
@@ -196,6 +196,8 @@ Secure copy the cflinuxfs2 container to the CF VM.
 
 Connect using the Cloud Foundry CLI.
 
+        # connect back to the bosh-lite VM instance
+        vagrant ssh
```
afeld, line 200: Looks like this isn't indented the same as the commands below? Causing a formatting problem.

sharms:
Looks like tabs vs spaces, good catch I will reformat --

afeld:
Feel free to use the same branch and re-open when ready!

PR #157, opened 2015-12-30
Update titles to be consistent

sharms:
Used Ruby Titleize to make all titles consistent across docs: https://github.com/granth/titleize
Clarified that commands need to be run in Bosh Lite Vagrant instance


dlapiduz:
:+1:

afeld:
Nice, thanks! Did you have to write a script to do it? Would you mind posting the code?

PR #166, opened 2016-01-08
add a page on managing teammates

afeld:
Right now, developers at 18F are asking the cloud.gov team to add new people that join teams, but with the invite functionality, I think this can be all self-service. Not positive these instructions are accurate, so please verify!

### DIFF
--- a/content/apps/managing-teammates.md
+++ b/content/apps/managing-teammates.md
@@ -2,7 +2,7 @@
 menu:
   main:
     parent: advanced
-title: Managing teammates
+title: Managing Teammates
 weight: 10
 ---
 
--- a/content/apps/managing-teammates.md
+++ b/content/apps/managing-teammates.md
@@ -9,7 +9,8 @@ weight: 10
 To add a new teammate to your organization:
 
 1. [Send them an invite](https://login.cloud.gov/invitations/new).
-1. Run
+1. Confirm that they received the email.
+1. Once they have accepted the invite, run
 
     ```bash
     cf set-org-role <email> <org> OrgManager
--- a/content/apps/managing-teammates.md
+++ b/content/apps/managing-teammates.md
@@ -9,17 +9,29 @@ weight: 10
 To add a new teammate to your organization:
 
 1. [Send them an invite](https://login.cloud.gov/invitations/new).
-1. Confirm that they received the email.
-1. Once they have accepted the invite, run
+1. Confirm with them that they have received and accepted the invite.
+1. Ensure you are running version >= [6.14.0](https://github.com/cloudfoundry/cli/releases/tag/v6.14.0) of the CLI. If not, you will need to [upgrade](https://docs.cloudfoundry.org/devguide/installcf/install-go-cli.html).
+
+    ```bash
+    cf -v
+    ```
+
+1. To give them read-only permissions to the organization, run
+
+    ```bash
+    cf set-org-role <email> <org> OrgAuditor
+    ```
+
+1. If you want to make them an admin for your organization, run
 
     ```bash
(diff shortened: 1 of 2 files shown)


content/apps/managing-teammates.md
```diff
@@ -0,0 +1,24 @@
...
+1. Run
+
+    ```bash
+    cf set-org-role <email> <org> OrgManager
+    ```
+
+1. Add them to each space:
+
+    ```bash
+    cf set-space-role <email> <org> <space> SpaceDeveloper
```
afeld, line 34: Should this be `SpaceManager` instead? Both? Is there an easier way to accomplish all of the role assignment, e.g. through the deck?

content/apps/managing-teammates.md
```diff
@@ -0,0 +1,24 @@
...
+weight: 10
+---
+
+To add a new teammate to your organization:
+
+1. [Send them an invite](https://login.cloud.gov/invitations/new).
+1. Run
+
+    ```bash
+    cf set-org-role <email> <org> OrgManager
```
mogul, line 28: Making them an OrgManager allows them to add other people as well... This should only be done if you want to distribute administrator powers for the org. At 18F maybe that's OK/normal, but I think not in most places. Maybe safer to have this doc suggest "OrgAuditor" (the read-only role) and let them find OrgManager on their own later when they are actually looking for that admin distribution.

content/apps/managing-teammates.md
mogul, line 34: It shouldn't be `SpaceManager` unless you intend for that user to also be able to manage space roles for other people.

It's possible to set roles through the deck, yes, but the UX is kind of sucky right now so I wouldn't send them there. Or if you do let them know it may be easier but that we're looking for feedback on an early product. 

(You should probably add a note to make sure they're running the updated CLI, though, since they may get errors from the CLI otherwise and I'm sure very few people have updated to the version that eliminates them.)

content/apps/managing-teammates.md
```diff
@@ -0,0 +1,24 @@
...
+  main:
+    parent: advanced
+title: Managing Teammates
+weight: 10
+---
+
+To add a new teammate to your organization:
+
+1. [Send them an invite](https://login.cloud.gov/invitations/new).
+1. Run
```
mogul: For new users, there's an implied waiting period between these steps... There's no way to set a role on someone who hasn't yet confirmed their invitation. Probably best to insert something like _wait for confirmation that their account is set up_ as a second step. It's pretty crappy UX in that way, but we have bigger UX fish to fry before we can tackle it (in the deck; can't do anything for the CLI).

arowla:
Thanks, @afeld. This was helpful. In my case, I just assigned folks to be OrgManager, and it seems, at least so far, like that was enough.

afeld:
Well, I suppose if they get made OrgManagers, they can assign the other permissions to themselves...?

content/apps/managing-teammates.md
afeld: Are you sure that's the case?

mogul: Pretty sure, yeah, although I'd love to be proven wrong.

mogul: You could leave it as-is and see how much it comes back to bite us...

content/apps/managing-teammates.md
afeld, line 34: Does anyone know what minimum CLI version they should have?

mogul, line 34: https://github.com/cloudfoundry/cli/releases/tag/v6.14.0

afeld:
Updated!

PR #167, opened 2016-01-11
move Managed Services under Deploying Apps, updated for v2

pkarman:


content/apps/managed-services.md
```diff
@@ -25,54 +21,42 @@ The v1 services in [cf-services-contrib-release](https://github.com/cloudfoundry
 
 	brew install jq
```
mogul, line 22: No idea what `jq` might have been used for... This must be some ancestral remnant. For CLI installation we should point users to the [setup page](https://docs.cloud.gov/getting-started/setup/).

pkarman, line 22: yes I just copy/pasted that from the old doc. I will create new PR that points at setup page.

PR #170, opened 2016-01-11
Fix table alignment

mogul:
This is just a stop-gap until major styling of the docs is done.

### DIFF
--- a/content/intro/pricing/rates.md
+++ b/content/intro/pricing/rates.md
@@ -20,22 +20,22 @@ Packages are selected based on the kinds of applications to be hosted.
 
 <table>
   <tr>
-    <th>Sandbox</th>
-    <th>Prototyping</th>
-    <th>FISMA Low</th>
-    <th>FISMA Moderate</th>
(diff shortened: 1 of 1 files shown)


msecret:
Does this actually work? I didn't think you could specify percentages as html attributes. But if it works then this is fine.

NoahKunin:
@msecret %s will work as HTML attributes (at least it does in Chrome). 

The problem is the use of `valign`. That's not a real property. The correct property is called `vertical-align`. The current edit there won't do anything.

mogul:
That's weird, because it works fine locally in Chrome:
image

`valign` is the HTML attribute equivalent of `vertical-align` in CSS... Maybe that's what you're thinking of. 

Are you guys actually seeing this not work locally, or are you just guessing it will not work based on the code?

msecret:
Just guessing.

NoahKunin:
I actually did throw it in the console and got an error for `valign` - now I see why. Page is supposed to be conformant to HTML5, and `valign` was deprecated.

PR #177, opened 2016-01-15
Add key rotation page

dlapiduz:
We need this policy to comply with SC-12.

/cc @NoahKunin @mogul @clovett3


NoahKunin:
Merging and then editing, as creating a branch on a fork (actually would be a fork of a fork) seems too complicated.

PR #186, opened 2016-01-24
Mild re-write of key policy

NoahKunin:
Edits for clarity / consistency / layman-ship (not sure if that's a word).

One thing I'm happy to go either way on is if we post links in the clear or hyperlink them to the plain text. I'm not sure we consistent on either style choice yet for our docs. For now, I'm a fan of hyperlinking and keeping full URLs out of the flow, but I'm not dogmatic if the consensus goes the other way.

### DIFF
--- a/content/ops/key-rotation.md
+++ b/content/ops/key-rotation.md
@@ -6,27 +6,27 @@ title: Key Rotation and Availability
 weight: 10
 ---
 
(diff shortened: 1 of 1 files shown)


dlapiduz:
I am not dogmatic either but I'd rather have a URL rather than "[click here](#this)" because I find myself missing the link a lot. That said lets merge the changes and have that discussion some other time..

PR #189, opened 2016-01-27
Simplify "amount of work"

thecapacity:
It was unclear from the wording if the planning (i.e. the "it") was a significant amount of work or if the actual O&M was the focus of "significant". I think the latter was intended more than the former, but this seems to simplify.


afeld:
:+1:

PR #214, opened 2016-02-09
add example of static site with redirect

pkarman:


dlapiduz:
This build passed but the status is not here :s

PR #216, opened 2016-02-15
Use the PORT env var instead of VCAP_APP_PORT

alex:
Per the CF docs, VCAP_APP_PORT is deprecated: https://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html#VCAP-APP-PORT


afeld:
Thanks!

alex:
Hapilly! I imagine some other tutorials might have this same issue, I didn't look at them all, just the one I was reading.

PR #221, opened 2016-02-22
Fix MB pricing to be daily not hourly!

mogul:


afeld:
Heh, whoops! :shipit:

PR #222, opened 2016-02-23
simplify Travis instructions

afeld:
Rely on the official docs rather than having them copied into our own.

/cc @LinuxBozo @konklone @adborden


konklone:
:+1:

PR #223, opened 2016-02-23
Update the path located in staticfile

jcscottiii:
Building the site static pages with hugo and building the assets (cg-style) now move things into the public directory. https://github.com/18F/cg-docs/blob/6162be53b1cb9f14082eddab85ff36f0f8f5be7b/package.json

### DIFF
--- a/Staticfile
+++ b/Staticfile
@@ -1,1 +1,1 @@
-root: out
+root: public


dlapiduz:
ok, lets give it a try

PR #226, opened 2016-03-07
clarifying the timeframe for the sandbox resource cap

gbinal:
@mogul, @dlapiduz, or another should confirm if it's correct that the 1 GB resource limit for sandbox accounts is per month, but if it is, this will help clarify the time period of that limit, since I didn't see one when reading it.  .


dlapiduz:
thanks @gbinal!

PR #229, opened 2016-03-10
Clarity on quotas and charges

NoahKunin:
There still seems to be confusion popping up about this. Getting a bit more literal to hopefully avoid confusion.


wslack:
One question about the last phrase you edited - I would just link to the quotas page instead of saying "you can raise" because it's not clear to me that people can raise their own quotas. If they can, I would still link to the quotas page.

wslack:
Merging per conversation with @NoahKunin

PR #238, opened 2016-03-21
recommend the good_migrations gem for Rails apps

afeld:
Just learned about this gem through http://blog.testdouble.com/posts/2016-02-05-good-migrations.html, and I've been bitten by the problem enough to recommend it across the board.

/cc @monfresh #36

### DIFF
--- a/content/apps/rails.md
+++ b/content/apps/rails.md
@@ -15,6 +15,7 @@ See the official CF guide: [Getting Started Deploying Ruby on Rails Apps](http:/
 
 1. [Exclude files ignored by Git]({{< relref "apps/deployment.md#exclude-files" >}}).
 1. [Add the `rails_12factor` gem](https://github.com/heroku/rails_12factor#install) for better logging.
+1. Add the [`good_migrations`](https://github.com/testdouble/good-migrations) gem for safer migrations.
 1. Create the application on Cloud Foundry.
 
     ```bash


jessieay:
this is a great idea! and will certainly keep me in check. I recently referenced AR in a migration and it completely failed when running on production a month or two later because of some additional migrations we'd added in the meantime.

jessieay:
one question: should this be somewhere else instead? I don't usually think to look at cloud.gov docs for info on how to build good software (although right now I am not sure which of our guides I would look at, tbh, since there are so many), usually just look here when I am trying to debug something with a deploy.

afeld:
Open to ideas! Could certainly see a section of https://pages.18f.gov/before-you-ship/ being dedicated to general good production practices for frameworks, but we don't have that yet.

monfresh:
Since this is specific to Rails, I think this would fit well into a Rails best practices guide, which could be a separate page under our development guide. I've been meaning to get a Rails guide started. I'll submit an initial draft this week.

afeld:
I say we merge this here, then move it if a better place avails itself.

monfresh:
I tried the gem a while back and it didn't do what I expected it to. See https://github.com/testdouble/good-migrations/issues/5 

Have you actually tried the gem and gotten it to fail a migration? 

I wouldn't want to recommend a gem that we haven't confirmed is useful.

afeld:
I don't really have any Rails apps I'm working on to test it with...can't remember if I did so at the time either, sorry! We can close if we don't feel confident in it.

dlapiduz:
Going to close this, if we have the opportunity to test further, lets revisit it.

PR #240, opened 2016-03-26
[WIP] Embeds FedRAMP status report exposed by 18f.aha.io

mogul:
WORK IN PROGRESS, do not merge yet.

This is nearly ready, just needs someone smarter than me to figure out why the timeline gets cut off when it’s fullscreen’d.

### DIFF
--- a/content/fedramp-tracker.md
+++ b/content/fedramp-tracker.md
@@ -1,0 +1,4 @@
+---
+title: FedRAMP Moderate JAB ATO tracker
+---
+<iframe src="https://18f.aha.io/published/e24a68aec8958c8a74786e699264116d" style="width: 100%; height: 550px;" frameborder="0" scrolling="auto" seamless="seamless" allowfullscreen="true"></iframe>
--- a/content/fedramp-tracker.md
+++ b/content/fedramp-tracker.md
@@ -1,4 +1,0 @@
----
-title: FedRAMP Moderate JAB ATO Tracker
----
-<iframe src="https://18f.aha.io/published/e24a68aec8958c8a74786e699264116d" style="width: 100%; height: 550px;" frameborder="0" scrolling="auto" seamless="seamless" allowfullscreen="true"></iframe>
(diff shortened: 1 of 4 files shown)


mogul:
@msecret can you help?  Compare with [visiting the aha.io notebook directly](https://18f.aha.io/published/e24a68aec8958c8a74786e699264116d).

mogul:
I haven't exposed this directly in the nav hierarchy at all yet... Should I?

afeld:
Hmm, this is a bit funky:

screen shot 2016-03-26 at 12 03 34 pm

Not a blocker, especially if you want this merged by Monday.

> I haven't exposed this directly in the nav hierarchy at all yet... Should I?

Yup! Maybe under Security?

``` yaml
menu:
  main:
    parent: security
```

Feel free to merge after adding that.

mogul:
@msecret not sure how to approach either of those (truncated timeline or squished arrows). I'm hoping to do a lot more stuff like this using views from aha.io... Maybe a different layout is warranted?

msecret:
@mogul I looked into this, the aha code is not responsive below a certain width. Once it gets to a small width, such as the width that it's being displayed at on the docs website, all parts of the page will cut-off or have layout issues.

Now that I'm seeing that it's more then just a problem with the arrows, I don't think the solution is to overwrite their CSS, that sounds like a lot of work in a codebase that isn't our own (plus it's hard to change CSS cross-origin). I think we'll need a different layout to handle these pages, or change the docs site so it's width is around 100-200px larger. This still won't help at all when the docs site has to be responsive.

Because of my permissions in aha, I'm not able to see the notebooks feature, but does it have any sort of controls for the branding or layout of the page? What is our support relationship with the company? Could we as for features like more responsive pages?

mogul:
Aha has an [ideas portal](https://big.ideas.aha.io). I see some in there
asking for other parts to be responsive, but not the "Notebooks" in
particular. Feel free to ask in detail for what you want!

I agree with having a special layout template in docs.cloud.gov for large
embeds like this... It's basically meant to be viewed in full-screen in any
case, like a presentation. So I'd be fine with a layout that basically just
fills the window, so long as it's still docs.cloud.gov in the URL bar! If
we do that, then I'll just add a link from the minimal text at the top of
this page to wherever we put it.

PR #241, opened 2016-04-04
Add `cf files` documentation;

rogeruiz:
This patch adds documentation for `cf files` under Databases as it seems
to be the most useful when paired with accessing `psql` on the server.

### DIFF
--- a/content/apps/databases.md
+++ b/content/apps/databases.md
@@ -62,3 +62,36 @@ To access a service database, use the [cf-ssh]({{< relref "getting-started/cf-ss
     ./psql/bin/psql $DATABASE_URL
 
 You should now have an open `psql` terminal connected to the service database.
+
+### Accessing files created on the server using cf files
+
+When accessing a database using `cf-ssh`, you may produce a database dump file
+and will very likely need to pull it down from the server. You can easily print
+out any file on your server to your terminal via the `cf files` command. You
+have to make sure you haven't disconnected from your ssh session that you
+created with `cf-ssh`.
+
+```sh
+cf files <APP_NAME> [PATH] [-i INSTANCE]
+```
+
+The `<APP_NAME>` can be found by running `cf apps` in your terminal. You should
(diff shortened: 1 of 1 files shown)


rogeruiz:
cc: @afeld @pkarman Let me know if that language works for using `cf files`. I can also move it to a more general place as well, or maybe link to it from the [General page](https://docs.cloud.gov/apps/deployment/) as well? 💭

content/apps/databases.md
```diff
@@ -62,3 +62,36 @@ To access a service database, use the [cf-ssh]({{< relref "getting-started/cf-ss
...
+```sh
+cf files <APP_NAME>
+```
+
+To save a file to your local machine, you can redirect the output from `stout`
+to a file using the `>` character and then a file path on your local machine.
+Here's an example:
+
+```sh
+cf files my-app-ssh app/path/to/database_dump.sql > ./database_dump.sql
```
jmcarp: At least on cf cli v6.15, this adds some extra lines to the file, e.g.

```
Getting files for app eregs-web in org eregs-nc / space dev as joshua.carp@gsa.gov...
OK

<real content starts here>
```

How about something like `cf files my-app-ssh path/to/thing | tail -n +3 > out.sql`?

rogeruiz: yeah. That's a great suggestion @jmcarp Thanks. I'll add it in.

content/apps/databases.md
```diff
@@ -93,5 +93,5 @@ to a file using the `>` character and then a file path on your local machine.
 Here's an example:
 
 ```sh
-cf files my-app-ssh app/path/to/database_dump.sql > ./database_dump.sql
+cf files my-app-ssh app/path/to/database_dump.sql | tail -n +4 > ./database_dump.sql
```
rogeruiz, line 96: Ignoring the first few lines so files don't need to be edited after downloading. Thanks @jmcarp! 🎉

pkarman:
tested and works wonderfully.

PR #242, opened 2016-04-11
Simplify confusing 3x FISMA Low package

mogul:
The 3x FISMA Low bundle no longer makes sense, so we're just simplifying to the price per single system.


mogul:
@dlapiduz Looks like this didn't deploy...?

dlapiduz:
The password for the deployer user expired, changed it now.

PR #260, opened 2016-05-13
switch to simple heading IDs

afeld:
Before and after:

screen shot 2016-05-13 at 4 02 28 pm

No real benefit to the longer ones, from what I can tell. Not sure why that's the default.

http://gohugo.io/overview/configuration/#configure-blackfriday-rendering

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -6,6 +6,8 @@ publishdir = "public"
 [params]
   description = "Cloud.gov documentation"
   author = "18f"
+[blackfriday]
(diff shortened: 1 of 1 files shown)


mogul:
It's due to Hugo going out of its way to avoid anchor collisions when you use `ref` and `relref`... There's a [section in the docs](https://gohugo.io/extras/crossreferences/#hugo-heading-anchors) about it. Let's try turning off the strict ones and see if we get complaints.

PR #262, opened 2016-05-24
add more explicit instructions around creating deployer users

afeld:


mogul:
This is mixing instructions for admins into the instructions given to users... Regular users can't create users from the `cf` CLI, which is why they have to ask admins. So I think the bulk of this belongs in the `Contributing > Operations` section.

mogul:
To make it truly easy, you could just make it all an alias that takes the $ORGNAME as an argument and does the rest. 

Also note `pbcopy` is Mac-specific... I know current admins are all Mac-based at 18F, but I don't know if we want to rely on that in the instructions, particularly as admin for the platform may transition to another organization over time.

mogul:
Maybe skip both use of `pbcopy` and the extra "paste" step by using `curl` to post the password at Fugacious and get the sharing link back... Not sure if Fugacious supports it but it'd be sweet. :)

mogul:
Last comment for now, sorry for not bunching these up: We should be setting only an OrgAuditor role, and telling the OrgManager/person-who-requested-the-deployer to set SpaceDeveloper roles appropriate to their use-case (just one space or all spaces, etc)

jmcarp:
Is this still needed?

afeld:
Nope!

PR #268, opened 2016-06-02
Update url for CF-SSH

dlapiduz:
Moving the releases to the cf-ssh repo in the 18F org


dlapiduz:
The new version has the ability to tell its version! revolutionary

PR #273, opened 2016-06-10
automatically deploy via Concourse

afeld:
Closes https://github.com/18F/cg-atlas/issues/40.

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -10,8 +10,8 @@ jobs:
       image_resource:
         type: docker-image
         source:
-          repository: jojomi/hugo
-          tag: '0.16'
+          repository: mhart/alpine-node
+          tag: '4'
       inputs:
       - name: cg-docs-source
       outputs:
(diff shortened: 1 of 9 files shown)


afeld:
It lives!
- https://ci.fr.cloud.gov/pipelines/deploy-docs/jobs/deploy-production/builds/18
- https://docs.fr.cloud.gov/

Ready for 👀 .

ci/pipeline.yml
```diff
@@ -0,0 +1,85 @@
...
+      - name: cg-docs-compiled
+      run:
+        dir: cg-docs-source
+        path: sh
+        args:
+        - -exc
+        - |
+          # install dependencies
+          # git required for NPM, the rest required for node-sass->node-gyp
+          apk add --no-cache --update g++ git make nodejs python
```
afeld: This was surprising. You learn a lot about your dependencies when you try to install them in minimalist Docker containers 😉 

/cc @msecret @jeremiak

afeld:
Note this pull request removes the deployment through Wercker, so we will need to set up the Concourse pipeline to keep continuous deployment to https://docs.cloud.gov.

ci/pipeline.yml
msecret: So we actually don't need to compile anything with node-sass in this project, the build process really just copies files from cg-style into a public folder so it can be referenced? At what part of the build process were you getting errors about not having python?

msecret: @afeld @dlapiduz , tested locally, and found `node-sass` isn't an actual dependency it was probably considered one early in development but then we found we could build the docs site without compiling from sass. If you remove `node-sass` from `devDependencies` then python, make g++ shouldn't be required. Hope that helps a bit.

afeld:
Updated!

## Post-merge
- [x] Set up pipeline for docs.cloud.gov

PR #282, opened 2016-06-16
Mention "cf files" in cf-ssh instructions

toolness:
I had to transfer some files from a `cf-ssh` session today and it took a little while to figure out, so I thought that mentioning `cf files` with example usage might help future passers-by.


afeld:
Thanks!

Issue #283, opened 2016-06-16
Configuration Management plan


LinuxBozo:
In order to satisfy the CM family of NIST controls (CM-9/10/11), the cloud.gov team needs to have a written Configuration Management plan.
- [x] docs.cloud.gov contains a page with the configuration management plan

PR #285, opened 2016-06-16
Re-purpose and expand code-review for configuration management plan

LinuxBozo:
@dlapiduz Pretty sure this satisfies #283 

@brittag For your review as well.


dlapiduz:
@LinuxBozo I think we need a section on "What should go into config management" (everything), "How is config management tested" etc.

Here are the notes from the follow ups:

```
All changes to a configuration needs to go to GitHub
GitHub changes need to be peer reviewed before accepted
Changes need to be deployed to a staging environment, tested for functionality and security before pushing to production
Tools should revert to known configurations if there are any unauthorized changes to the environment
Secrets should be uploaded to a file 
```

dlapiduz:
I know this covers the need peer review part

dlapiduz:
I think most of it is there... we just need to spell it out a bit more

LinuxBozo:
@dlapiduz Ping. Made more changes that I think addresses everything now.

PR #292, opened 2016-06-23
IR guide/checklist suggestions to sync better with 18F IR process

brittag:
While reading the cloud.gov IR checklist update at https://github.com/18F/cg-docs/pull/290 and the all-18F reporting process at https://github.com/18F/security-incidents, I noticed that the cloud.gov IR guide probably also needs to be updated to match with both of those. Here's a suggestion to try to help make these three documents mesh with each other better - totally fine to treat this as a thought experiment and make a somewhat different change instead. Things I did:
- Update the "Initiate" phase of this IR guide to match more with the updated cloud.gov checklist and the 18F IR guide: ask everyone to send emails and notify #incident-response as their first priority, with notifying #cloud-gov as a second priority.
- More directly explain what to do if you’re part of the cloud.gov team and you’re the first to notice a potential incident.
- Make the checklist language a little plainer; fix a typo in the later part of the guide.

I also [filed a suggestion for the 18F IR guide](https://github.com/18F/security-incidents/issues/6) to ask it to sync better with this guide - basically, it’d be helpful for that guide to specify at what point people should prioritize talking to the product team.


brittag:
Oh, should mention @NoahKunin @jacobian @kimberbat as people potentially interested in looking at this.

jacobian:
👍 awesome, thanks! 👏

content/ops/security-ir.md
```diff
@@ -44,14 +45,26 @@ For full details, read on.
...
 
-The reporter should contact the cloud.gov team in [`#cloud-gov`](https://18f.slack.com/messages/cloud-gov/) to begin the response process. They should escalate by contacting the cloud.gov leads directly until they receive acknowledgment of their report. The final escalation point is to page the infrastructure lead, whose contact info is in the channel details of [`#incident-response`](https://18f.slack.com/messages/incident-response/).
+When a person outside the cloud.gov team (the *reporter*) notices a cloud.gov-related incident, they should begin reporting it by using the [18F incident response process](https://github.com/18F/security-incidents#process), and then post about it in [`#cloud-gov`](https://18f.slack.com/messages/cloud-gov/). If they don't get acknowledgment from the cloud.gov team right away, they should escalate by contacting the cloud.gov leads directly until they receive acknowledgment of their report.
```
mogul, line 52: We should probably be more forward-thinking here, in that the reporter may be outside 18F entirely. This means they won't have access to our Slack or that private `security-incidents` repo. Maybe break it out into a list of best-to-worst latency (eg Slack, then repo, then cloud-gov-support)? Or maybe we need a special e-mail address specifically for security reports that is prioritized and goes to specific people/processes on the back-end?

brittag, line 52: Ah yeah, we should have brief clear explanations for how outside-18F people (both totally-random people and outside-18F cloud.gov users) should non-publicly report potential sensitive security incidents to the team, probably in all of the following places:
- A future more comprehensive "contact" page (at least on https://docs.cloud.gov/help/ for now)
- As a note on https://docs.cloud.gov/intro/security/auditing-contributing-security/
- This IR guide (which would just have a brief note saying "are you here to report a thing and aren't at 18F? email x please!")

The usual practice I've seen is to ask external people to email a special email address (like security@cloud.gov) for security reports that might be sensitive (even if right now that email would just go to the support mailing list), and ask people to report public issues in the appropriate GitHub repository for non-sensitive suggestions.

jacobian, line 52: My intent with this document was specifically internally-facing -- this is a guide for _staff_. I completely we need a catch-all contact for external reporters (having a security@cloud.gov and https://cloud.gov/security to point there would be my suggestion). 

But from an IR standpoint, those external people aren't reporters, they're a sort of signal or alert system. That is, an external ping to security@cloud.gov about an issue is technically the same as an alert from a IDS system, at least as far as IR process is concerned.

brittag, line 52: 👍  Yeah, let's figure out that external process as a separate task/issue/process that feeds into this one.

mogul:
Given the discussion above, is this ready to merge as-is?

content/ops/security-ir.md
mogul, line 52: Maybe just be explicit that this is aimed at 18F-internal folks, then.

brittag:
I added another commit to address these points in tiny incremental ways for now:
- In the "Initiate" overview, I clarified the "person outside the cloud.gov team" and "cloud.gov team member" role descriptions I'd written, switching this to just "An 18F staff member inside or outside the cloud.gov team", which feels more correct.
- I added a small sentence for "outside 18F? email us!" - we can fix that later with a more comprehensive strategy in all the right places.

Deferring to @jacobian to decide if this is ok to merge!

dlapiduz:
@jacobian what do you think?

PR #309, opened 2016-07-05
[WIP] Adding security responsibilities info to security page

brittag:
For https://github.com/18F/cg-docs/issues/257 I took a look at https://github.com/18F/cg-docs/pull/300 and tried to present a simplified version on a more relevant page. The ["What is cloud.gov?" overview page](https://docs.cloud.gov/intro/overview/what-is-cloudgov/) seems too high-level for details about security, so I tried adding this to the ["Security for all applications" overview page](https://docs.cloud.gov/intro/security/improving-application-security/) instead.

I took that chart and [revised it for readability in this Google Drawing](https://docs.google.com/drawings/d/1UBiOteSPXpA72KE52Kh-j7aYr73zTkzJ_oMuw5F293I/edit). That Google Drawing is easier to collaboratively edit than a static image, so I figured I'd leave the drawing there until we decide it looks good, and then we could integrate it into this PR.

cc @frsfx @afeld @dlapiduz for thoughts and revisions


frsfx:
@brittag I'm sorry you had to redraw the diagram.  I actually have the visio file if you like to use it?  I can export to another format with all the layers.  Let me know.

brittag:
@frsfx Thanks! Google Drawings is easy to work with, so that didn't take much time. I like that it allows multiple people to edit the same chart and see the updates live.

dlapiduz:
@brittag do you think that "security" is the right page for this? I think it should be as early on as possible...

afeld:
Content looks good to me! I kinda agree with Diego in that this might not be the right place for it, since the division of responsibilities is about security _and_ operations.

brittag:
Good point @afeld - this is really an illustration of the overall responsibility boundary, rather than specifically a security responsibility boundary.

I'm stuck on where this fits! This seems to be too much specific technical detail for the very first high-level "What is cloud.gov" summary page, so I'm curious to get a better understanding of the user need. Here's my hypothesis:

> I'm a potential customer curious about purchasing cloud.gov. In order to figure out if cloud.gov is the right product that I want to buy, I need to know what components this cloud.gov PaaS provides and what I'd be responsible for (compared to buying an IaaS or buying a different PaaS). I'm technical but not yet familiar with cloud.gov or Cloud Foundry, so I need a high-level plain-language overview of the cloud.gov platform components and what's supported (and what's not). I need to know this while reading the intro/overview materials, since this will have a big influence on my purchasing decision.

Does that sound right?

If that makes sense, it sounds like a reasonable approach is a brief sentence summarizing this information on "What is cloud.gov" summary page, which would link to a new page (with this detailed content) titled something like "What the cloud.gov PaaS offers" under "Introduction -> Technology".

Let me know if that makes sense, and then I'll dig in and make those changes. :D

brittag:
I went ahead and tried that alternate structure! I'm wondering if the bottom of the chart should explain which parts are handled by the IaaS provider underneath cloud.gov, but maybe that's a topic to illustrate with a different chart - this one is probably already complex enough.

afeld:
I say we get the diagram in there, merge this, and iterate.

dlapiduz:
Sounds good, @afeld would you do the honors?

PR #312, opened 2016-07-05
Update max `cloudgov-style` version to be <= 1.0

rogeruiz:
This patch uses the caret-ranges from the `npmjs` docs to allow for
updates from version `0.0.0` to `1.0.0`. Because the versions are being
updated across minor version numbers, targeting `^0.x` will prevent this
repo from needing to be updated between `0.6.x` as of this writing until
`1.x`.

@see https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
   },
   "dependencies": {
     "anchor-js": "git+https://github.com/msecret/anchorjs.git",
-    "cloudgov-style": "^0.3.4",
+    "cloudgov-style": "^0.x",
     "highlight": "^0.2.4",
     "jquery": "^2.2.0",
     "jquery.scrollto": "^2.1.2"


afeld:
LGTM, but I defer to @jeremiak and @msecret, as I know people have feels about how to do version specification 😉

msecret:
Just to check, won't this still require an update because the build and deploy for the project will still have to run? or I guess you could run a build/deploy manually in concourse?

rogeruiz:
@msecret the `npm install` happens with each deployment, so updating the version number for the package is done less frequent than the actual contents of the doc. So for instance, when I updated the documentation for `Troubleshooting Bosh` and it was merged in I was expecting it to take the latest `cloudgov-style` but the version was hardcoded to not upgrade past `0.3.4`. With this, it'll try upgrading to the latest version as long as it is below `1.0.0`. So as the documentation is updated, the styles package will be updated between deployments.

But yeah, when `cloudgov-style` is updated either a change to the repo, causing a Concourse deployment, or a build in Concourse will trigger an update.

msecret:
I'm fine to try this out. If we find that cg-style updates are breaking things unexpectedly (which shouldn't be the case), we can discuss other options then

PR #317, opened 2016-07-07
Update assets to use new navigation style

jeremiak:
A recent upgrade to `cloudgov-style` resulted in the docs site breaking in two ways:
1. The fonts were not found
2. The side navbar had wonky icons

This changes the path for fonts used in the build process and uses different icon classes.

It would also be great to see if we can work this repo and the cg-landing repo into the automated visual regression testing so we can maybe pick up on these breaking changes earlier.


dlapiduz:
@jeremiak is this ready to merge? LGTM

PR #318, opened 2016-07-07
Add note about sso accounts in fedramp

dlapiduz:


content/ops/managing-users.md
```diff
@@ -5,6 +5,12 @@ menu:
 title: Managing Users
 ---
 
+## Note for GovCloud environment
```
brittag, line 8: How do readers/users know whether they're in the GovCloud environment? I'd like us to add a little context here explaining how to tell, so readers know for 100% certain whether this note applies to them.

dlapiduz, line 8: @brittag good point, this is a note for administrators though and they should know which environment they are targeting. I was thinking of removing invite.cloud.gov from the GovCloud environment as a way to protect the users.

content/ops/managing-users.md
```diff
@@ -5,6 +5,12 @@ menu:
...
+
+Only single sign-on user accounts are allowed. Service accounts, such as deployer credentials, are permissible with the condition that they are scoped to a particular space with limited access.
+
+No local accounts to UAA shall be created for user access.
```
mogul, line 12: I don't think our users will know what UAA or "local accounts" are, so we should probably just strike this line, or move it to documentation for operators.

dlapiduz, line 12: This is the documentation for operators, isn't it?

content/ops/managing-users.md
```diff
@@ -5,6 +5,12 @@ menu:
 title: Managing Users
 ---
 
+## Note for GovCloud environment
+
+Only single sign-on user accounts are allowed. Service accounts, such as deployer credentials, are permissible with the condition that they are scoped to a particular space with limited access.
```
mogul, line 10: On the first sentence: Will users know what a single sign-on user account is, and do they have any control over it? I think we should reframe this from the user's perspective... For example "We only provision access to this environment for agencies prepared to actively manage their user accounts, eg via a SAML identity provider. If you don't see your agency listed on [the login page] you may wish to check with your agency to see if they can provide this capability; we will set up integration upon request."

dlapiduz, line 10: We definitely should have language like that if this is not an ops page, but I think it is...

content/ops/managing-users.md
mogul, line 12: Oh right, I thought this was the "managing-teammates" page.

content/ops/managing-users.md
mogul, line 8: That's not a bad way to handle it, but in that case we need to message it carefully on the user-facing docs too.

rogeruiz:
We'll iterate on the messaging that @dlapiduz and @mogul brought up in the `line-note` comments in another PR.

PR #320, opened 2016-07-11
Added instructions, how to add different OAuth client, and how to use it in CI pipelines to obtain long lived tokes

s-matyukevich:


mogul:
@brittag can you put :eyes: on this one?

mogul:
We don't want to manage the complexity of our users having to use a special version of the CLI, so this one is blocked on the acceptance of the CLI change upstream. (Also the repositories it points to for the plugin, etc. should be moved to the 18F org.)

jmcarp:
Are we still planning to handle deployer accounts like this @LinuxBozo @mogul? I have a vague memory of the UAA team planning work on persistent tokens, but maybe I dreamed it.

brittag:
(I removed my assignment from this since it seems to need additional technical decision-making before I should look over it from a content perspective - feel free to re-assign me when I can be helpful for this. :))

mogul:
> Are we still planning to handle deployer accounts like this @LinuxBozo @mogul? I have a vague memory of the UAA team planning work on persistent tokens, but maybe I dreamed it.

That's correct. I'm going to close this one as we won't be pursuing this path.

PR #323, opened 2016-07-12
link to the Python buildpack documentation from the relevant pages

afeld:
Fixes https://github.com/18F/cg-docs/issues/322.

/cc @pburkholder

### DIFF
--- a/content/apps/django.md
+++ b/content/apps/django.md
@@ -5,6 +5,8 @@ menu:
 title: Deploying Django
 ---
 
+***See also: the [official Python buildpack documentation](http://docs.cloudfoundry.org/buildpacks/python/).***
+
 ## Creating the App
 
 There's no explicit command to create an app, but:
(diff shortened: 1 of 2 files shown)


content/apps/django.md
```diff
@@ -40,6 +38,8 @@ dj-database-url==0.3.0 #grabs environment variables and dumps them into a Django
 waitress==0.8.9 #a pure python WSGI server that is a replacement for gunicorn
 ```
 
+Note that [you will need to vendor the dependencies](http://docs.cloudfoundry.org/buildpacks/python/#vendoring) before deploying.
```
cmc333333, line 41: "Need" is probably too strong here -- I don't think any of the current cloud.gov users do this. The build pack installs libs from pypi.

afeld, line 41: The link there makes it sound required...possible that we are using an old buildpack and `pip install` has since been removed?

/cc @LinuxBozo

pburkholder, line 41: The buildpack will include pypi dependencies, but if you have things in your `requirements.txt` like:

```
git+https://github.com/JoshData/commonmark-py-plaintext
git+https://github.com/if-then-fund/django-html-emailer
```

then you will  need to run `pip download`

pburkholder:
Would the end-users be better served by saying that Cloud.gov support Django, and then providing links to the docs at Pivotal/cloudfoundry if there's no difference in the install process? 

Trying to maintain two document repos sounds hard.

content/apps/django.md
afeld, line 41: Asked for clarification in https://github.com/cloudfoundry/docs-buildpacks/issues/83.

afeld:
Absolutely...a medium-term goal for us (me) is to move as much of our documentation upstream as possible, and language/framework-specific information would be a good place to start. @mogul @brittag Should we create an issue to do an audit, or...?

PR #324, opened 2016-07-13
fix instructions around account creation

afeld:
/cc @brittag


brittag:
This makes sense to me! Let's update https://docs.cloud.gov/apps/managing-teammates/ as well to note that GSA + EPA people don't need invites.

PR #326, opened 2016-07-14
How to export a database

dogweather:
Here's what I'm thinking. I realize that I performed some major surgery here:
- Made a new higher-level heading containing the word **export**
- Removed non-essential information, such as the full usage specs for `cf files`. Instead, we can link to its full docs.
- Rewrote the text as sequential steps.
- Presented the code snippets as concretely as possible, with parameters filled in.

To consider:
- [x] Is `pg_dump` installed by default on cloud.gov VM's? These docs assume it is. (But `psql` isn't.) **Answer: Not installed in the C2 production VM.**
- [x] How hard would it be to automate this entire process?


dogweather:
@afeld Can `pg_dump` be added to the default VM? Alternatively, can manifests be written to include `pg_dump`?

afeld:
> Can `pg_dump` be added to the default VM?

What if there were a standalone `pg_dump_buildpack` that you could specify as part of a short-lived app?

> can manifests be written to include `pg_dump`?

`cf push` will upload any files in the local directory, so the way to do it is have `pg_dump` downloaded there.

dogweather:
Thanks! So I've discovered that the `psql-9.4.4-ubuntu-14.04.tar.gz` which we point to from [here](https://docs.cloud.gov/apps/databases/) actually contains a full set of postgresql executables, including pg_dump.

So I can definitely script its download & extraction. And can create one line to do this for these docs.

Yeah, long term, a buildpack for this might be a nice solution. I've also seen that we could make a pg tools project by using [apt-buildpack](https://docs.cloud.gov/apps/apt-get/).

afeld:
All of these are band-aids until we get DEAgo, so I wouldn't put a _huge_ amount of effort into it unless you hit that need often. Follow along at https://github.com/18F/cg-atlas/issues/102. Note that I'm not sure if DEAgo _automatically_ gets us `cf ssh`, but that's certainly the main step.

dogweather:
Ok, thanks.

content/apps/databases.md
```diff
@@ -63,35 +63,49 @@ To access a service database, use the [cf-ssh]({{< relref "getting-started/one-o
...
+Leave the ssh connection open.
+
+
+### Download
+
+On your local host:
 
 ```sh
-cf files <APP_NAME> [PATH] [-i INSTANCE]
+$ cf files <app name> /tmp/backup.sql > backup.sql
```
dlapiduz: @dogweather should we add the `-ssh` here? I think that is missing now that the following paragraph is killed

jessieay:
When I deploy database migrations that potentially transform data, I like to do a backup of my database beforehand. Until this is merged, I will need to reference this PR instead of the CG docs. 

Why not merge this to help me and others like me who want to export a dump of their db?

dogweather:
FYI, this is the process I use with C2 which I translated to this doc.

screen shot 2016-08-05 at 3 16 45 pm

afeld:
Thanks for doing this!

PR #335, opened 2016-07-20
Improvements to our docs for describing cloud.gov

brittag:
@berndverst and I spent an hour drafting some basic rewrites of our docs intro pages, as a start toward communicating better with people interested in cloud.gov.

(We also plan to do more in-depth restructuring of the docs later. This is intended to be an incremental improvement.)

### DIFF
--- a/content/index.md
+++ b/content/index.md
@@ -2,22 +2,16 @@
 type: index
 ---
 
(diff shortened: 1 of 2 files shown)


mogul:
👏

PR #341, opened 2016-07-22
Add rules of behavior for account usage

mogul:
The new section of this page is intended as a target for the link in the "acceptable use" clause behind the "agree and continue" button on login.cloud.gov.

cc @brittag


mogul:
(This is my best effort at condensing the stuff in the FedRAMP “Rules of Behavior” template located here: https://www.fedramp.gov/resources/templates-2016/)

afeld:
Made a few minor tweaks. Merging now so that we have it in there sooner than later...we can do follow-up edits as a new pull request.

PR #351, opened 2016-07-29
Adding system boundary diagram for reference

brittag:
Uploading our system diagram that explains our FedRAMP authorization boundary, so that it's easily available for reference in a handy public spot.


brittag:
P.S. Crediting @LinuxBozo for the most recent diagram update, thanks!

afeld:
Can we embed it on some page? Seems easy to lose track of otherwise.

afeld:
...with a comment about where to find the ~~original~~ source.

brittag:
Good point; I'm going to close this for now until we decide a few more things about this (ref #cloud-gov-highbar).

PR #355, opened 2016-08-02
Updating readme to clarify cloud.gov instead of CF

brittag:
"18F CF Docs" and "This repository contains the site code for the 18F Cloud Foundry documentation" could be confusing since cloud.gov isn't just Cloud Foundry, so I updated those bits. I also copyedited this a tiny bit.


jeremiak:
👍

PR #356, opened 2016-08-02
Attempt at better info about logging in

brittag:
We get questions from GSA/EPA people confused about why their password isn't working on the command line, so here's an effort at clarifying this system in the docs.


berndverst:
LGTM

PR #359, opened 2016-08-04
Adding context to custom domains page

brittag:
Moved Managed Service Method to the bottom (no text changes to that section), and put an intro on top of the page. The intro notes that the Managed Service Method is not generally available to users yet.


content/apps/custom-domains.md
```diff
@@ -145,6 +111,46 @@ Rather than create the `parent.tld` domain within CF we can create the full `app
 
 [API](http://apidocs.cloudfoundry.org/206/routes/associate_app_with_the_route.html)
 
----
 
-See also: [Official Cloud Foundry Documentation](https://docs.cloudfoundry.org/devguide/deploy-apps/routes-domains.html).
+## Managed Service Method
```
afeld, line 115: I'd reiterate here:

> **_GovCloud only.**_

brittag, line 115: OK! Added a note under the heading (rather than to the heading), so that we have a relatively stable anchor link for this heading.

afeld, line 115: Right, yep, that's what I meant.

PR #362, opened 2016-08-08
Add notes to IR retro section about cg-postmortems

brittag:
The [IR-4 control](https://web.nvd.nist.gov/view/800-53/Rev4/control?controlName=IR-4) requires a couple things:
- "Coordinates incident handling activities with contingency planning activities"
- "Incorporates lessons learned from ongoing incident handling activities into incident response procedures, training, and testing, and implements the resulting changes accordingly"

...and we already have some of that at https://github.com/18F/cg-postmortems ! So this IR guide should at least cross-link to that repo and suggest using it if appropriate. :)

Note that this is a (minor-ish) change to the IR process that we discussed in our training, so we should make sure that the cloud-gov-team on Slack is ok with (and aware of) this change.

### DIFF
--- a/content/ops/security-ir.md
+++ b/content/ops/security-ir.md
@@ -171,7 +171,7 @@ Once the incident is no longer active — i.e. the breach has been contained, th
 
 The final step in handling a security incident is figuring out what we learned. The IC (or one of the ICs if there were multiple, or a designated other party) should lead a retrospective and develop an incident report. 
 
-Conducting retrospectives is out of the scope of this document, but as a crash course, here's an [introduction to blameless postmortems](https://codeascraft.com/2012/05/22/blameless-postmortems/).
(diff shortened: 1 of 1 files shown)


brittag:
cc @jacobian since your opinions on this would be welcome!

jacobian:
Looks good! In the future it'd be good to do some documentation/training around holding retrospectives, but that's a yak shave and probably doesn't have to happen now. I'm happy to ✋  to work on that at some point, if y'all like.

brittag:
@jacobian Thank you, and yess I bet all of us would appreciate that! We need to figure out an overall cloud.gov training schedule anyway (for regular/recurring IR and contingency planning training and exercises), so I'll make an overall issue for that in cg-product to keep track of the idea.

Issue #328, opened 2016-07-15
Continuous deployment page should recommend best practices for zero-downtime deployment


brittag:
In order to set up great digital services on cloud.gov, cloud.gov users need good instructions for how to set up zero-downtime deployment (blue-green deployment). Currently our [continuous deployment page](https://docs.cloud.gov/apps/continuous-deployment/) has a little bit of info but should have a stronger opinion on what to do and link to appropriate further info.

Reference: [notes from Aidan's training](https://docs.google.com/document/d/1N0VQf5y0ZeO-9BS0Xk1W2p-YetF7HUqzSB4D-Kj7MXw/edit#heading=h.3vjjn2otsjs4).

Basics:
- Be more cautious about recommending `cf-blue-green`
- Provide a little more context
- Link to useful resources
- Link to a good example of an 18F product that is set up to do zero-downtime deployment

brittag:
Noting that @jezhumble knows ~a few things~ about this and offered to help. :)

jessieay:
Some ideas:
-  it seems like `autopilot` is the more popular choice (vs `cf-blue-green`) for zero-downtime deployment (at least that is what we moved to on C2 https://github.com/18F/C2/pull/1112)
- discuss the need to manage environment variables if you are doing a zero-downtime deploy. If all `ENV` vars are defined in your `manifest.yml` file, then you are fine. But if not (and it is likely your case is _not_ because frequently ENV vars are for sensitive keys and you don't want to check those into version control), you need to find a way to let _both_ your app and your second app (that is spun up during zero-downtime deploy) know about `ENV` vars. 
- it seem that the most popular way to deal with env vars is to set them via user-provided services, which are available to all apps in the space in which they are installed. We have some good docs on how we deal with user-provided services and ENV vars for Micropurchase: https://github.com/18F/micropurchase/blob/develop/docs/deployment.md#setting-environment-variables-on-staging-or-production
- right now on Micropurchase we are converting user-provided services into ENV vars via https://github.com/18F/cf-env-parser-buildpack but this is problematic because we can only use that buildpack if we also use https://github.com/ddollar/heroku-buildpack-multi, which is deprecated.
- the solution I am going to look into now is to use this setup https://github.com/18F/acqstackdb/blob/develop/.travis.yml#L26-L33 which takes advantage of https://github.com/18F/18f-cli/blob/develop/scripts/deploy for converting user-provided service instance keys into ENV vars
[...]

PR #365, opened 2016-08-16
Make some updates production ready best practices

LinuxBozo:
Updates based on review with C2 team on their deployment process and issues raised in #328. 

TODO: Should also consider linking into before-you-ship

/cc @afeld @brittag

### DIFF
--- a/content/apps/production-ready.md
+++ b/content/apps/production-ready.md
@@ -62,7 +62,7 @@ Your application should be able to be deployed without generating any downtime.
 To reduce the risk associated with manual deployments, consider automating the process so it is repeatable.
 
 #### How
-* See the [continuous deployment]({{< relref "continous-deployment.md" >}}) page.
+* See the [continuous deployment]({{< relref "continuous-deployment.md" >}}) page.
 
 ### Caching
 The best way to prevent performance issues is by having caching enabled on your
--- a/content/apps/production-ready.md
+++ b/content/apps/production-ready.md
@@ -8,9 +8,10 @@ title: Production Ready Guide
 
 # Your guide to having production ready apps on cloud.gov
 
-Read this guide early and often, especially when you’re starting to consider a future project. It explains simple things you can do for reliable and responsive applications deployed on cloud.gov.
+Read this guide early and often, especially when you’re starting to consider a future project. It explains things you can do for reliable and responsive applications deployed on cloud.gov.
 
-## Basic best practices
+## Core best practices
+To build consistent, healthy, production-ready applications on cloud.gov incorporate the following practices into your development workflow from the beginning.
 
 ### Configuration as code
 To ensure consistency and reproducibility, you should capture your application configuration in version control.
@@ -20,7 +21,7 @@ To ensure consistency and reproducibility, you should capture your application c
 
 ### More than one instance
 It is critical that your production application has more than one instance so if
-there any issue with one of the platform runners where your app is assigned, your app continues to function correctly.
+there any issue with one of the platform runners where your app instances are assigned, your app continues to function correctly.
 
 #### How
 * See the [multiple instances]({{< relref "multiple-instances.md" >}}) page.
@@ -50,10 +51,11 @@ You want to be able to receive alerts about application errors, downtime, and th
 * NewRelic provides uptime monitoring with "Insights". It is easy to set up and
 receive alerts on a variety of metrics.
 
-## Other considerations
+## Additional practices
+The following practices are very helpful to incorporate into most cloud.gov apps. Check these out and evaluate which ones you need, depending on the needs of your team and users.
 
 ### Zero-Downtime deploy
-Your application should be able to be deployed without generating any downtime.
+Your application should be able to be deployed without generating any downtime. Be aware there are known issues if your application automatically does database migrations when deploying.
 
 #### How
 * Use the [autopilot](https://github.com/concourse/autopilot) Cloud Foundry CLI plugin or the [`cf-blue-green`](https://github.com/18F/cf-blue-green) tool; see instructions in the respective repositories.
@@ -65,8 +67,7 @@ To reduce the risk associated with manual deployments, consider automating the p
 * See the [continuous deployment]({{< relref "continuous-deployment.md" >}}) page.
 
 ### Caching
-The best way to prevent performance issues is by having caching enabled on your
-application.
+The best way to prevent performance issues is by having caching enabled on your application.
 
 #### How
 * Cloud.gov has a memcached service but you can also rely on S3 or file storage for caches.
--- a/content/apps/production-ready.md
+++ b/content/apps/production-ready.md
@@ -2,54 +2,83 @@
 date: 2015-08-28T10:32:59-04:00
 menu:
   main:
-    parent: advanced
-title: Production Ready Checklist
+    parent: apps
+title: Production Ready Guide
 ---
 
-## Zero-Downtime deploy
+# Your guide to having production ready apps on cloud.gov
 
-Your application should be able to be deployed without generating any downtime.
-Check out the [cf-blue-green](https://github.com/18F/cf-blue-green) script as a
-sample script.
+Read this guide early and often, especially when you’re starting to consider a future project. It explains simple things you can do for reliable and responsive applications deployed on cloud.gov.
 
-## More than one instance
+## Basic best practices
 
+### Configuration as code
+To ensure consistency and reproducibility, you should capture your application configuration in version control.
+
+#### How
+* Write an [application manifest](https://docs.cloudfoundry.org/devguide/deploy-apps/manifest.html) that defines your application configuration, such as the application name, or what previously created services to use.
+
+### More than one instance
 It is critical that your production application has more than one instance so if
-there any issue with one of the runners your app is on your app continues to work.
+there any issue with one of the platform runners where your app is assigned, your app continues to function correctly.
 
-### How
+#### How
+* See the [multiple instances]({{< relref "multiple-instances.md" >}}) page.
 
-See the [multiple instances]({{< relref "multiple-instances.md" >}}) page.
+### Prefer User Provided Services
+Prefer user provided services over environment variables for a secure centralized way to store application credentials and variables. Environment variables are ephemeral to the specific deployment of each application.
 
-## Prefer User Provided Services
+#### How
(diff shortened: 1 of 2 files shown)


dogweather:
Lots of very useful information here — I'm really happy to see it. cc: @jessieay

content/apps/production-ready.md
```diff
@@ -2,54 +2,83 @@
...
+title: Production Ready Guide
 ---
 
-## Zero-Downtime deploy
+# Your guide to having production ready apps on cloud.gov
 
-Your application should be able to be deployed without generating any downtime.
-Check out the [cf-blue-green](https://github.com/18F/cf-blue-green) script as a
-sample script.
+Read this guide early and often, especially when you’re starting to consider a future project. It explains simple things you can do for reliable and responsive applications deployed on cloud.gov.
```
brittag: A suggestion from a reader-centered perspective: I'd remove "simple" here. I've learned to avoid saying things are "simple" or "easy" or "just do x" in documentation, since some readers then look at the page and think "uhhh these things don't look simple to me" and unnecessarily feel concerned about their level of knowledge. :)

brittag:
Aw yeah, thank you @LinuxBozo for doing this! I've checked this over, and this looks very helpful and necessary.

I like that this page is now split into "Basic best practices" and "Other considerations" - can we add a context sentence under each of those headings to give even more of an informative opinion?  Right now as a reader I might not really be sure how much I should pay attention to "Other considerations". For example, this is my informal impression of what these sections really mean:
- **Core best practices**: "To build a functional production-ready application on cloud.gov, incorporate the following practices from into your work from the beginning."
- **Additional best practices**: "The following practices are very helpful to incorporate into most cloud.gov apps. Check these out and evaluate which ones you need, depending on the needs of your team and users."

...but you'll have a better idea of how to update the titles and descriptions!

LinuxBozo:
@brittag good points on the headings and being a bit more explicit. Have updated accordingly based on your suggested text, which is excellent (as always).

content/apps/production-ready.md
```diff
@@ -2,54 +2,84 @@
...
+## Additional practices
+The following practices are very helpful to incorporate into most cloud.gov apps. Check these out and evaluate which ones you need, depending on the needs of your team and users.
 
-## NewRelic (or similar)
+### Zero-Downtime deploy
+Your application should be able to be deployed without generating any downtime. Be aware there are known issues if your application automatically does database migrations when deploying.
 
-You want to be able to receive alerts about application errors and throughput issues.
+#### How
+* Use the [autopilot](https://github.com/concourse/autopilot) Cloud Foundry CLI plugin or the [`cf-blue-green`](https://github.com/18F/cf-blue-green) tool; see instructions in the respective repositories.
```
jessieay: someone mentioned that we should be more cautious about recommending `cf-blue-green` [here](https://github.com/18F/cg-docs/issues/328) but I am not sure why...

dogweather: I'm wondering whether we should recommend either, tbh. Aren't C2's current problems due to using autopilot? 

I think our rule of thumb should be: if we have an 18F team successfully, safely using a technique, then we can include it as a recommended practice.

jessieay: Micropurchase is successfully using autopilot. We are doing it in  different way than C2, we are using it conjunction with user-provided services. We have been using it successfully for ~6 months.

The only problem we had with that was that we were previously parsing the user-provided services with a cloud foundry buildpack that required another deprecated buildpack. So, we moved to parsing those env vars manually. For how we do that, see [this PR](https://github.com/18F/micropurchase/pull/999/files)

dogweather: Are you using manifests, or are you downloading them dynamically from cloud.gov with `cf create-app-manifest`? (That's the core of C2's problem.)

jessieay: we are using version-controlled manifests that only contain non-sensitive keys. check out the repo for more! (we deploy via Travis)

jmcarp: AFAIK, the only reason to use cf-blue-green over autopilot is if the application doesn't have a proper manifest. Since we're already recommending that relevant configuration lives in the manifest, I don't think we need to recommend cf-blue-green. I also think that using `create-app-manifest` during deployment should be avoided if possible and doesn't belong in a best practices guide.

dogweather: Yes yes yes! I'll check out your repo! @jessieay 🏃  💨

brittag:
Not a blocker for integrating this improvement, but something I just noticed: should https://pages.18f.gov/before-you-ship/infrastructure/good-production-practices/ link here? Should that page be annotated with which items are taken care of if you follow this docs.cloud.gov page? :thinking_face:

LinuxBozo:
@brittag Great minds.. if you'll notice, that was a TODO I mentioned at the top ;-)

content/apps/production-ready.md
```diff
@@ -2,54 +2,84 @@
...
+# Your guide to having production ready apps on cloud.gov
 
-Your application should be able to be deployed without generating any downtime.
-Check out the [cf-blue-green](https://github.com/18F/cf-blue-green) script as a
-sample script.
+Read this guide early and often, especially when you’re starting to consider a future project. It explains things you can do for reliable and responsive applications deployed on cloud.gov.
 
-## More than one instance
+## Core best practices
+To build consistent, healthy, production-ready applications on cloud.gov incorporate the following practices into your development workflow from the beginning.
```
afeld: Minor: I would prefix these lines with `Goal:`.

afeld: Sorry, I misread it. Just needs a comma after "cloud.gov".

content/apps/production-ready.md
```diff
@@ -2,54 +2,84 @@
...
+### More than one instance
 It is critical that your production application has more than one instance so if
-there any issue with one of the runners your app is on your app continues to work.
+there any issue with one of the platform runners where your app instances are assigned, your app continues to function correctly.
+
+#### How
+* See the [multiple instances]({{< relref "multiple-instances.md" >}}) page.
+
+### Prefer User Provided Services
+Prefer user provided services over environment variables for a secure centralized way to store application credentials and variables. Environment variables are ephemeral to the specific deployment of each application.
```
afeld: I'd go a step further and say "anything sensitive should be in a user-provided service instance, anything non-sensitive should be in the manifest under [`env:`](https://docs.cloudfoundry.org/devguide/deploy-apps/manifest.html#env-block)."

content/apps/production-ready.md
```diff
@@ -2,54 +2,84 @@
...
+### Prefer dedicated over shared services
+Shared services do not have the same constraints as a dedicated service and can be affected by other customers that are performing CPU or memory intensive tasks against that service.
 
-### How
+#### How
+* Use the `cf marketplace` to consider your options and sizes for each service and choose appropriately for the resources your application will need.
 
-See the [multiple instances]({{< relref "multiple-instances.md" >}}) page.
+### Space per environment
+When running an application for development or testing, it is best to have a separate space from your production instance in which to run the application.
```
afeld: I'd add "ideally, each space should look identical, aside from the routes and perhaps the number of instances".

content/apps/production-ready.md
```diff
@@ -2,54 +2,84 @@
...
+#### How
+* As an Org Manager for your organization, use the `cf create-space` command to create new spaces for each environment
 
-Prefer user provided services over environment variables for a secure centralized way to store application credentials and variables.
+### Health monitoring
+You want to be able to receive alerts about application errors, downtime, and throughput issues.
 
-### How
+#### How
+* NewRelic provides uptime monitoring with "Insights". It is easy to set up and
```
afeld: I think you mean Synthetics.

content/apps/production-ready.md
afeld: As the author of cf-blue-green, I concur with @jmcarp about not even mentioning it on the "best practices" page 😉

content/apps/production-ready.md
```diff
@@ -2,54 +2,84 @@
...
+* Cloud.gov has a memcached service but you can also rely on S3 or file storage for caches.
 
-## Asset Serving
+### Asset Serving
+It is best not to serve static files from Cloud.gov directly.
 
-It is best not to serve static files from Cloud.gov directly. You can store your
-files in S3 or point CloudFront to an assets folder so you serve your assets with a CDN.
+#### How
+* You can store your files in S3 or point CloudFront to an assets folder so you serve your assets with a CDN.
```
afeld, line 78: Do we actually have anyone in 18F serving assets from S3? Since we're offering CloudFront in GovCloud, I'd actually remove this recommendation in favor of saying "use our CloudFront CDN service (if you're in GovCloud), and make sure your application sets the appropriate [caching headers](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching?hl=en)."

afeld:
Great work on this! 👏

content/apps/production-ready.md
LinuxBozo, line 78: Not sure about S3, but we do provide instructions for setting up S3 as a service, and how to modify the policy to make it available outside of cloud.gov, so will link there. As for our CloudFront service, I don't believe we actually are using the CDN features when setting up that service. @jmcarp?

content/apps/production-ready.md
LinuxBozo: Insights is the higher level product that includes Application performance, and Synthetics unless I'm mistaken.

afeld: I think Synthetics can feed into Insights, but they are separate.

New Relic nav bar

content/apps/production-ready.md
dogweather, line 78: I agree with @afeld 's sentiment: avoid the extra complexity of an asset server, and let the CDN optimize it.

brittag:
I'd suggest to merge this and keep working on the CDN bit in another PR, since this is already much better than the existing page overall. :)

content/apps/production-ready.md
jseppi: Per [discussion in Slack](https://18f.slack.com/archives/dev/p1472140610002098), it sounds like the best practice here is _not_ to use S3 or a CDN _unless_ you have huge traffic concerns. So maybe that first sentence in this section should be something along the lines of:

"If your application is expected to receive [THRESHOLD # OF REQUESTS], then you should use the CloudFront CDN to serve your static assets"

and the "How" section should include a link to https://docs.cloud.gov/apps/custom-domains/#managed-service-method (with the caveat that this isn't available until the migration to GovCloud is complete).

PR #373, opened 2016-08-19
Improve 'leveraging auth' section.

### LINKED ISSUE #321: Improve leveraging-authentication.md
I used the code sample at the bottom of [Leveraging Authentication](https://docs.cloud.gov/apps/leveraging-authentication/) to make a trivial OAuth client recently and noticed a few issues with it:
- The snippet assumes that `request` refers to node's [`request`](https://github.com/request/request) module but never mentions it.
- The snippet assumes that `app` is an [`express`](https://expressjs.com/) instance, but never mentions it.

Aside from that, the snippet isn't actually self-contained in any way, as it uses other constants that aren't defined anywhere, and doesn't provide a login link; most documentation that offers snippets also provides a "full code listing" that readers can copy-paste into an actual editor and get running with minimal effort. This also makes it difficult to double-check whether the snippet (or any changes we make down the road) is actually correct.

Rather than fix the snippet, though, I recommend just linking out to a fully self-contained UAA client.  I've taken the one defined in the snippet and added minimal extra code (~80 total LOC) and a `package.json` to install external dependencies here:

https://github.com/18F/cg-fake-uaa/tree/master/example-client

So we could potentially just link to that instead of providing the snippet.

Besides that, I'd also like to link to [cg-fake-uaa](https://github.com/18F/cg-fake-uaa) as a resource that might help folks integrate with cloud.gov authentication.

Oh, and I'd also like to link to [The Importance of the state parameter in OAuth2](http://www.twobotechnologies.com/blog/2014/02/importance-of-state-in-oauth2.html) when discussing the `state` parameter, to provide some context around it and nudge people to use it (depending on what our compliance folks think, we might even want to just mandate its use, but I'm not sure).

Thoughts @brittag @jseppi ?

toolness:
This fixes #321:
- Converts some of the workarounds to https://github.com/18F/web-design-standards/issues/1328 into nested lists, as that USWDS issue has been resolved.
- Links to [The Importance of the state parameter in OAuth2](http://www.twobotechnologies.com/blog/2014/02/importance-of-state-in-oauth2.html).
- Mentions [cg-fake-uaa](https://github.com/18F/cg-fake-uaa).
- Replaces the incomplete code snippet with a link to [cg-fake-uaa's example client](https://github.com/18F/cg-fake-uaa/tree/master/example-client).

### DIFF
--- a/content/apps/leveraging-authentication.md
+++ b/content/apps/leveraging-authentication.md
@@ -29,112 +29,84 @@ There are two important cloud.gov URLs you will need to use:
 - `https://login.cloud.gov/oauth/authorize`, which is where you will direct the user to login with their agency credentials
 - `https://uaa.cloud.gov/oauth/token`, which is where you will exchange auth codes for auth tokens
 
-If you are already familiar with OAuth 2.0, you might know where to go from here. If not, read on for a minimal example of how to do the OAuth dance. Here's the basic gist of it:
+If you are already familiar with OAuth 2.0, you might know where to go from here. If not, read on for a minimal example of how to do the OAuth dance. Below is the basic gist of it.
 
-`1.` Generate a link (or redirect the user) to the authorize URL with these query parameters:
+#### Send your user to login.cloud.gov
+
+First, generate a link (or redirect the user) to the authorize URL with these
+query parameters:
 
 * `client_id=<YOUR APP'S REGISTERED NAME>`
 * `response_type=code`
 * (optional) `state=<ANYTHING>`
 
 You can optionally set a `state` parameter with any value you'd like.
-It will be returned to you in a later step. An example of something you might
-want to set in state is the URL in your application that the user was attempting
-to access.
+It will be returned to you in a later step. While optional, it's highly
+recommended that you use it for [security reasons](http://www.twobotechnologies.com/blog/2014/02/importance-of-state-in-oauth2.html).
 
 Here is an example:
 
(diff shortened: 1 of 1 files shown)


content/apps/leveraging-authentication.md
```diff
@@ -29,112 +29,84 @@ There are two important cloud.gov URLs you will need to use:
...
+    to get the authenticated user's `email`, which you can then use within
+    your application to identify and/or authorize the user.
+
+    If you get an expired token error at some point in the future, you can
+    exchange the `refresh_token` from the previous step to get a new `access_token`,
+    so you might want to securely save the `refresh_token` associated with the
+    authenticated user.
+
+For an example of this entire process in under 100 lines of node JS, see the
+[example client](https://github.com/18F/cg-fake-uaa/tree/master/example-client).
```
afeld, line 101: Thoughts about moving all of these instructions to a README there?

afeld:
I'm not able to take the time to verify these instructions right now, but would be great to have someone else run through them at _some_ point. Thanks for doing this!

afeld:
Merging now, but maybe we can take care of the comment above separately.

toolness:
Thanks Aidan!  I actually changed very little of the copy aside from the formatting and indentation... Perhaps I should have done this as two separate PRs, one that just changed the formatting/indentation and another that changed the bits of content.  Ah well.

PR #378, opened 2016-08-23
add information about managing quotas

afeld:
/cc @jeremiak


jeremiak:
Much obliged for your help with this, @afeld. I added a link to your new content from the S3 page that I found my original instructions on.

PR #381, opened 2016-08-25
Another way to add info about SSH in E/W and GovCloud

brittag:
This is an experiment based on the info at https://github.com/18F/cg-docs/pull/379 and https://github.com/18F/cg-docs/pull/380. Instead of removing the E/W info (like 379) or fully duplicating the sections (like 380), I tried just duplicating the sections that changed, since I suspect this might be easier for readers to understand (and for us to maintain).

cc @sushiandbeer - I'm curious if this makes sense to you.

### DIFF
--- a/content/apps/databases.md
+++ b/content/apps/databases.md
@@ -5,7 +5,7 @@ menu:
 title: Databases
 ---
 
-There are multiple options for how to set up databases for use with applications in Cloud Foundry.
+There are multiple options for how to set up databases for use with applications in cloud.gov's version of Cloud Foundry.
 
 ## Services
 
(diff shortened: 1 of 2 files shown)


brittag:
Noting that this has a 👍  from Mike (https://18f.slack.com/archives/cloud-gov-atlas/p1472088215002383); since we both worked on this, we'd like a third person to review this and decide whether to merge it.

ghost:
https://github.com/18F/cg-product/issues/156 blocked waiting for merge

content/apps/databases.md
```diff
@@ -63,10 +67,25 @@ To access a service database, use the [cf-ssh]({{< relref "getting-started/one-o
 
 You should now have an open `psql` terminal connected to the service database.
 
+### *GovCloud environment:* Using cf ssh
+
+To access a service database, use the [cf ssh]({{< relref "getting-started/one-off-tasks.md#cf-ssh" >}}) CLI command to login to an instance that is bound to the service and download the `psql` binary to that instance:
+
+    cf ssh APP_NAME
+    curl https://s3.amazonaws.com/18f-cf-cli/psql-9.4.4-ubuntu-14.04.tar.gz | tar xvz
+    ./psql/bin/psql $DATABASE_URL
```
afeld, line 76: Not a blocker, but we might consider recommending [tunneling](https://docs.cloudfoundry.org/devguide/deploy-apps/ssh-services.html#ssh-tunnel) instead so they don't have to install anything extra in the container.

mogul: It's a bigger topic... Even if they tunnel they have to have the client _somewhere_. Given most agency laptops/desktops are locked down pretty tight, they might actually have to stick with the client-installed-in-container route. There's also the potential for [using scp](https://docs.cloudfoundry.org/devguide/deploy-apps/ssh-apps.html#other-ssh-access) to actually get files dump files in and out. For future!

PR #397, opened 2016-09-07
Fix paid services link in S3.md

bmcguirk:
Link was hard-coded to `http://localhost:1313/apps/managed-services/#paid-services`.


afeld:
Good catch, thanks!

PR #398, opened 2016-09-08
remove outdated information about creating RDS instances by hand

afeld:


mogul:
lgtm, but needs to be brought up to date before merging...

PR #401, opened 2016-09-09
Stop posting notifications in liberator channel

msecret:


juliasolorzano:
Just to clarify, this setting for making concourse alerts show up in our news channel - correct?

PR #404, opened 2016-09-14
improvement to explanations

NoahKunin:
added more plain language even when redundant with special govt finance terms, changes to reflect invoicing improvements.


NoahKunin:
There's one editorial note, which is "instantly" is probably a bridge too far, unless we have fine print that says "instant" means 1s or less. 😁

mogul:
LGTM

PR #409, opened 2016-09-18
remove redundant `port` variable assignment

### LINKED ISSUE #407: Flask docs error
Hello all,

the following code:

```
port = port = int(os.getenv("VCAP_APP_PORT"))
```

exists at https://docs.cloud.gov/apps/flask/

I _believe_ it should be:

```
port = int(os.getenv("VCAP_APP_PORT"))
```

afeld:
Fixes #407.

/cc @EricSchles


jmcarp:
`VCAP_APP_PORT` is deprecated (but still works). While we're editing this, let's change to `PORT`.

jmcarp:
Ack--we actually don't recommend using waitress for flask or django either. But we can fix that in another patch.

cc @cmc333333

PR #411, opened 2016-09-20
Refer to a stable multi-buildpack fork which won't expire

jpluscplusm:
The ddollar multi-buildpack is unmaintained, and has had a kill switch inserted so that any app using it, restaged or pushed after 1 Jan 2017, will break.

I've created a fork of this buildpack, specifically aimed at CF usage, which will be (a) stable and (b) actively developed.

This PR updates your docs to point towards the new fork. Obviously you don't have to do so, but there's a definite drop-dead date you should be aware of, regardless!

### DIFF
--- a/content/apps/multi-buildpack-deploys.md
+++ b/content/apps/multi-buildpack-deploys.md
@@ -6,20 +6,20 @@ menu:
 title: Multi-Language Projects
 ---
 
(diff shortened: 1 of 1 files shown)


afeld:
We're considering dropping mention of the multi buildpack altogether (for various reasons), but good to know about that fork in the meantime. Thanks!

mogul:
Just to follow-up: We moved mention of the multi-buildpack to the "custom buildpacks" page in our new (explicitly unsupported) experimental features section. Thanks again for the note!

PR #415, opened 2016-09-23
Correct communication of resource costs

mogul:
We've been implying that resource usage is free in the chart! This attempts to communicate the differences between resource usage and managed services precisely. /cc @dlapiduz


mogul:
@dlapiduz can you confirm that you saw this and agree with the presentation?

nikzei:
Also see https://github.com/18F/cg-product/issues/299, which is a reference to this as well.

dlapiduz:
looks good, thanks!

PR #421, opened 2016-09-26
Be explict that DNSSEC is a customer responsibility

cnelson:
Do we need to say anything else here to meet the SSP needs mentioned in https://github.com/18F/cg-product/issues/314 ?


mogul:
@NoahKunin, I defer to you on all matters DNSSEC.

PR #431, opened 2016-09-27
DO NOT MERGE Add guidance for running Docker Registry

NoahKunin:


mogul:
This PR is going to concern customer advice around pushing their own Docker containers, with regard to reducing their ATO burden. It looks like we can advise them with this thumbnail sketch:
1. Provision an S3 service instance, then
2. `cf push` [the official Docker registry](https://docs.docker.com/registry/deploying/)...
   -  ...[configured to use the S3 service](https://docs.docker.com/registry/storage-drivers/) and 
   - ...[restricting access](https://docs.docker.com/registry/deploying/#/restricting-access) (eg like our UAA proxy, but scoped to only accept users with a SpaceDeveloper role in the space)
3. Then [use that as their "trusted" registry](http://docs.cloudfoundry.org/adminguide/docker.html#private) for managing the images they push into the space

mogul:
More thoughts on how to secure this... We want two routes to the registry:
- One route is for humans, allows pull and push, and is secured by a UAA Proxy route-service that limits access to SpaceDevelopers
- One route is for robots, allows only pull, and is secured by a ??? route-service that limits access to "appropriate clients". 

The specific way you identify "appropriate clients" needs work... 
- Is it the set of IPs associated with all runners in cloud.gov? That would be fine for a platform-level registry, but is overbroad for a private registry
- Is it the set of clients presenting a certain cert? That would probably be good. If we broker the setup of the registry, the broker can also be responsible for providing a client cert and secret, visible to the SpaceDevelopers, that they can incorporate... elsewhere.

/cc @jezhumble @dlapiduz

mogul:
Building on the sketch above, there's now a bunch of documentation on [how Federalist hosts their Docker registry in cloud.gov](https://federalist-docs.18f.gov/pages/how-federalist-works/cloud-gov-setup/).

jmcarp:
Is work still happening here @mogul ?

mogul:
No, but I kinda want the docs... **somewhere**.

jmcarp:
This has both merge conflicts and no real content. Let's move to favro if we want to work on it.

PR #433, opened 2016-09-28
Moving custom buildpack info to experiments section

brittag:
This moves info about [apt-buildpack](https://docs.cloud.gov/apps/apt-get/) and [buildpack-multi](https://docs.cloud.gov/apps/multi-buildpack-deploys/) into our experiments section (without breaking their URLs).

I reframed the buildpack-multi page to start with clearly-marked supported recommendations and then explain the experimental option. Please fact-check this change to make sure I didn't introduce errors. :D

This helps with https://github.com/18F/cg-docs/issues/248, although it doesn't fully resolve the problem.

I also removed a link to https://github.com/ozzyjohnson/heroku-buildpack-multi since that seemed unmaintained.

### DIFF
--- a/content/apps/apt-get.md
+++ b/content/apps/apt-get.md
@@ -1,10 +1,14 @@
 ---
 menu:
   main:
-    parent: apps
+    parent: experimental
 title: Using apt-get
 ---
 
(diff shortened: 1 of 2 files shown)


afeld (approved):
I'm not able to see why the build failed (:cry:), but looks good to me!

PR #439, opened 2016-09-29
Use official node image.

### LINKED ISSUE #437: Fix build errors
Concourse builds fail fairly often with `No space left on device` errors during `apk add`, even though the container appears to have plenty of space available. I've also been able to `intercept` failed builds and manually run `apk add` without errors.

Help @sharms !

jmcarp:
[Resolves #437]

I know alpine is great and all, but building with the official image is still fast, and there should be no difference in build time once the image is cached. If @LinuxBozo or @afeld think this is dumb, I'll make a new image from alpine-node that includes requirements for gyp instead.

### DIFF
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -3,18 +3,10 @@
 set -e
 set -x
 
-# install dependencies, for the following (respectively): NPM, node-sass, Hugo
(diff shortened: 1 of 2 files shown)


LinuxBozo (approved):
:shipit:

PR #444, opened 2016-09-30
add policy around Fugacious use

afeld:
Fixes https://github.com/18F/handbook/pull/98#discussion_r80753434. Fixes https://github.com/18F/cg-product/issues/312.

Paired with @brittag on this. Per https://github.com/18F/cg-product/issues/312#issuecomment-249958937:

> one of my foundational compliance values is we turn back the tide on government making rules (either for others our ourselves) that we cannot enforce

Not quite sure what we should/shouldn't say in here in order to stay in line with this.

### DIFF
--- a/content/ops/secrets.md
+++ b/content/ops/secrets.md
@@ -3,7 +3,6 @@ menu:
   main:
     parent: operations
 title: Secret Key Management
-weight: 10
 ---
 
 ### Sharing secret keys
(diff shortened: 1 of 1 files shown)


mogul:
Any reason Noah has to be the merger? It looks fine to me and I'm happy to do the honors, but if there's something you're looking for from him specifically I don't want to jump the gun.

afeld:
We can certainly merge...I just wanted to make sure it covered what we needed for addressing the JAB feedback, and he seemed to have specific ideas about that.

PR #455, opened 2016-10-06
updating django docs

EricSchles:
@dlapiduz and I found a weird thing, that he was able to quickly resolve.  I am adding this to the docs.

### DIFF
--- a/content/apps/django.md
+++ b/content/apps/django.md
@@ -155,7 +155,7 @@ You may want to run `manage.py` commands on your app to perform maintenance task
 
 ### Gotchas
 
-If you're running the `python manage.py` commands from the server, you might have to do:
+If you're running the `python manage.py` commands from the server, and you're in the gov-cloud space you'll have to do:
 
 ```
 cd ~/app
@@ -163,4 +163,4 @@ source .profile.d/python.sh
 .heroku/python/bin/python
 ```
 
-An easy way to check this is to `cf ssh [APPNAME]` and run python -c "import django" 
+Otherwise python won't be aware of any of your libraries.  An easy way to check this is to `cf ssh [APPNAME]` and run `python -c "import django"`.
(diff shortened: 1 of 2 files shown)


brittag:
Nice, thanks for adding this! This has even [already helped somebody else](https://gsa-tts.slack.com/archives/cloud-gov-support/p1475871459001627).

Can you add a tiny bit of context to this tip - how would another person know if they should do this? For example, did you get a particular kind of error message that was resolved by doing this? (Can we include that?) And it sounds like this only happens in GovCloud, so I suggest adding a note saying that this only applies to GovCloud.

If the real situation here is "if you have a Django app in GovCloud, you'll need to do this before anything works", I suggest putting that in. :D

EricSchles:
Ack @brittag sorry about the delay here - my github notifications are being not great.

mogul:
Can you split the bit about creating spaces into a separate PR? I'm pretty sure that's not the right place for that kind of information as it's operator- rather than user-facing.

EricSchles:
For sure @mogul how would you like me to split this out - sorry for putting this in the wrong place!  Please feel free to suggest the right place

cnelson:
@EricSchles Can you just revert/remove your change to `content/ops/managing-users.md` and I think we can accept this PR.

We already have documentation on creating spaces in our [concepts](https://github.com/18F/cg-site/blob/master/content/docs/getting-started/concepts.md#spaces) page.

EricSchles:
Thanks @cnelson Will get on this asap

stvnrlly:
I was just using `cf ssh` again, and found that it's necessary to run `export HOME=/home/vcap/app` before `source .profile.d/python.sh`. If not, the `PATH` won't be set properly.

That also accords with what [the Cloud Foundry docs](https://docs.cloudfoundry.org/devguide/deploy-apps/ssh-apps.html#ssh-env) say to do. As a bonus, those instructions _should_ work for any language/framework, I think.

cnelson:
Do we still need these changes after @mogul's change ( https://github.com/18F/cg-site/commit/dd86e8e2352ae1b8f2bbf02552aadc44892de75a ) to specifically link to the docs @stvnrlly mentions?  Can we close this PR?

mogul:
The `cf ssh` page seems like the better place to include this information due to wider audience.

PR #465, opened 2016-10-15
Update access information on GovCloud page

brittag:
This should be merged when we've confirmed that EPA people can log into the GovCloud environment - I'll leave that up to @cnelson.

I also ended up updating a few links in the intro material while looking for any other GSA/EPA info that should be updated.

### DIFF
--- a/content/apps/govcloud.md
+++ b/content/apps/govcloud.md
@@ -11,7 +11,7 @@ When your org starts using cloud.gov's GovCloud environment, here are changes to
 
 ### Breaking changes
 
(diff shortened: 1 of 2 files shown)


cnelson:
I'm hugo ignorant, but is there some way could include a template with the list of SSO enabled agencies in all the places that need it, so next time we add one we only need to update one well-known template?

mogul:
EPA is confirmed able to access now. 

Chris, Hugo supports partials if you're so inclined: http://gohugo.io/templates/partials/

cnelson (approved):

PR #471, opened 2016-10-19
Removing Backhub mention from Contingency Plan

brittag:
Since this is not currently part of our plan for handling a GitHub problem. cc @datn @NoahKunin


afeld:
Should we keep it in as a contingency for "in case we accidentally delete a repo"?

brittag:
@afeld Let's add it back in later for that if we have it enabled again later!

brittag:
Closing this, since removing this sentence will be handled by this new contingency plan PR: https://github.com/18F/cg-docs/pull/479

PR #472, opened 2016-10-20
Unify the website repos and use Hugo

berndverst:
Everything now uses relative (to the root) paths so that the website should load correctly during local development, staging deployment, or production release.

There are _no_ content changes. Git got a a bit confused with files I moved / renamed and then edited. So it looks like files are being deleted but then almost the same file is being edited. The only changes though are updating the links.


berndverst:
@cnelson feel free to push your CI changes to this branch and we'll include them in this PR. We'll need to replace the existing CI configuration here with the one from `cg-site`. I explicitly haven't done that yet since I wasn't sure how that affects your changes.

berndverst:
Feel free to review, but please
_DO NOT MERGE_ (without syncing with both @cnelson and @berndverst)

berndverst:
This is running in production at https://combodocs.fr.cloud.gov/

@brittag you can now do a quick review. Also note that once we merge this we can much more quickly fix any little things that come up.

It looks good to me, so unless there are any object I would like us to merge this by _3 pm on Monday_.

afeld:
- [x] The links for each post in https://combodocs.fr.cloud.gov/updates/ are broken
- [x] The `CONTACT` link is broken on https://combodocs.fr.cloud.gov/docs/ and https://combodocs.fr.cloud.gov/updates/

Looks good otherwise—thanks for putting in all of the effort!

berndverst:
@afeld can you give an example of broken links in `/updates` ? I cannot reproduce.

The links were working, but going to what is hosted on `cloud.gov`. Apparently using `.Permalink` uses the root defined in `config.toml` and makes absolutely paths. I updated this to `.RelPermalink` now.

Re-deployed the staging version. Please take a look.

berndverst:
Also needless to say, I will manually resolve the conflicts before merging any of this :)

afeld (approved):

PR #478, opened 2016-10-22
Necessary change to be in compliance with IR-9 (4)

mogul:


wslack (approved):
LGTM but not on core team

PR #481, opened 2016-10-25
Coordinate contingency plan and IR plan

brittag:
To fulfill the requirements of https://github.com/18F/cg-product/issues/398, we need to make our contingency plan and IR plan fit together better.

This change updates both plans (and the IR checklist) to have appropriate cross-references in several places.

I also updated some links.


mogul (approved):
LGTM!

PR #482, opened 2016-10-25
Add a custom nginx config to the site

cnelson:
The custom nginx.conf uses the cloud.gov name in all redirects so we do not bypass cloudfront.

It also enables Cache Control headers with a max age of 5 minutes to shorten the time it takes content changes to appear on the site after publishing.

### DIFF
--- a/ci/build.sh
+++ b/ci/build.sh
@@ -17,6 +17,9 @@ npm run build
 # build the site
 hugo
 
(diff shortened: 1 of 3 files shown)


brittag:
Yay, thanks! 👍  for the purposes of this PR; I'll leave the technical review to another person.

sharms (approved):

PR #494, opened 2016-10-26
Add nodejs sizing documentation

jbarnicle:
Documenting some issues we've observed when deploying nodejs apps in govcloud.
@rogeruiz 

Tied to the work from 18F/federalist#559


jbarnicle:
@rogeruiz - Ugh - the Diego documentation update snuck in with this one.  I can put that in a separate PR if you like.

content/docs/apps/node.md
```diff
@@ -21,6 +21,14 @@ Deploying a basic node.js application to Cloud Foundry is relatively uncomplicat
     ```
 
 - Avoid using the `command:` manifest key to start your application, as this functionality [may be removed](https://github.com/cloudfoundry/nodejs-buildpack/pull/11#issuecomment-67666273). The aforementioned npm start script is preferred.
+
+- Vendor your dependencies.  Cloud Foundry [documentation](https://docs.cloudfoundry.org/buildpacks/node/index.html#vendoring) specifies that you should do a `npm install` to install vendored dependencies and pushing your entire application with `cf push`. This is especially
+important for large applications as Cloud Foundry limits the size of an app upload to `1GB` -
```
rogeruiz: It's important to differentiate here between AWS E/W and GovCloud. Also, the limit is a default of `1GB` unless more space is given using the `disk_space: <Value>` property in the manifest. Maybe something like GovCloud application containers have a default disk limit of 1GB. But I'd verify this in `#cloud-gov-atlas` to be sure.

content/docs/apps/node.md
```diff
@@ -21,6 +21,14 @@ Deploying a basic node.js application to Cloud Foundry is relatively uncomplicat
...
+see [Deploying a Large Application](https://docs.cloudfoundry.org/devguide/deploy-apps/large-app-deploy.html) for more information.
+
+- Do not put test harnesses or transpilers in your dependencies object. View the npm
+[documentation](https://docs.npmjs.com/files/package.json#dependencies) for more on this.
```
rogeruiz: In this section it'd be good to include a section pointing out that `NODE_ENV` should always be set up to `production` in order to only have `dependencies: { /**/ }` be downloaded otherwise `devDependencies: { /**/ }` will also be downloaded which could affect disk space and memory quotas.

rogeruiz (changes requested):
It's okay to keep these in the same PR. NBD.

rogeruiz:
This PR is output from the `#cloud-gov-support` request that generated 18F/federalist#559.

jbarnicle (commented):
I think these have been resolved.

rogeruiz (approved):

rogeruiz:
Yep! Thanks @jbarnicle 🎉

PR #507, opened 2016-10-30
Use safest, simplest bucket URLs

konklone:
This recommends the use of the `https://s3.amazonaws.com` form of S3 URLs, without the region, and uses the HTTPS URL version of it.

This is simplest and safest because it doesn't depend on the region staying the same, and HTTPS can consistently be used whether or not the bucket name itself has `.`'s in it.

### DIFF
--- a/content/docs/apps/s3.md
+++ b/content/docs/apps/s3.md
@@ -91,9 +91,15 @@ aws s3 ls s3://${BUCKET_NAME}
 ## Bucket URLs
 Objects in your bucket can be accessed most reliably via the following endpoint:
 
-* `https://s3.amazonaws.com/${BUCKET_NAME}/`
+* `https://s3-${AWS_DEFAULT_REGION}.amazonaws.com/${BUCKET_NAME}/`
 
-If the bucket is private, attempting to access this endpoint will result in `AccessDenied` errors unless your application generates [pre-signed URLs](http://docs.aws.amazon.com/AmazonS3/latest/dev/ShareObjectPreSignedURL.html) for objects that need to be shared.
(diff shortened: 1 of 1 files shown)


dlapiduz:
@konklone there are a couple issues with this change: 
- If you don't specify the region then it assumes it is in the east/west space, not in GovCloud (which we use).
- Some people use the website endpoint since it has a bit of a different behavior than the main S3 endpoint.
- HTTPS (believe it or not) doesn't work well on the website endpoint.

That said we should recommend using https on the non website urls.

wslack:
I'd be happy with whatever set up allows use to alias federalist.18f.gov/site or federalist.18f.gov/preview to the buckets, as is the present case.

That said, I tested the other format suggested here and it didn't work; see here. @konklone, please LMK if something is off with these links:
- http://cg-f28e32aa-d42e-4906-a813-7d726f69183c.s3-website-us-gov-west-1.amazonaws.com/site/el-mapache/federalist-modern-team-template/
- https://s3.amazonaws.com/cg-f28e32aa-d42e-4906-a813-7d726f69183c/site/el-mapache/federalist-modern-team-template/

konklone:
Damn @dlapiduz, you're right, I didn't realize that:
- Works: https://s3-us-gov-west-1.amazonaws.com/cg-ff27f612-cd16-4a76-94f8-628986af2158/realtime.json
- Doesn't work: https://s3.amazonaws.com/cg-ff27f612-cd16-4a76-94f8-628986af2158/realtime.json

And yeah, you're right that if you're using the website endpoint it doesn't support HTTPS. I'll update the PR to be more flexible.

konklone:
@dlapiduz I made some changes to the PR.

dlapiduz (approved):

PR #508, opened 2016-10-31
Add open data plan

dlapiduz:


dlapiduz:
What do you think @berndverst @mogul @nikzei ?

mogul:
Looks fine to me; you double-checked it with a bunch of folks who are likely customers of it, right?

berndverst:
@dlapiduz I don't understand this.

The account is limited to 2 GB memory usage per month? Would those 2 GB be included?

Are you saying $10k fee and you get up to $2500 in resources which could be 2 GB of memory or other services?

dlapiduz:
@berndverst no, to make this clear I used the same language as all the other plans but added the "limit" to the plan description and the services fees.

berndverst (approved):

PR #522, opened 2016-11-02
Fix command for downloading a database backup

ccostino:
Hi cloud.gov folks!

This is a very minor change to fix the command for downloading a database backup.  When you're running an SSH session, you are in the `app` directory by default as a part of your `pwd` (present working directory) and therefore need to have `app/` in front of the file name to get to the database backup file.


jbarnicle (approved):
Thanks @ccostino

PR #525, opened 2016-11-03
specify Python version

afeld:
Python 2 required by node-gyp, which is required by node-sass, which is required by cg-style. :unamused:

https://github.com/nodejs/node-gyp#installation

/cc @msecret @adborden


msecret:
Not required by cg-style, required by bourbon/neat, required by WDS, required by cg-style.

adborden (approved):

PR #526, opened 2016-11-03
Overview content cleanup

jameshupp:
## Changes
- Getting PaaS and IaaS into the same style
- Reordering the why-use page to emphasize security and conforming with rules, as well as shifting how we describe the usability of the platform a little
- Simplified why-build to de-emphasize some of the self-focused portion
- Introduce TTS ownership

## Review
@berndverst 

## Notes
I’d like to emphasize cloud.gov even more vis-a-vis 18F, but there are a number of 18F URLs. That speaks to keeping our name somewhere on the product, even if it’s not forefronted.


content/overview/overview/what-is-cloudgov.md
```diff
@@ -2,29 +2,31 @@
...
+cloud.gov runs on top of industry-provided infrastructure ([currently Amazon Web Services is our “Infrastructure as a Service” provider]({{< relref "overview/technology/iaas.md" >}})). cloud.gov removes the complexity of managing this infrastructure from the customer’s experience.
 
 All the software components of the cloud.gov Platform as a Service are [open source]({{< relref "docs/ops/repos.md" >}}).
 
+### Why work with this Platform as a Service?
+
+cloud.gov gives teams working for federal government a secure, fully compliant foundation on which to build and release products and updates quickly. It handles the shared technical and legal requirements common to federal government systems, so that teams can skip a large amount of repetitive work and instead focus on delivering software that serves people’s needs. It’s built to keep applications online even with large numbers of users and sharp increases in usage.
```
brittag: This reminds me of a change I wanted to make to reflect our environment better: "legal requirements" -> "policy requirements". Up to you all, just figured I'd suggest it since this PR is open.

jameshupp: @brittag I think this makes more sense anyway. In plain understanding, I would say "policy" would be inclusive of "legal" while the opposite is not true.

content/overview/overview/why-available-other-agencies.md
```diff
@@ -5,22 +5,18 @@ menu:
...
 weight: 30
 aliases:
-  - /docs/intro/overview/why-available-other-agencies
-  - /intro/overview/why-available-other-agencies
+  - /overview/overview/why-available-other-agencies
+  - /overview/why-available-other-agencies
 ---
 
-Just as it has for us at 18F, we think cloud.gov will help other federal agencies deliver taxpayer-funded services the American people deserve in a faster, more user-centered way.
+cloud.gov will help other federal agencies deliver taxpayer-funded services the American people deserve in a faster, more user-centered way.
```
brittag: Suggest changing "cloud.gov will help other federal agencies" to "cloud.gov helps federal agencies".

brittag:
Looking at this, I'm not sure if I understand the `aliases` changes, such as from `/docs/intro/overview/building-process` to `/overview/overview/building-process` - I thought we needed the /docs/ aliases to preserve redirects from old URLs? But testing it locally, the old URLs don't seem to be broken - so basically I'm confused!

jameshupp:
@brittag The alias change was me misunderstanding a problem earlier. Like you said, they still seem to work. Not that hard to roll them back, though; there's only one other change in that commit, I think, and it's pretty minor.

brittag:
@jameshupp To minimize risk of unexpected side effects, I'd suggest rolling them back, so that this is just a content-update PR. We can update the aliases later if that seems to help with a problem.

jameshupp:
@brittag At long last, rolled back. Had to do a couple other little things but it should be good now.

content/overview/overview/what-is-cloudgov.md
```diff
@@ -2,29 +2,31 @@
 menu:
   overview:
     parent: overview
-title: What is cloud.gov?
+title: What Is cloud.gov?
```
brittag: Suggest sentence case here; ideally all our titles would be sentence case!

content/overview/overview/what-is-cloudgov.md
```diff
@@ -2,29 +2,31 @@
...
 weight: 10
 aliases:
   - /docs/intro/overview/what-is-cloudgov
   - /intro/overview/what-is-cloudgov
 ---
 
-cloud.gov gives teams working for federal government a secure, fully compliant foundation, integrated with private sector infrastructure providers, on which to build and release products and updates quickly.
+## What is cloud.gov?
```
brittag: We don't need this extra heading here, since the page will automatically have the title on top ("What is cloud.gov?").

brittag (approved):
Thanks James for these revisions!

PR #527, opened 2016-11-04
SSH docs refactor

berndverst:
Adds better instructions for `cf ssh`


mogul (approved):
I paired on this one, and it looks good to me!

PR #539, opened 2016-11-09
cloud.gov IdP documentation

rogeruiz:
This is a work in progress branch for the documentation coming out of 18F/cg-product#138.


rogeruiz:
This pushes 18F/cg-product#138 along and should help with completing the explanation bits of 18F/cg-product#228.

rogeruiz:
🤔 about a bit about the documentation here. I think just having something under `apps/managing-teammates` might be the best and only place to update with the cloud.gov IdP since most of the login documentation points users _eventually_ to this page from the other sections that original had `TODOs`.

screen shot 2016-11-15 at 12 23 26 pm

LinuxBozo:
@rogeruiz can we just cut out the bit about the username/password box in both the screenshot and text? Hopefully we'll be able to completely remove that box so there is less confusion.

rogeruiz:
@LinuxBozo updated.

rogeruiz:
I'm wondering if we ~~can to~~ should include some links to authenticator apps for phones like Google Authenticator, et al. 🤔 @LinuxBozo @brittag

LinuxBozo:
@rogeruiz I would say yeah, it would be a good idea to mention that some TOTP app will be required now, and we'll want to make changes to the invitation app email template to reflect this as well.

brittag:
Agreed, we should recommend authenticator apps! I'd be happy to take an outsider look at this after the TODO bits get filled out - pretending in my head to be a new cloud.gov user and seeing where I get confused. :)

rogeruiz (commented):
:shipit: Thanks @datn for the updates. This is good for an MVP.

datn (approved):
Approved!

PR #542, opened 2016-11-09
Fix dead links in docs

bleach:
I found a couple of dead links in your docs and think I was able to work out where they should go.


afeld (approved):
Thanks so much for the contribution! 🇺🇸

Issue #485, opened 2016-10-25
As a reader, I want to know what content is specific to GovCloud vs. East/West


afeld:
We already do this on a bunch of pages, but I suggest that we
1. [x] Audit/update the docs for what _isn't_ marked as such and shouldn't be
2. [x] Come up with a standardized way of indicating the environment(s?) a page or section of a page to pertain to; Don't Make Them Think

/cc @jameshupp

afeld:
My other thought here is that we could even have a toggle to say whether you're using East/West or GovCloud, and it would hide the non-relevant parts of the documentation.

mogul:
It's great if you can figure out how to do that... The other use-case is for people using CLI or dashboard when an action could be taken in either. See for example [the client-specific instructions](https://get.slack.help/hc/en-us/articles/201330256-Invite-new-members-to-your-Slack-team) in Slack's docs.

Issue #333, opened 2016-07-20
Separate cloud.gov user documentation from contribution documentation


brittag:
We have two kinds of technical documentation in our docs right now:
- "Usage" for users (application team and agency admin staff, such as developers, org admins, product managers)
- "Contributing" for contributors and potential contributors to cloud.gov (18F staff, external volunteers, users curious about contributing fixes)

Users are our primary audience for the "Documentation" link in the cloud.gov header, and it'd be good to help them find what they need as quickly as possible - so let's make this "Documentation" link (docs.cloud.gov) be dedicated to user documentation. This will also allow our user documentation to have more space to grow, with more sub-menus, while still being fast to access.

The contribution documentation should be clearly linked from the user documentation, since users may be interested in contributing.

afeld:
The current "Contribution" section is really more about "Operation", rather than "here's how to contribute to our repositories/community".

brittag:
Yes! I imagine part of this task would be to write a little preface for the Operations docs explaining that, and also write a little intro to how to contribute as a volunteer (or as an employee of a different org/agency/company).

mogul:
cc @jameshupp since he'll likely want to look at this when he rejoins!

berndverst:
I would like the operational information to not live on docs.cloud.gov and definitely not in the same navigation hierarchy as our technical content.

I don't feel that having these documents on docs.cloud.gov is beneficial to our customers. Whether any customer has benefited from these documents so far, I do not know. That being said, I believe that we would be best advised not to use the cloud.gov / docs.cloud.gov websites to track our own procedures, but instead moving this to Github or some sort of separate Wiki.

I'm also on the fence as to whether this information should be open at all. But if we choose to keep it public, why not just have it in Github via Github flavored markdown? We can still link to whichever Github repo we use from the cloud.gov marketing page.

/cc @mogul

dlapiduz:
We can do something like docs.cloud.gov/ops and have all the docs there. I don't think we should hide that info though.

mogul:
Right. The intent is that this information is open for people who want to
follow in our footsteps, and understand how our operations are handled.
It's fine to keep out of search results, but operators need to search too,
so we should think about how to preserve that capability. ;)

brittag:
[...]

PR #544, opened 2016-11-10
visually distinguish environment-specific blocks of content

afeld:
Closes https://github.com/18F/cg-site/issues/485.

Uses [Hugo shortcodes](https://gohugo.io/extras/shortcodes/) to accomplish this. This is mostly a proof-of-concept - needs design :heart: I have no strong attachment to _how_ they are visually distinguished...just that we get this in sooner than later, since more and more of our content is split in this way.

screen shot 2016-11-10 at 1 51 04 am

/cc @thisisdano @vz3 @berndverst @jameshupp @msecret @adborden

### DIFF
--- a/static/assets/css/main.css
+++ b/static/assets/css/main.css
@@ -17,7 +17,7 @@
 
 .env-specifier {
   color: #333; /* $color-textblack */
-  font-family: '18Franklin'; /* $font-sans */
+  font-family: '18Franklin-webfont', system, -apple-system, BlinkMacSystemFont, 'Roboto', 'Helvetica Neue', 'Helvetica', 'Arial', sans-serif; /* $font-sans */
   font-size: 14px; /* $sans-s4 */
   font-weight: 400;
 }
--- a/static/assets/css/main.css
+++ b/static/assets/css/main.css
@@ -1,41 +1,57 @@
 .env-block {
-  margin-bottom: 1.5rem;
-  margin-top: 1.5rem;
+  margin-bottom: 24px; /* $grid-3 */
+  margin-left: 16px; /* $grid-3 */
+  margin-top: 24px; /* $grid-3 */
   position: relative;
 }
 
 .env-block::before {
   border-radius: 3px;
   border-style: solid;
-  border-width: 3px;
+  border-width: 2px;
   content: '';
   height: 100%;
-  left: -1rem;
+  left: -16px; /* -$grid-3 */
   position: absolute;
 }
 
 .env-specifier {
-  font-size: 0.8em;
-  font-weight: bold;
+  color: #333; /* $color-textblack */
+  font-family: '18Franklin'; /* $font-sans */
+  font-size: 14px; /* $sans-s4 */
+  font-weight: 400;
 }
+
+.env-specifier strong {
+  font-size: 14px; /* $sans-s4 */
+  font-weight: 600;
+}
+
 .env-specifier a {
+  color: #333; /* $color-textblack */
   text-decoration: underline;
 }
 
-/* $color-primary-darker from cg-style */
-.env-east-west::before {
-  border-color: #2c3e50;
-}
-.env-east-west .env-specifier,
-.env-east-west .env-specifier a {
-  color: #2c3e50;
+.env-specifier + h2,
+.env-specifier + h3,
+.env-specifier + h4{
+  margin-top: 8px; /* $grid-1 */
 }
 
-/* $color-blue from cg-style */
-.env-govcloud::before {
-  border-color: #056dd4;
+.env-east-west .env-specifier,
+.env-east-west .env-specifier a {
+  color: #333; /* $color-textblack */
 }
+
+.env-east-west::before {
+  border-color: #74B1DB;
+}
+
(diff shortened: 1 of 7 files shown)


jameshupp:
I love this idea. A couple thoughts:
- I assume the idea is that `East/West only` and `GovCloud only` would also be added by the short code? Semantically, it's probably easier to edit the content if editors aren't applying a short code that says "This is E/W" and then typing a second time something that delivers the same message.
- Not sure I'd use the word "only" in the labels unless there's an excluded middle (environments that are neither E/W nor GovCloud). If there is one that applies to only a few cases (I know there _used to be_ one or two) then keep the word as a way to get those folks to ask directly.

afeld:
> I assume the idea is that `East/West only` and `GovCloud only` would also be added by the short code?

Yep! The shortcode names are arbitrary - the content lives in (and is injected by) the shortcode template. Good news is, that means we can change it in a single place and it will apply everywhere.

> an excluded middle (environments that are neither E/W nor GovCloud)

Well, that would be everything that isn't surrounded by these shortcodes. No strong feelings about the wording though. Figure we'd probably want to include a link to "tell me more" that explains the two environments anyway.

jameshupp:
> the content lives in (and is injected by) the shortcode template

Awesome!

> that would be everything that isn't surrounded by these shortcodes

Sorry, I didn't mean the "only" point in terms of the content of this page.

I meant in terms of other environments. The word "only" does the work of saying "this environment and not other environments". But if there are only two, saying that about both of them feels extraneous. I'd prefer just using `East/West` and `GovCloud` (no "only").

afeld:
screen shot 2016-11-10 at 12 16 02 pm

adborden:
Nice! I would like to eventually see a similar pattern for showing examples in specific programming language, or specific OS.

msecret:
I feel like we can use some of the ideas here to provide docs for CLI vs dashboard.

thisisdano:
I like this — I've already developed a bit of a pattern for including rounded lines to the side of blocks, so it wouldn't be hard to implement. If we ever had a situation where there were more than 2 options, we might want to consider tabs. But for now, consecutive seems cool.

env-blocks 2x

thisisdano:
The only other consideration is that red is an error state pattern on the dashboard. I don't necessarily think that's a problem here in the docs, but it might be worth looking at other color options. (I like red and blue, tho, btw.)

env-blocks-more-colors 2x

afeld:
> I've already developed a bit of a pattern for including rounded lines to the side of blocks

👍 

> If we ever had a situation where there were more than 2 options, we might want to consider tabs

The good news is that this is relatively easy to change later.

> red is an error state pattern on the dashboard

I have zero attachment to the colors I put in there—they were just placeholders. Want to give me a couple of hex values and I can update?

brittag:
In case it helps: I'm always in favor of reserving reddish colors for error states and greyish colors for disabled/inactive states. Blue and green might be fun here, as a nod to [blue-green deployments](https://docs.cloudfoundry.org/devguide/deploy-apps/blue-green.html). :)

afeld:
Well, [the yellow and green colors from cg-style](https://github.com/18F/cg-style/blob/65e1885238d23e8b4968d5c9c44cc25b1edef002/src/css/override_vars.scss#L7-L8) don't meet the contrast thresholds for accessibility (unfortunately), so went with (a conservative choice of) blue and very-dark-blue.

screen shot 2016-11-15 at 11 00 05 pm

Would love to get this pull request in sooner than later to have _something_, and anyone is welcome to do a follow-up to play with the colors. Sound good?

thisisdano:
I've got a couple suggestions for colors and styles that I've been working on. I'll modify the branch tomorrow AM if that's OK. And agreed, I'd love to get it all merged asap.

afeld:
Sounds good! You are the final word on colors anyway 😉

thisisdano:
OK, since the blue, red, green, and yellow colors all have specific meaning elsewhere in the platform, I've tried to avoid those colors for distinguishing color here. Also, since contrast is an issue with colored text, I've tried to avoid it — the contrast rules mean that the colors will either be very saturated (too hot for our palette) or very dark (too difficult to distinguish from each other). I tried a version where the heds are both black. I chose a lighter blue than active/link blue for E/W and plain ol text black `#333` for GovCloud.

Since the heds are no longer distinguished by color, I tried to make them clearer by making them more verbose ("In the [env] environment"). This seems to make the hed clearer — is the wording correct?

How does this strike folks?

---

screen shot 2016-11-15 at 10 44 55 pm

static/assets/css/main.css
```diff
@@ -1,34 +1,41 @@
...
   left: -1rem;
   position: absolute;
-  width: 3px;
 }
 
 .env-specifier {
   font-size: 0.8em;
   font-weight: bold;
 }
+.env-specifier a {
```
msecret: Should be a line above this

msecret (approved):
This should technically be moved into cg-style, but I'm fine with it existing here for a while.

afeld:
Something weird happened with the commit history...rebased to clean it up.

PR #545, opened 2016-11-10
remove SCSS file that isn't included in the build

afeld:
Not sure if it _should_ be included in the build...? Anyway, was just confusing to me, and it's not clear how we'd integrate this file with the cg-styles anyway.


msecret:
This file looks like it's for styles specific just to cg-site. Did you check if anything in the layout breaks after taking it out?

afeld:
Yep, no change. The stylesheet isn't being included in the site, directly or indirectly.

msecret (approved):
I think it was brought over from landing and wasn't really being used.

PR #552, opened 2016-11-11
Adding search to overview and updates pages

brittag:
This would make the header nav consistent across all pages except Dashboard.

### DIFF
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -54,6 +54,11 @@
             <li class="nav-link">
               <a href="/docs/help/">Contact</a>
             </li>
+            <li class="js-search-button-container">
+                  <a class="js-search-button">
+                    <span>Search</span>
+                  </a>
+            </li>
           </ul>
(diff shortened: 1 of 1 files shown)


msecret (approved):
Looks good from a code standpoint.

afeld:
Not sure if I understand what this change is doing...maybe a before and after screenshot? Is there a corresponding change in another repo?

brittag:
@afeld Sure! This change adds a "search" button to the top right of the homepage and the Updates section. It's the same search button that the Documentation and Overview sections have right now. Here's before (current site):

screen shot 2016-11-15 at 10 49 16 am

And here's after this change:

screen shot 2016-11-15 at 10 48 59 am

This small code change (to create a visible button) is the only change that needs to happen to create the search box, since there's already `search.usa.gov` code in the page. I figured this out by going to the partials directory (https://github.com/18F/cg-site/tree/master/layouts/partials) and comparing the current `overview-header.html`, `docs-header.html`, and `header.html` files.

PR #553, opened 2016-11-11
Add services section with pricing and details.

dlapiduz:
Here is a sample of what this looks like: https://docs-services.fr.cloud.gov/docs/services/

What do you think @brittag @jameshupp @berndverst @nikzei @mogul?


mogul:
First, 👏 👏 👏 I think it's long overdue and I mostly like it. 
- I think things like "compliance status", "paid/free" and other metadata belongs on the main page, but I think I'd expect that the production-ready, alpha, and beta status only be indicated at the individual plan level on the sub-pages for each service.
- The descriptions for "Beta" and "Alpha" vs "production ready" are a little too terrifying, at least on the main page. Obviously people are using ElasticSearch for production stuff, for example. So maybe we should instead speak in terms of risk, since that's what ATOs are all about. "To use this service, you must accept risk that..." (I think the descriptions in the page, eg "downtime and data loss are possible" are fine, it's just the names used which imply not production-ready.) This is where I think the use of the word "experimental" for a lot of our stuff is really good.

I bet @sharms also has feels about this, since we've talked about it previously.

mogul:
> So maybe we should instead speak in terms of risk, since that's what ATOs are all about. 

Taking this thought further: How about we cleave to the FIPS/NIST language, and for each plan there are `Confidentiality`, `Integrity`, and `Availability` ratings of Low/Medium/High. So for example:
- In RDS' `shared-psql` plan, the Confidentiality is `Medium`, but for other plans it's `High`.
- The public S3 plan is Confidentiality:`Low`, even while `High` for Integrity and Availability.
- Services that could lose data are Integrity: `Low`
- Services that are not HA are Availability: `Low`
- Services that are HA but not auto-scaling are Availability `Medium`
- Services that auto-scale (not that we have any now) are Availability: `High`

Etc. etc.

Then people can make more informed choices for what to use in dev, in production, and for what kind of system their deploying and what their risk-tolerance is for that system and the services that it relies on.

brittag:
This is really interesting! I tried looking at this while pretending to be a relatively new cloud.gov customer, and I see some potential flow problems. The following is meant to be encouragement toward figuring out how to get this done really well, not discouragement!
- If you’re new to cloud.gov (and AWS) and you're going through the sidebar to try to figure out how to get a database set up for your application, "Apps" -> “Databases” is a more obvious thing to click than “Services” -> “AWS RDS”. We _can_ sort the sidebar by "Services", but it requires us to think carefully about the intro/onboarding flow (and probably refactor it) so that people learn about cloud.gov's architecture early in their experience with cloud.gov, so that they know when to go to "Services".
- Similarly, if you're new to cloud.gov, you probably wouldn't know to look for “deployer account” under “services” - users probably wouldn’t guess that a deployer account is architecturally a “service” until cloud.gov explains that to them, and I’m not sure if we have some kind of intro that explains this so they know where to find it.
- If you’re reading “Custom domains” and you want to use the managed service method, you get pointed to “CDN Service”, and right now that page would be confusing to land on - you’d have to scroll to the bottom to start finding the instructions for accomplishing your task.
- There’s nothing under "Services" -> “Introduction” - it goes to a blank page.

Once we have the flow sorted out, this would need some copyediting to get the details smooth.

Also we merged in some S3 copy improvements at https://github.com/18F/cg-site/pull/543, and this PR looks like it would undo those changes.

dlapiduz:
@brittag I fixed the PR check issue and I changed the name for the AWS RDS service to say "Relational Databases". I think its easier that way.

dlapiduz:
@mogul I think its a great idea to move to Confidentiality, Integrity, and Availability lingo in these pages but maybe we can start with this and add that later. What do you think?

dlapiduz:
@brittag I also merged the changes from S3 now

berndverst:
Given the huge value of these pages, I think we should merge this and iterate afterwards.

berndverst (approved):

PR #554, opened 2016-11-14
Reference Diego migration guide

mogul:
Not ideal, since the referenced doc targets operators and customers in the same document, but oh well!


mogul:
@jmcarp @msecret I'm having trouble making sense of the build error... Is this a transitory thing with NPM?

dlapiduz (approved):

PR #555, opened 2016-11-14
[WIP] Update homepage content to address user needs

brittag:
@jameshupp and I spent about an hour revising the first few sections of the website homepage to reflect what we've learned about the needs of prospective users.

Here's a work in progress to show what we came up with:

* Revise headline and subtitle to be more specific about what cloud.gov is. This puts "Platform as a Service" in the headline, with an explanation of "Platform as a Service" in the subtitle.
* Revise first chunk to provide a list of value propositions up front, to explain that our value is broader than just taking care of cloud operations.
* Revise "who we are" chunk to mention TTS and speak more plainly.

Our goal is simply to suggest something *better* than the current homepage, not to make a perfect homepage in one iteration. (We specifically timeboxed this.) Suggestions welcome!

This is marked [WIP] since it's not ready to merge.

Context: good to look at the [current cloud.gov one-sheet](https://docs.google.com/document/d/12keNo9E9SCxnVHZ6LcIUXXAyt6I5TyBXm_P2lUp6qYs/edit) + [overview](https://cloud.gov/overview/).

[Draft doc](https://docs.google.com/document/d/1VZ6N8Lmb5tg76dV23HEcxDMU4B74bvaARCVTuC_554U/edit#)

### DIFF
--- a/layouts/index.html
+++ b/layouts/index.html
@@ -13,8 +13,8 @@
     id="main">
   <div class="usa-grid section-bg-parent">
     <div class="usa-width-whole billboard  section-bg">
-      <h1 class="billboard-display_text">A Platform as a Service designed for your government team</h1>
-      <h2 class="billboard-second_text">Focus on your applications. We manage the security, compliance, and maintenance of the underlying virtual machines and servers.</h1>
+      <h1 class="billboard-display_text">A Platform as a Service for government teams</h1>
+      <h2 class="billboard-second_text">Focus on developing your applications. We manage the security, compliance, and maintenance of the virtual machines and servers.</h1>
     </div>
   </div>
 </section>
@@ -23,15 +23,15 @@
   <a name="about"></a>
   <div class="usa-grid">
     <div class="section-content usa-content">
-      <h1>Deploying applications that serve the public shouldn’t be slow and difficult.</h1>
+      <h1>Deploying applications that serve the public shouldn’t be slow or hard.</h1>
       <p>
-        cloud.gov gives government teams access to cloud deployment with fewer hassles. Now you can:
+        cloud.gov gives government teams access to modern cloud deployment with fewer hassles. Now you can:
       </p>
       <ul>
-        <li>Quickly deploy applications on a technology stack that complies with federal policies, without needing to manage and update your own infrastructure.</li>
-        <li>Run modern cloud-native applications. cloud.gov is a platform that provides services on top of Amazon Web Services (GovCloud), so you can use many AWS services as well.</li>
+        <li>Quickly deploy applications that comply with federal policies &mdash; without needing to manage infrastructure.</li>
+        <li>Run scalable cloud-native applications. cloud.gov provides services on top of Amazon Web Services (GovCloud), so you can use AWS services.</li>
         <li>Try experiments: build and test prototypes without adding extra expense.</li>
-        <li>Shorten the path to ATO (Authority to Operate) for each new or updated application from your team, because after your agency gives cloud.gov ATO, only your application needs to be evaluated for security and compliance.</li>
+        <li>Shorten the path to ATO (Authority to Operate) for each new or updated application, because after your agency gives cloud.gov ATO, only your application needs to be evaluated for security and compliance.</li>
       </ul>
     </div>
   </div>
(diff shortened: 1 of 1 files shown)


layouts/index.html
```diff
@@ -13,8 +13,8 @@
     id="main">
   <div class="usa-grid section-bg-parent">
     <div class="usa-width-whole billboard  section-bg">
-      <h1 class="billboard-display_text">The Government Innovation Platform</h1>
-      <h2 class="billboard-second_text">A platform by government developers, for government developers.</h1>
+      <h1 class="billboard-display_text">A Platform as a Service for government teams</h1>
+      <h2 class="billboard-second_text">Focus on developing your applications. We manage the security, compliance, and maintenance of the virtual machines and servers.</h1>
```
mogul, line 17: "maintenance of _virtual machines and servers_" seems redundant. Maybe "maintenance and hosting"?

layouts/index.html
```diff
@@ -13,8 +13,8 @@
     id="main">
   <div class="usa-grid section-bg-parent">
     <div class="usa-width-whole billboard  section-bg">
-      <h1 class="billboard-display_text">The Government Innovation Platform</h1>
-      <h2 class="billboard-second_text">A platform by government developers, for government developers.</h1>
+      <h1 class="billboard-display_text">A Platform as a Service for government teams</h1>
```
mogul, line 16: I'm still worried by `Platform as a Service`. Prefaced with `A` makes it seem like the reader should already know what this is.  I know it's defined in the next sentence but I think some people may stop here. I liked `Compliance as a Service` as more immediately accessible to govt; was that considered?

layouts/index.html
mogul: We don't fully manage all aspects or absolve teams of responsibility. So maybe something like "We reduce the burden of [...] for your team"?

layouts/index.html
```diff
@@ -23,10 +23,16 @@ <h2 class="billboard-second_text">A platform by government developers, for gover
   <a name="about"></a>
   <div class="usa-grid">
     <div class="section-content usa-content">
-      <h1>Cloud deployment is better. But that doesn’t make it easy.</h1>
+      <h1>Deploying applications that serve the public shouldn’t be slow or hard.</h1>
```
mogul, line 26: ❤️

layouts/index.html
```diff
@@ -23,10 +23,16 @@ <h2 class="billboard-second_text">A platform by government developers, for gover
...
       <p>
-        Any federal agency deploying an app has to follow regulations, manage an ATO process, and design a robust service infrastructure. Cloud hosting saves one hassle — managing data centers — and creates another: operating cloud infrastructure and keeping it compliant. It requires a different set of skills that isn’t widely available in government yet.
+        cloud.gov gives government teams access to modern cloud deployment with fewer hassles. Now you can:
       </p>
+      <ul>
+        <li>Quickly deploy applications that comply with federal policies &mdash; without needing to manage infrastructure.</li>
+        <li>Run scalable cloud-native applications. cloud.gov provides services on top of Amazon Web Services (GovCloud), so you can use AWS services.</li>
```
mogul, line 32: This over-emphasizes both AWS, and the access they have to AWS directly. The reality is that it's backed by AWS, but the details of managing AWS-specific parts are abstracted away. There's danger here in them thinking "oh, you procure AWS for me, got it"... _Technically_ that's true but someone will mis-class themselves as a customer for us based on that assumption.

layouts/index.html
```diff
@@ -23,10 +23,16 @@ <h2 class="billboard-second_text">A platform by government developers, for gover
...
+        <li>Try experiments: build and test prototypes without adding extra expense.</li>
```
mogul, line 33: "without adding extra expense" feels weird. Technically it will add resource usage and managed service costs to their bill. Maybe "without obstacles" or "friction-free"?

layouts/index.html
```diff
@@ -23,10 +23,16 @@ <h2 class="billboard-second_text">A platform by government developers, for gover
...
+        <li>Shorten the path to ATO (Authority to Operate) for each new or updated application, because after your agency gives cloud.gov ATO, only your application needs to be evaluated for security and compliance.</li>
```
mogul, line 34: Maybe: "Leverage cloud.gov's FedRAMP ATO (Authority to Operate) to reduce the scope for your agency's compliance processes." ?

layouts/index.html
```diff
@@ -41,9 +47,12 @@ <h2 class="billboard-second_text">A platform by government developers, for gover
...
+      <h1>We’re a government team, and we designed cloud.gov to handle real challenges.</h1>
       <p>
-        18F is a team within the General Services Administration that builds applications for the federal government, rapidly and iteratively. So as we scaled up the number of teams deploying applications, we couldn’t let cloud operations talent become the kind of bottleneck that slows down development for many federal teams. Instead of working on infrastructure support for every individual product, our cloud ops staff focused on how they could give all our development teams a better way to work.
+        cloud.gov is built by 18F in GSA’s Technology Transformation Service (TTS), where our mission is to help agencies build, buy, and share technology that helps them better serve the public. We found that the more apps we built, the more cloud operations and ATOs slowed down deployments.
+      </p>
+      <p>
+        Instead of trying to provide infrastructure support for every product, our cloud ops staff designed a platform that every team could use. Now we’re making it available to other government teams as part of our mission.
```
mogul, line 55: ❤️

mogul (commented):
I can feel where the timebox ended, but this is definitely an improvement... Keep going!

layouts/index.html
nikzei: I agree with cutting line 16, though we should perhaps replace it? For now, we could simply replace it with "Cloud.gov." Otherwise, we don't really emphasize the title of this product on the overview page.

RE: line 17 (PaaS), I hear Bret's concern, though my reaction is that PaaS is a very googleable term, whereas CaaS is less so, or when googling this term, the results present a less standardized concept.

Perhaps we're not at the final point but I still find these edits to be improvements so I'd like to merge with the intention that we'll keep tweaking and improving. Good @mogul?

nikzei (commented):
A comment re: language choices: PaaS versus CaaS.

mogul:
Sorry, didn't mean to say you should hold this up on me... I'll add an explicit review so it can be merged.

mogul (approved):

PR #556, opened 2016-11-15
Update pricing-terminology.md

suprenant:
Updated "March 2016" to "soon" for setting prices given that March has passed


brittag (approved):
Makes sense to me!

PR #557, opened 2016-11-15
Updates custom domain example to use govcloud prefix

konklone:
Since the Let's Encrypt integration is GovCloud-only, this uses `fr.cloud.gov` as the example suffix.


afeld (approved):
This is about to change with https://github.com/18F/cg-product/issues/499, but good catch!

PR #566, opened 2016-11-16
Create how-to-purchase.md

nikzei:
A new page providing prospective customers with an intro into our IAA process and a link to our intake form.
cc @berndverst


berndverst:
Sounds good. Merging this.

Please note that we may need to create a PDF or some other way to submit the form that will work in case Google Forms is blocked by the requesting agency. Maybe a special form app on cloud.gov :)

berndverst (approved):

Issue #560, opened 2016-11-15
The "deploying x" guides need copyediting 


brittag:
In order for cloud.gov tenants to set up apps quickly and easily, we need our "deploying x" guides to be clear, concise, and up-to-date.

Looking at this guides, they have some outdated bits ("It should now be running at APPNAME.18f.gov!") and passive voice. We should review them and fix them up:

- [x] https://cloud.gov/docs/apps/django/
- [x] https://cloud.gov/docs/apps/flask/
- [x] https://cloud.gov/docs/apps/meteor/
- [x] https://cloud.gov/docs/apps/rails/
- [x] https://cloud.gov/docs/apps/node/
- [ ] https://cloud.gov/docs/apps/static/

cc @jameshupp

mogul:
I can appreciate what we did in #572 but now there's no indication of what languages we support. We should probably expand [this page](https://cloud.gov/docs/apps/frameworks/) to announce that clearly rather than burying that info behind buildpack jargon.

Issue #162, opened 2016-01-06
audit for what can be merged to the upstream Cloud Foundry docs


afeld:
_Will update this list with specific areas to look at._

* https://docs.cloud.gov/ops/deploying-the-admin-ui/
    * See discussion at https://github.com/18F/cg-docs/pull/161/files#r48972043.
* https://github.com/18F/cg-site/pull/572

afeld:
This could really be (part of) a more general "audit our docs to figure out what isn't cloud.gov-specific and then contribute them upstream" story. Sounds like [the Cloud Foundry Foundation might be interested in helping](https://twitter.com/cloudfoundry/status/701914529651941376).

/cc https://github.com/18F/cg-docs/issues/180

PR #572, opened 2016-11-17
Link to existing docs and drop duplicates.

jmcarp:
We currently document deploying apps built with django, rails, and a few other frameworks. Most of this documentation can be found elsewhere, so let's link to the official buildpack docs and sample applications instead of writing our own. To start with, I cut out lots of framework-specific docs and replaced them with upstream links. Let me know if I took out too much or didn't say enough!


brittag:
Related issue about these guides being a lil bit rusty right now: https://github.com/18F/cg-site/issues/560

afeld:
I am 💯  for using upstream docs instead of our own, but also don't want to lose any important information that isn't reflected elsewhere (yet). I haven't gone through line-by-line to check... @jmcarp how confident are you about it?

jmcarp:
To clarify, I don't think that the upstream buildpack docs cover every line of these docs. I don't think upstream describes meteor. That's fine! It's beyond the scope of cloud.gov (and beyond our competence) to document every language and framework that runs on the platform.

In some cases, the heroku docs are more detailed than the cf docs, so we can link to those too.

sharms:
Is there any evidence to suggest that a developer getting started in Meteor (not a core 18F framework) would look to the Cloud.gov team for direction in that area?

jmcarp:
@afeld: would you prefer breaking this up into smaller patches? For example, we could start by replacing the meteor docs with a link to the unsupported meteor buildpack, or just delete them entirely since they're unsupported, and go from there?

content/docs/apps/frameworks.md
```diff
@@ -0,0 +1,8 @@
+---
+menu:
+  docs:
+    parent: advanced
```
afeld: Thoughts about using `parent: apps` instead?

afeld:
It's ok, we can punt on it: https://github.com/18F/cg-site/issues/162. Just one more thing: Thoughts about adding [aliases](https://gohugo.io/extras/aliases/) for the old URLs to redirect to the Frameworks page?

/cc https://github.com/cloudfoundry/docs-buildpacks/pull/114

afeld (approved):

PR #578, opened 2016-11-18
Move user-facing account info to Accounts page

brittag:
This is a procrastination-inspired refactor of the new info in https://github.com/18F/cg-site/pull/539, with the following changes:

* Move instructions about how to use MFA to the beginner-facing "Accounts" page rather than the "Managing Teammates" page, so that the info is presented straight to the person who needs to use the info (rather than to their teammate).
* Edit those instructions to remove implementation information that the reader doesn't need to learn about to get their task done (the distinction between UAA and IdP).
* Revise the "Accounts" page to try to better describe our complex account situation.
* Change link to "who can use cloud.gov" to "email us", since that page no longer really helps the reader of this page.
* Crop the screenshot so that it takes up less visual space, allowing the reader to read more without scrolling.

cc @rogeruiz, @LinuxBozo, @datn


LinuxBozo:
@brittag Couple things I see:
1. Under `Agency single-sign-on accounts` it says, here are the details, but there don't seem to be any.
2. `content/docs/getting-started/setup.md` probably needs modified as well so that govcloud cloud.gov account holders know to use `--sso`

mogul (approved):

PR #583, opened 2016-11-20
Adds FDIC agency login to various docs pages.

berndverst:
Now that we have integrated FDIC's SAML-based agency login we can include this in the documentation. This change makes that happen.


afeld (approved):

afeld:
Good to merge - just has a conflict.

PR #586, opened 2016-11-21
blockquote styling

afeld:
Leveraged the environment specifier styling to give a vertical bar next to `<blockquote>`s.

screen shot 2016-11-21 at 4 29 50 pm

Once again, I have zero attachment to the color. Also made a couple of rendering tweaks to the org creation steps.

/cc @thisisdano @msecret @brittag

### DIFF
--- a/content/docs/apps/govcloud.md
+++ b/content/docs/apps/govcloud.md
@@ -81,7 +81,6 @@ Any labor involved from your project team is the responsibility of your project.
     * Specify that the org should be created in `GovCloud`.
     * An admin will confirm the information, and [create the org]({{< relref "docs/ops/create-org.md#non-sandboxes" >}}) for you.
 1. Install the [CF Targets plugin](https://github.com/guidowb/cf-targets-plugin).
-1. [Set up the spaces.]({{< relref "docs/apps/production-ready.md#space-per-environment" >}})
 1. [Give permissions to the appropriate people.]({{< relref "docs/apps/managing-teammates.md" >}})
 1. Deploy the application to the GovCloud environment.
     * This is a good time to ensure that you are following [deployment best practices]({{< relref "docs/apps/production-ready.md" >}}).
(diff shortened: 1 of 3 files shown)


content/docs/apps/govcloud.md
```diff
@@ -81,7 +81,6 @@ Any labor involved from your project team is the responsibility of your project.
     * Specify that the org should be created in `GovCloud`.
     * An admin will confirm the information, and [create the org]({{< relref "docs/ops/create-org.md#non-sandboxes" >}}) for you.
 1. Install the [CF Targets plugin](https://github.com/guidowb/cf-targets-plugin).
-1. [Set up the spaces.]({{< relref "docs/apps/production-ready.md#space-per-environment" >}})
```
afeld, line 84: This is done by the org creation script.

brittag:
Oh cool! Could we have a color/style that's a bit different from the GovCloud/E/W styling, to minimize confusion? Also, it looks like there are a few content changes in here as well - not sure if those are meant to be in this PR with the styling change.

afeld:
> Could we have a color/style that's a bit different from the GovCloud/E/W styling, to minimize confusion?

I'm running out of in-theme color ideas 😉 

> it looks like there are a few content changes in here as well - not sure if those are meant to be in this PR with the styling change

Yeah, they didn't feel significant enough to be their own PR, but I can break them out if desired.

afeld:
e

brittag:
Ahh, I just checked the GovCloud + E/W colors, and those are black/blue while this is grey - nevermind. And I'm fine with these content changes being here - just checking.

LGTM from a content design perspective - I'll leave this open for visual design review though. Maybe @thisisdano?

thisisdano:
Lightened the gray color and changed the typeface to distinguish.

- - -

screen shot 2016-11-29 at 12 56 43 pm

thisisdano (approved):
I just made a couple changes to the blockquote styling
- changed the bar color
- changed the typeface

afeld:
That color won't have high enough contrast ratio. Since the font is different, I think it's ok for now, but we should probably do an accessibility review at some point.

PR #587, opened 2016-11-21
template list got broken when space was added, it seems

datn:


afeld:
Umm, are you sure? Seems like the other way around...

datn:
The site as it currently stands looks broken to me:

screen shot 2016-11-25 at 10 31 27

afeld (approved):

afeld:
Rebased to fix it up.

PR #589, opened 2016-11-22
Update govcloud invitation URL

LinuxBozo:
The invite app has evolved. This means that it provides more than just invitation functionality. As such, the app has a new URL, and there is an invite endpoint. This PR addresses those changes.

@rogeruiz @brittag

### DIFF
--- a/content/docs/apps/govcloud.md
+++ b/content/docs/apps/govcloud.md
@@ -23,7 +23,7 @@ GovCloud-specific information
 
 ### Breaking changes
 
-- GSA, EPA and FDIC accounts already work in GovCloud. Others need to be explicitly [invited](https://invite.fr.cloud.gov).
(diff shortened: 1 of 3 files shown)


brittag:
This seems ok from a content perspective, but it's ended up with a merge conflict. Also not sure whether we should have a redirect from the old URL to the new URL, in case people have bookmarked the old one?

LinuxBozo:
@brittag this should have merge conflicts solved now.

brittag:
Thanks! I'll merge, since my question about redirects is separate from getting this updated in the docs. :)

brittag (approved):

brittag:
Ohh I've now run into a confusing bug with the new URL, so I'm no longer sure I should merge this. I'll file it separately.

brittag:
@LinuxBozo here's what I'm running into: https://github.com/18F/cg-product/issues/529

brittag:
Merging since those concerns have been resolved. Hooray!!!!

PR #599, opened 2016-11-26
remove fork of 18F wide process

NoahKunin:
This process lives elsewhere, and will be a Service wide process eventually, so having the steps repeated here is duplicative and will lead to content rot as people forget to update it in its canonical home (the Handbook).

Also changes tense of first `Assess` step to better reflect actual flow (report first, identify false alarm second)


NoahKunin:
X-posted to SkyPorter Slack, which I _think_ is correct.

NoahKunin:
related to 18F/infrastructure#632

mogul (approved):
It's HighBar, but looks good in any case

PR #602, opened 2016-11-29
One of task without route should not use health check.

berndverst:
Minor tweak to the one-off tasks docs.


mogul (approved):

rogeruiz:
@berndverst to bubbled up @jmcarp's comment [here](https://gsa-tts.slack.com/archives/cloud-gov-atlas/p1480379020010817). I think since we're talking about one-off tasks, we should just have all the options in the command. 🤔 Since that would mean the minimum amount of changes to the copied over manifest.

```shell
cf push -f task_manifest.yml  --health-check-type none --no-route
```

mogul:
Seems legit... Sorry I was too fast on the draw. I'd happily approve another PR!

PR #603, opened 2016-11-29
Make one-off tasks easier for all users

rogeruiz:
With a single command, power-users can learn from this documentation to
leverage the command-line more, while new-users can keep changes to
their manifests to a minimum and learn to use the command-line. Curious
users can read the CF documentation to learn even more about the `yaml`
stanzas if they're feeling particularly thirsty. :sweat_drops:


mogul (approved):
Nice and parallel now.

PR #605, opened 2016-11-29
Add a link to the #sharing-secret-keys docs

rogeruiz:
This came up during my time on second-tier support. It's good to remind users that the links we share will expire if they wait until the next day to view them.


brittag:
@rogeruiz I added a commit with a few changes:

* Update links to inline style for easier editing (and to match the links in most of the rest of the docs) + use the built-in relative link style for more durable links - for example: `[sharing secret keys policy]({{< relref "docs/ops/secrets.md#sharing-secret-keys" >}})`
* Tell readers more assertively to follow requirements and best practices, since they should do those things.
* Update intro to have more context.

mogul (approved):

PR #616, opened 2016-12-02
Removing Compliance Toolkit description

nikzei:
Edit intended to shift focus of this page to only the services that we are actively offering to customers. 

 When discussing the toolkit, it would be more accurate to describe this as separate from our main offering, and something that the open control community maintains (rather than something that we maintain ourselves).

### DIFF
--- a/content/overview/overview/what-is-cloudgov.md
+++ b/content/overview/overview/what-is-cloudgov.md
@@ -25,8 +25,3 @@ cloud.gov runs on top of industry-provided infrastructure ([currently Amazon Web
 
 All the software components of the cloud.gov Platform as a Service are [open source]({{< relref "docs/ops/repos.md" >}}).
 
-## cloud.gov provides a compliance toolkit
-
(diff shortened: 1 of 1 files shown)


afeld:
Going to defer to @mogul on this.

brittag:
We also mention the toolkit in a couple other places: https://github.com/18F/cg-site/search?utf8=%E2%9C%93&q=toolkit - not sure if those should be updated as well.

nikzei:
Thanks for that. And the goal here isn't to leave out any mention of the toolkit so much as to make our communications about our offerings accurate. @jameshupp and @brittag - throwing this in your court!

brittag:
I started looking into this to make the compliance information in the intro docs reflect what we offer (including reducing the emphasis on Compliance Toolkit), but it looks like I'd run into a bunch of conflicts with the PR at https://github.com/18F/cg-site/pull/632 - so it looks like I should set this aside until that gets merged.

nikzei:
PI8 75% assessment:
SHOULD
--> this is directly related to our work this PI on making communications about our product clearer on the website.

brittag:
Filed a new PR that includes a fix for this: https://github.com/18F/cg-site/pull/669

Issue #447, opened 2016-10-03
update Metrics page


afeld:
Seems to be very out-of-date.

https://docs.cloud.gov/ops/metrics/

brittag:
We solved this by deleting the page: https://github.com/18F/cg-site/pull/619

PR #619, opened 2016-12-06
Remove outdated metrics page and "common procedures" page

brittag:
I know this page is outdated overall (https://github.com/18F/cg-site/issues/447), but the access parts seem outdated in a way that could cause confusion with our compliance work - so here's a suggestion to delete those parts. :)


jmcarp:
@brittag: can we delete this page altogether? It's basically all wrong or irrelevant, and most of the steps it describes have been automated since it was written.

brittag:
@jmcarp Yes! Added a new commit that deletes the Metrics page and also https://cloud.gov/docs/ops/procedures/ - which talks about (1) outdated metrics stuff and (2) stuff that is already in the 18F handbook that we don't need to duplicate.

jmcarp (approved):

PR #626, opened 2016-12-07
suggested edits on s3 instructions

LindsayYoung:
I am pretty sure you need a manifest to cf push and your app needs a name. I couldn't get it to work otherwise. Also, you don't want dummy apps eating up your quota so it is good to set that low.


afeld (approved):
Thanks!

PR #627, opened 2016-12-08
Adding focus to search text box

line47:
When the search button was clicked the focus would not jump to the
search text box. This has been updated as well as pressing the escape
key to change the focus to the search close button.

### DIFF
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -55,7 +55,7 @@
               <a href="/docs/help/">Contact</a>
             </li>
             <li class="js-search-button-container">
-                  <a class="js-search-button">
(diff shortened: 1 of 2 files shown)


afeld:
YES. Thank you. @msecret @adborden Should this be in cg-style?

afeld:
Funny, @jcscottiii and I were just talking about how we wanted this yesterday.

msecret (approved):
I don't think cg-site, or cg-docs even use the cg-style JS right now, so we can keep here in the meantime.

mogul:
Hmm, not seeing it live. Maybe this didn't deploy automatically?

PR #628, opened 2016-12-08
Point to hint on setting up SSH environment

mogul:
We're seeing support requests about this pretty regularly so far.

### DIFF
--- a/content/docs/apps/using-ssh.md
+++ b/content/docs/apps/using-ssh.md
@@ -10,7 +10,7 @@ Here's how to get a SSH shell in the environment where your app is running. You 
 {{% govcloud %}}
 ### CF SSH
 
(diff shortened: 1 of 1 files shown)


brittag:
Can we make "You may want to" a little more specific, to help people determine whether they want to? Such as "If you need to do x, do this" or "If you get x behavior, set up your session to match your application's environment"? (I'm not sure what symptoms people are sending us in support requests.)

mogul:
I think it's something pretty much everyone will want to do... I'm sure there's a good reason it's not the default, but if it were up to me this would be automatic! How about just leaving it imperative, eg "Configure your..." or else just predictive of the need, eg "You'll want to..."?

brittag:
"Configure your" or "You'll want to" both sound reasonable to me!

brittag (approved):

PR #629, opened 2016-12-08
Use RFC 5737 addresses in documentation

cnelson:
In order to avoid Private IP disclosure false positives when scanning this site we should use TEST-NET addresses in our docs where appropriate.

cc: @brittag


brittag (approved):
Thank you! This makes sense to me; nice to use the official 555-5555 of IP addresses.

PR #634, opened 2016-12-14
link to the Copy plugin

afeld:


brittag:
I'd love for this to be stated as more direct guidance if we can. For example, what are cases where the copy plugin helps, and what are cases where it won't help? How can the reader figure out if the manual steps are required?

afeld:
I haven't actually used it, so couldn't say 😛

jmcarp:
I'm thinking we should either figure out how the plugin works so that we can describe it, or not mention it. Saying "here's a thing ymmv" could be confusing.

afeld:
Fair point - won't bother right now then.

PR #635, opened 2016-12-14
improving the "what it is" homepage section

jameshupp:
## Changes
- shifts the “what it is” section from “this is why we built it” framing to “this product serves you” framing.
- added a section to "Contact us" on sending us email and updated the signup form intro text to differentiate the two.
- refocused the section on why the product works, including shifting "Scalability" to "Economy of scale" to more narrowly describe what it was about. _(added this one after merge, forgot to add before)_

## Notes
- for the moment, still short on the “why the product works” section, but doesn’t seem any reason not to fold in the “what the product is” section.
- calling this WIP, happy to keep working on this branch, but can fold in at any time and just build on it.

### DIFF
--- a/layouts/index.html
+++ b/layouts/index.html
@@ -81,9 +81,12 @@
             <use xlink:href="assets/img/cloudgov-sprite.svg#i-shipping"/>
           </svg>
         </i>
-        <h3>Platform-as-a-service</h3>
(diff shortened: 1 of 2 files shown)


nikzei (approved):
Nice improvement!

PR #638, opened 2016-12-15
December 14 Platform Release Notes.

berndverst:
image

Lots of CVEs:

image


content/updates/2016-12-14-release-notes.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Added
+- New [dashboard](https://dashboard.fr.cloud.gov) features
+  - The number of application instances, instance memory and disk quota can now be edited from the application page.
+  - Org and space level views now include a quick overview of the health of application instances.
+- We’ve launched [cloud.gov community](https://community.cloud.gov), a new resource for asking questions, and sharing best practices for using cloud.gov.
+
+### Changed
+- We upgraded Cloud Foundry (the underlying open source project that powers cloud.gov) to [Release v249](https://github.com/cloudfoundry/cf-release/releases/tag/v249) (see also [Release v248](https://github.com/cloudfoundry/cf-release/releases/tag/v248)), which includes updates to buildpacks and improvements for internal components.
+- CloudFoundry upgraded the command line interface (CLI) client. Please upgrade to the [latest version](https://github.com/cloudfoundry/cli/releases/latest).
```
mogul: s/CloudFoundry/Cloud Foundry/

content/updates/2016-12-14-release-notes.md
```diff
@@ -0,0 +1,36 @@
...
++++
+
+Curious what’s new that you might find helpful as a cloud.gov application developer? Here are highlights from our platform updates over the past two weeks.
+<!--more-->
+
+### Added
+- New [dashboard](https://dashboard.fr.cloud.gov) features
+  - The number of application instances, instance memory and disk quota can now be edited from the application page.
+  - Org and space level views now include a quick overview of the health of application instances.
+- We’ve launched [cloud.gov community](https://community.cloud.gov), a new resource for asking questions, and sharing best practices for using cloud.gov.
```
mogul: s/questions,/questions/

content/updates/2016-12-14-release-notes.md
mogul: I'd be specific here... "Changes in Cloud Foundry require an upgrade to the latest version of the CLI in order to `cf logs` to work properly.

content/updates/2016-12-14-release-notes.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Changed
+- We upgraded Cloud Foundry (the underlying open source project that powers cloud.gov) to [Release v249](https://github.com/cloudfoundry/cf-release/releases/tag/v249) (see also [Release v248](https://github.com/cloudfoundry/cf-release/releases/tag/v248)), which includes updates to buildpacks and improvements for internal components.
+- CloudFoundry upgraded the command line interface (CLI) client. Please upgrade to the [latest version](https://github.com/cloudfoundry/cli/releases/latest).
+
+### Deprecated
```
mogul: Sadly, CF's bungling of the CLI/Loggregator compatibility means we have no ability to put this in the `Deprecated` purgatory... It's simply no longer supported. So this should be `Removed`.

mogul (commented):
Do we want to mention the availability of the CLI plugin? And did we already announce the migrate-db thingie?

berndverst:
Incorporated all changes.

image

berndverst:
@mogul which CLI plugin are you talking about? The one to look up routes by @afeld ?

🤔 It would be great if we had a general plugin / tools page that we could reference.

I'm not sure it makes sense to list the db migration (EW->GovCloud) plugin here though because not all tenants need to perform a migration. I think that is better just mentioned on the GovCloud guide page. I'd prefer if these release notes assumed that EW never existed.

mogul:
Well, the `migrate-db` plugin is useful to customers even if they're not migrating between environments! They can use it for taking snapshots of prod environments for reproducing problems in test environments, etc.

It turns out there is already [a plugins page](https://cloud.gov/docs/apps/plugins/) that I've never seen before, but the content is completely unsuited to our typical customer! We should probably just gut this page and start listing stuff we think is helpful and often recommend... These two, the stats plugin, blue-green deploy options, etc. /cc @afeld

berndverst:
@afeld will submit the plugin to https://plugins.cloudfoundry.org/ to we will refer to this in the next platform release notes.

@mogul as the `cg-migrate-db` plugin only currently allows creating snapshots in E/W and uploading them to a S3 bucket in E/W. Let's announce this when the plugin also support GovCloud for exports and therefore can be used in a more general manner. cc @jcscottiii 

So I'd say let's wait until next sprint for both plugins

mogul (approved):

PR #651, opened 2016-12-21
/contact now goes to /docs/help

berndverst:
Per suggestion from @nikzei 

cc: @brittag

### DIFF
--- a/content/docs/help.md
+++ b/content/docs/help.md
@@ -2,6 +2,7 @@
 title: Help
 aliases:
   - /help
+  - /contact
 ---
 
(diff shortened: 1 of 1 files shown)


berndverst:
We can later decide if we want `/contact` to be the main URL for this page and then make `/docs/help` the redirect there instead.

nikzei:
Looks like we wrote similar issues. See: https://github.com/18F/cg-product/issues/583

brittag (approved):

brittag:
Cool, merging since this looks fine to me as a step in the process of switching to /contact/.

PR #652, opened 2016-12-21
Add info about support hours expectations

mogul:


brittag:
Added a commit with minor style changes (plus reflecting change at https://github.com/18F/cg-site/pull/651), although I'm not entirely sure about capitalizing east and west in this context. Checking with content guild on that; merging anyway since that's easy to fix later.

brittag (approved):

Issue #135, opened 2015-12-09
as a developer, I would love it if there were a link to this repo in the footer


ultrasaurus:
I always forget the repo name!

brittag:
This is now consistent in the cg-site footers!

PR #659, opened 2016-12-23
Fixing search focus and updating header and footer consistency 

line47:
There were inconsistencies across the headers and footers on the site.
I have consolidated these into single items to make for easier
development.

This was some tech-debt that made sense to take on as the search was
only working on a few sections of the site as a result of these
separate header files.


line47:
This update also addresses https://github.com/18F/cg-site/issues/135, https://github.com/18F/cg-design/issues/25

brittag:
Hooray, thanks for refactoring these things!

When I try this locally and look at /docs/ (which is a short page), there's sort of a big gap between the end of the content box and the footer, which may be a side-effect of how the footer is placed. Not sure if there's also something wrong/outdated with cg-style in my local version - that happens sometimes for me. Just might be good to check and see if it works the way you want it to work.

Instead of having "Edit this site" and "This site is in the public domain" as two links in the footer, how about just switching "Edit this site" to "[Source code (public domain)](https://github.com/18F/cg-site)" ? That seems simpler.

Bonus round, which could be for a different PR: I also just noticed that our current header + footer contain three ways of directing people to talk to us: "Contact" in header, "cloud.gov help" in footer at left (goes to same place as Contact), and cloud-gov-inquiries in footer at right. Maybe the bottom should have just "Contact" at left and remove the email address at right? cc @jameshupp

line47:
1. That gap looks like a styling issue with the side nav where height is set to 100%. I can adjust that.
2. No problem I can easily make that update.
3. That update should be easy enough as well, we will wait for @jameshupp to weigh in. In the meantime I'll get 1 and 2 knocked out.

line47:
This PR is tied to https://github.com/18F/cg-style/pull/236

brittag (approved):

brittag:
Thanks, looks good to me - approved! Should I merge this now or wait for the cg-style update?

line47:
The cg-style update is to fix the spacing that shows at the bottom of the page when the content is shorter. This PR isn't necessarily dependent on that happening in tandem.

layouts/partials/footer.html
```diff
@@ -7,7 +8,7 @@
              <a class="usa-footer-primary-link" href="/docs/help/">cloud.gov help</a>
            </li>
            <li class="usa-width-one-fourth usa-footer-primary-content">
-             <a class="usa-footer-primary-link" href="https://github.com/18F/cg-site">Edit this page</a>
+             <a class="usa-footer-primary-link" href="https://github.com/18F/cg-site">Source code <br>(public domain)</a>
```
mogul, line 11: I think @brittag 's intent was to make this a link to go edit this particular page, as used to be possible via the "Refine this page" item in the menu; is that right? 

That menu item is gone now, and simply linking to the top of the repository here means that people have to understand Hugo and go navigate through the repository to find the right source file to even to suggest a typo fix. That adds enough friction that we're likely to get fewer contributors! (At least, it's enough friction that it dissuades me, and I imagine others feel the same.)

layouts/partials/docs-menu.html
```diff
@@ -48,9 +48,7 @@
             <li>
               <a href="/docs/help/"><i class='fa fa-fw fa-life-ring'></i> <span>Help</span></a>
             </li>
-            {{ if .IsPage }}
-            {{ $File := .File }}  {{with $File.Path }}<li><a href="https://github.com/18f/cg-site/edit/master/content/{{ $File.Dir }}{{ $File.LogicalName }}" target="blank"><i class='fa fa-edit fa-fw'></i> Refine this Page</a> </li>{{end}}
-            {{ end }}
+
```
mogul, line 51: This is the line for which there's no replacement functionality.

layouts/partials/overview-menu.html
```diff
@@ -45,9 +45,7 @@
               </a>
             </li>
             {{end}}
-            {{ if .IsPage }}
-            {{ $File := .File }}  {{with $File.Path }}<li><a href="https://github.com/18f/cg-site/edit/master/content/{{ $File.Dir }}{{ $File.LogicalName }}" target="blank"><i class='fa fa-edit fa-fw'></i> Refine this Page</a> </li>{{end}}
-            {{ end }}
+
```
mogul, line 48: And here's the other place it was.

brittag:
Ah good point @mogul - I'd like the site to have both the "Refine this page" item in the menu (for docs/overview pages) and the "Source code (public domain)" link in the footer, and I missed that this PR removed the "Refine this page" item. I'll file an issue for putting it back.

jameshupp:
Re. the multiple contact links: I agree on the email address. Not sure whether help/contact are independently meaningful text (and the page is currently titled `Help`) so I'm okay with having those both exist. But we actually don't want people to email the inquiries inbox if they haven't filled out the pre-inquiry form that's linked on the contact page. It'll require a super easy fix to the grid, too. I can open a PR.

jameshupp:
Never mind above. Looks like #687 is tackling this.

PR #656, opened 2016-12-23
Homepage new copy

jameshupp:
## Dependencies
 - [x] Merge #661 into the `homepage_new_copy` branch.
 - [x] Merge [cg-style 237](https://github.com/18F/cg-style/pull/237) into `master` in that repo.

## Changes
* New copy for the homepage, streamlined and focused on customer business. We’re reserving all details about how the platform works and the technology it’s based on to internal content.
* Updated contact.html, which is a partial that at some point we probably need to rename. It includes a number of things, but it’s basically the full set of “next steps” a person might take who is interested to some degree or another
* Lower case ‘g’ in the government banner at the top. This is consistent across 18F products. That’s not necessarily a reason to do it in cloud.gov, but if nobody has a reason to oppose it I want to keep this change.

## Review
* @nikzei 
* @vz3 (or anyone else from Navigator — see Notes below)

## Notes
* I’m going to file an issue for the folks from Navigator momentarily. There are a couple design issues that the new content presents for the homepage. *Please do not merge until we incorporate those changes.*

PR #661, opened 2016-12-27
New homepage layouts/copy [FOR REVIEW]

thisisdano:
**A pair with https://github.com/18F/cg-style/pull/237 — they need to update together for everything to work properly**

OK, so I was working from this PR (https://github.com/18F/cg-site/pull/656), trying to work with the new copy and bring the homepage up to current style standards.

In doing so, I updated the style/size of some of the copy elements, which affected their length and presentation. Because I am a dumb idiot who can't keep himself from meddling, I rewrote some of the copy to fit with the new size/flow. (I am truly sorry, but I honestly cannot help myself — I think it's a compulsion.)

Anyhow, I've cleaned up the styles and removed the obvious layout errors. 

Copywise — I hope you can see my meddling as a good-hearted attempt to be helpful, but I understand if you see it otherwise.

cg-homepage-changes

### DIFF
--- a/layouts/index.html
+++ b/layouts/index.html
@@ -14,7 +14,7 @@
   <div class="usa-grid section-bg-parent">
     <div class="usa-width-whole billboard  section-bg">
       <h1 class="billboard-display_text">A Platform as a Service for government teams</h1>
-      <h2 class="billboard-second_text">Focus on the things you want to build. cloud.gov gives you a powerful environment to build them in.</h1>
+      <h2 class="billboard-second_text">Focus on what you want to build: faster, simpler, and more secure.</h1>
     </div>
   </div>
 </section>
@@ -23,11 +23,9 @@
   <a name="about"></a>
   <div class="usa-grid">
     <div class="section-content usa-content">
-      <h1>cloud.gov is a Platform as a Service built specifically for government work.</h1>
-      <p>
-        It’s a secure, fully compliant cloud environment where your team can build, manage and
-        release digital applications. And because all of cloud.gov’s security controls are
-        documented in an open source and software-friendly way, it’s Compliance as a Service, too.
+      <h1 class="section-title">We understand the compliance requirements necessary for ATO.</h1>
+      <p class="section-displaytext">
+        The cloud.gov platform’s flexible, machine-readable, always up-to-date compliance documentation automates huge parts of the ATO process. Imagine a secure cloud environment where your team can build, manage and release digital applications — with platform compliance built in. This is cloud.gov.
       </p>
     </div>
   </div>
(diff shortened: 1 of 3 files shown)


jameshupp (approved):
Looks good!

If I read [cg-style pull 237](https://github.com/18F/cg-style/pull/237/files) correctly, looks like the changes here are generic enough that we can reuse them even if we have to move things between sections? (Trying to think of a future in which we decide the content should be reordered again.)

jameshupp:
@thisisdano Can this be pulled in right away, or do you need to do cg-style 237 first? What's the better hygiene?

thisisdano:
It really needs the cg-style PR to go through first — I'll push to get that merged asap

thisisdano:
It's a bit too complicated to sync site and style, imo

thisisdano:
In terms of generic-ness and flexibility, it _should_ be a bit flexible but I'm not 100% sure that reordering or recoding wouldn't have unpredictable consequences.

thisisdano:
And... wrt `cg-style` — we can get the changes merged soon (I think) but they won't actually be _live_ until whenever marco does the next cg-style release (usually wednesday) — so it would make sense to hold off until we know the new styles are available.

jameshupp:
How about this merge? Not into `master` but into `homepage_new_copy`? Still better to hold off, or should I merge it in and state that `cg-style` is a dependency before merging #656?

thisisdano:
This "sub-merge" should be fine — but it just won't totally look right locally unless you are running `dw-site-style` in `cg-style`

layouts/index.html
```diff
@@ -42,21 +40,9 @@ <h2 class="billboard-second_text">Focus on the things you want to build. cloud.g
...
     <div class="three_up">
-      <div class="three_up-part">
-        <i class="hex_icon hex_icon-primary"
-          style="height: 5.625rem; width: 4.9rem">
-          <svg class="icon">
-            <use xlink:href="assets/img/cloudgov-sprite.svg#i-shipping"/>
-          </svg>
-        </i>
-        <h3>Build great things in the cloud with the expertise you already have.</h3>
+      <div class="three_up-part b-blue bt2px">
```
msecret: `b-blue bt2px`, look like utility classes, shouldn't they be removed?

msecret (changes requested):

layouts/index.html
thisisdano: 'tis true — I built this without thinking about that

thisisdano: looks like I'll have to add some new classes to the ol' master

thisisdano:
Fixed utility class stuff — but now there's a new PR needed in cg-style. 
https://github.com/18F/cg-style/pull/240

jameshupp:
Not totally sure if this "sub-merge" as Dan called it is ready to go based on @msecret's comment and @thisisdano's subsequent changes but if it is feel free to go ahead any time (or get someone else to do so, for appropriate merging's sake)! I'm going to be out for a few days so can't handle it myself for a little bit.

adborden (approved):

PR #668, opened 2016-12-28
Elide image source in rendered output

mogul:


brittag:
Hm, this doesn't work when I try it locally:

screen shot 2016-12-28 at 4 50 39 pm

mogul:
I was following @berndverst's pointer and it seems like it should be supported based on the docs: https://golang.org/pkg/text/template/#hdr-Actions 

@berndverst any other ideas for accomplishing this?

brittag:
Going to close this up for now - maybe we can figure this out later! It's ok to have a HTML comment, so this isn't a blocker.

PR #673, opened 2016-12-28
Fix footers in docs and overview sections

brittag:
This should make the footers of pages such as https://cloud.gov/docs/services/ and https://cloud.gov/overview/overview/what-is-cloudgov/ consistent with the rest of the site - see https://github.com/18F/cg-design/issues/25#issuecomment-269403642 for context.

cc @line47 - up for review by anyone.


line47 (approved):
This looks great!

PR #675, opened 2016-12-29
Clarify some homepage text

brittag:
* Spell out Authority to Operate
* Make question more user-centered
* "Your engineering approach values clarity" is a little bit nonspecific so I removed that part


nikzei (approved):

nikzei (commented):
- (line 28) Great for us to write out Authority to Operate;
- (line 90) I'm not sure that "is cloud.gov right for you?" communicates the right message, which really is more like, "do you meet our requirements?" Perhaps there is a softer way to deliver this message than the way we currently do it? 
- (line 97) great

PR #680, opened 2016-12-31
Update docs about multiple instances

brittag:
* Explain on multiple instances page that you should set up a way to share session data between instances ([to help prevent this kind of problem](https://cloudgov.statuspage.io/incidents/xy5tjhhl7f7y))
* Note in production-ready guide that setting up multiple instances helps prevent downtime/errors caused by platform updates

cc @adborden @jmcarp

### DIFF
--- a/content/docs/apps/multiple-instances.md
+++ b/content/docs/apps/multiple-instances.md
@@ -5,23 +5,25 @@ menu:
     parent: advanced
 title: Running multiple instances
 ---
-## Starting Multiple Instances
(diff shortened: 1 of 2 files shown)


jmcarp:
To me, this crosses from explaining how the platform works to general best practices--might be a better fit for the development guide instead of the cloud.gov docs?

mogul (approved):
It's only a sentence or two, and we don't really have any more appropriate place to put a reminder where folks using cloud.gov (who may not realize the requirements for scaling apps horizontally because they're new to cloud) might see it.

PR #686, opened 2017-01-04
link to Cloud Foundry Tasks documentation

afeld:


mogul:
I'm not especially happy with the change here because running SSH really is a separate thing. The replacement ends up with three bullets on the page where one of them points to something below and the other two shoot off to other places, with no context about which you would use for what. It just feels weird! So I'm going to reach for the @brittag on this one...

brittag:
When a cloud.gov user wants to run a one-off task, what are the cases where they should use CF Tasks instead of using SSH or deploying a short-lived app?

If we define that, we can gently restructure this page (and related pages) to start with the user need and deliver the solution, along the lines of: "You want to do x? Here's how to choose the right tool and use it."

mogul:
- SSH is generally for hands-on tinkering or (in the case of port-forwarding) interacting with a brokered service. 
- Tasks are good for jobs that are automated, and may need to run periodically or in a repeatable way, but don't need to stay up consuming resources and costing money when they're not running. They're also good for long-running batch jobs, eg a big data processing job that has no user-facing aspect.

mogul:
(Tasks are also the foundation for a future "serverless" application mode, where the app only spins up long enough to handle one request, then is terminated.)

afeld:
> SSH really is a separate thing

I disagree. I would visit this page with the question "I have a command related to my live application that I need to run...how should I do it?"

> what are the cases where they should use CF Tasks instead of using SSH or deploying a short-lived app?

This doesn't directly answer your question, but may help:

Method | East/West | GovCloud
--- | --- | ---
CF Task | n/a | good - runs in its own container
`cf-ssh` | flaky | n/a
`cf ssh` | n/a | good - runs inside the application instances themselves
Short-lived app | works, but tedious; runs in a separate instance, and requires a new manifest | <-- (ditto)

mogul:
That's addressing the reliability/isolation, but I think we're better off thinking about this from a user's perspective.

- "I need to monkey around inside an app instance to figure out what's wrong or manually mess with one of the bound services as if I was my app" --> `cf ssh`
- "I want to work with one of my app's bound services" --> `cf ssh` with port-forwarding (with a hefty recommendation for @afeld's new service connect CLI plugin!)
- "I want to run a non-interactive process that [generates a report|cleans up garbage|mails people|processes some data] --> Tasks

afeld:
@mogul What I was getting at with the table is that the considerations may be a bit different for East/West vs. GovCloud...not sure how to indicate that cleanly with the user stories.

afeld:
Took another pass:

screen shot 2017-01-10 at 12 06 04 pm

afeld:
Bump!

brittag (approved):
Thank you Aidan!

PR #692, opened 2017-01-05
Multiple URIs are also needed

sharms:


jmcarp:
Now that we're able to provision uaa clients with the uaa-credentials-broker, can we drop this page altogether?

mogul:
Yup, and I believe you already did that in another PR.

PR #700, opened 2017-01-07
Streamline "how it's used"

mogul:
Reduce redundancy and address the reader more directly


brittag (approved):

brittag:
Yay, this is an improvement, thanks! I added one commit with a tiny change, to switch a plain link to a fellow overview page to a relative link.

PR #711, opened 2017-01-23
Remove outdated alpha note from header of site

brittag:
The definition at https://18f.gsa.gov/dashboard/stages/#alpha no longer applies, and "beta" doesn't really either, so here's a PR to remove that bit of the website header. The left-aligned "official website" note matches the other ones on 18F websites (such as https://18f.gsa.gov/); no need to change that part.


sharms (approved):

sharms:
Let's merge this when comms finishes

PR #731, opened 2017-02-02
Change sandbox cutoff date from 1/15 to "soon"

suprenant:
I'm not sure of our protocol for updating updates that have already posted, but this is a correction that more accurately reflects our current state regarding sandbox cutoff dates.


brittag:
Thanks Andrew, you're right this doesn't reflect reality right now and should be fixed. For clarity, instead of an inline edit I'd suggest we add a note at the top of this post saying something like "Update: We decided to postpone this cutoff date. We'll announce a new one." (But said more nicely and accurately than that - I'm not sure what the details should say here.)

(In general I think it's ok to edit old posts with minor edits. For edits that change something substantial, making a more visible update note can help prevent "but I thought I read something different here" confusion.)

suprenant:
@brittag ok, sounds good! I went ahead and updated this PR.

nikzei (approved):

PR #738, opened 2017-02-06
clarify that the managed CDN service is available in both environments

afeld:
Per https://github.com/18F/cg-migration/issues/32#issuecomment-272006460. The cloud.gov team might consider removing the Manual method from the page just to be less confusing, but I'll leave that up to y'all.

/cc @cmc333333 @LindsayYoung


afeld:
Just noticed with @jcscottiii that it might not be enabled in East/West by default? Might need to add a note about that (or just enable it by default).

nikzei (approved):

brittag:
@jcscottiii Would you be up for following up on that - figuring out if this page should have a note or seeing if it should be enabled by default? (Totally ok to re-delegate it to somebody else.)

Issue #679, opened 2016-12-30
consolidate information about setting up custom domains


afeld:
* https://cloud.gov/docs/apps/custom-domains/#managed-service-method
* https://cloud.gov/docs/services/cdn-route/

brittag:
We should also figure out whether to rename "CDN service" to "Custom domain service".

mogul:
We shouldn't, because that's not what CDN refers to in this case. Instead, it's [Content Delivery Network](https://en.wikipedia.org/wiki/Content_delivery_network) by virtue of the fact that it's using Cloudfront in the implementation. It just happens to also be a somewhat convenient way to support custom domains, albeit one that [we're considering replacing](https://gsa-tts.slack.com/archives/cloud-gov-atlas/p1483757159003615).

brittag:
Done as part of https://github.com/18F/cg-site/issues/752 !

nikzei:
marked as closed - I'm archiving.

PR #752, opened 2017-02-09
Update cdn broker docs.

jmcarp:
To reflect changes being made to the broker in https://github.com/18F/cf-cdn-service-broker/pull/65. I have a feeling I could be clearer about non-cloud.gov apps here, suggestions welcome 😄


brittag:
@jmcarp What do you mean by "non-cloud.gov apps" and "not managed by cloud.gov"? Do you mean apps that have domains that aren't at `*.cloud.gov` ? I'll also suggest to Skyporter to take a look at this, thanks!

jmcarp:
I was right--it is confusing! The distinction here is between apps that are run by cloud.gov and everything else. "Everything else" could include a public s3 bucket, an app hosted on another platform, etc.

cnelson:
@jmcarp does this line need updating?  https://github.com/18F/cg-site/pull/752/files#diff-a1e0116cc431d6c48977db5889604163R68  create only needs domain now right?

jmcarp:
@cnelson: good call, I dropped those lines.

@brittag: any suggestions about clarifying what we mean by externally managed applications?

content/docs/services/cdn-route.md
```diff
@@ -74,6 +76,17 @@ shows the old content after the DNS changes propagate.
...
+## External services and applications
+
+To create a custom domain for a service or application that is not a cloud.gov tenant application (such as a public S3 bucket), pass the `origin` option:
+
+```bash
+cf create-service cdn-route cdn-route my-cdn-route \
+    -c '{"domain": "my.example.gov", "origin": "my-app.external-example.gov"}'
+```
+
+There's no need to create a private domain for externally managed services and applications.
```
brittag: This line is a bit confusing still - we should work a bit on clarifying it (for example - "in other words, you don't need to do the `cf create-domain` step first")

jmcarp: That sounds good to me 😄

brittag:
I added a commit with clarifications for a few more things. I also added a "Troubleshooting" section - it's not a very _good_ troubleshooting section, and we can remove it if it doesn't seem helpful, but I figure it might help and we can expand it. @jmcarp, can you look at this last commit, and maybe add a typical error message a person might run into if they use the old instructions after this change gets released?

So, next thing to consider: what kind of notification do we need to give people about these new instructions? Factors, based on Josh's helpful explanations to me:

* Not having this is blocking at least two migrations, so this release (docs + technical change) should happen this week if possible.
* For people with service instances already set up (with live custom domains), they don't need to do anything - everything will continue normally.
* People use these instructions infrequently - the main users right now are people migrating applications.
* If people create or update a domain the old way after we release this change, they're either going to get an error message or nothing will happen. In both cases, they'll probably just come to support. Ideally we can just link them to the Troubleshooting section.

So I'm thinking we just release this, write an at-channel message in #cloud-gov-support, and include a note in our next /updates/ post+email, rather than bug all of our users with a proactive email notification about it.

brittag:
I just noticed that another thing needs to happen too: https://cloud.gov/docs/apps/custom-domains/#managed-service-method needs to be consolidated with these instructions!

brittag:
@jmcarp @nikzei: Would either or both of you be up for working on a commit that merges together the best aspects of both sets of instructions into one authoritative set of instructions? That would close https://github.com/18F/cg-site/issues/679 as well. :)

wslack:
One tiny note: I would suggest that people check these create commands with a second party because the provisioning process can't be immediately stopped if there's a typo. (Learned this the hard way.)

content/docs/services/cdn-route.md
```diff
@@ -8,30 +8,40 @@ description: "Custom routes with HTTPS certificates using Amazon CloudFront"
...
+`path` | *Optional* | The path for the application within the main domain supplied | `""`
+`insecure_origin` | *Optional* | Read the application over HTTP instead of HTTPS | `false`
 
-## How to create an instance
+## How to create an instance of this service
 
-To create a service instance, run the following command:
+*Use these instructions for cloud.gov tenant applications. If you're creating a custom domain for something else (such as a public S3 bucket), see [external services and applications](#external-services-and-applications).*
+
+Note: CDN service instances cannot be updated or deleted until they have been successfully configured. Until this issue is resolved, please double-check all commands related to provisioning and updating instances.
```
brittag: What do you mean by "CDN service instances cannot be updated or deleted until they have been successfully configured"? How would you manage to successfully configure this if you make a mistake?

content/docs/apps/custom-domains.md
```diff
@@ -69,47 +69,4 @@ Or:
...
-
-    ```
-    $ cf service my-cdn-route
-
-    Last Operation
-    Status: create in progress
-    Message: Provisioning in progress; CNAME domain "my.domain.gov" to "d3kajwa62y9xrp.cloudfront.net."
-    ```
-
-1. Repeat last step until the CloudFront distribution is provisioned. Note this can take up to 30 minutes.
```
brittag, line 110: Do we need this step?

content/docs/apps/custom-domains.md
```diff
@@ -69,47 +69,4 @@ Or:
...
-1. Create/update your DNS configuration.
```
brittag, line 111: Is there anything useful in this step we should salvage?

content/docs/apps/custom-domains.md
```diff
@@ -69,47 +69,4 @@ Or:
...
-    * In AWS Route 53, this will (confusingly) be an `A` record with an `Alias` to `<something>.cloudfront.net`.
-    * If you can, wait to **do this step during low-traffic hours**. It can take up to 15 minutes before the certificate is in place, and visitors will get a certificate error when visiting the custom domain in the meantime.
-1. Wait for the DNS changes to propagate.
```
brittag, line 114: Last step that seems missing from the new instructions - do we need this one?

content/docs/services/cdn-route.md
```diff
@@ -8,30 +8,57 @@ description: "Custom routes with HTTPS certificates using Amazon CloudFront"
...
+`domain` | *Required* | Your custom domain (or domains separated by commas) |
+`origin` | *Optional* | *Don't put this in your command for cloud.gov tenant applications.* For services/applications that are not cloud.gov tenant applications ([more info](#external-services-and-applications)): the origin root URL of the application |
+`path` | *Optional* | The path for the application within the main domain supplied | `""`
+`insecure_origin` | *Optional* | Read the application over HTTP instead of HTTPS | `false`
+
+## How to create an instance of this service
+
+*Use these instructions for cloud.gov tenant applications. If you're creating a custom domain for something else (such as a public S3 bucket), see [external services and applications](#external-services-and-applications).*
+
+Before you begin, note that once you create a CDN service instance, you can't update or delete it until it has been successfully configured. We consider this a bug and plan to make mistakes easier to fix, but for now, consider checking your commands with a teammate to help you avoid typos and mistakes.
```
brittag: I tried updating this to be more user-friendly, but I'm not sure I got it right, and I still have a question about "you can't update or delete it until it has been successfully configured" - how would you manage to successfully configure this if you make a mistake?

brittag:
Ha, I put the wrong commit message in - "Trim extra whitespace" is actually a commit with a bunch of changes for review. :) Tried to salvage some info that seemed useful from the custom domains page.

content/docs/services/cdn-route.md
wslack: The half-hour long process to "finish" provisioning a route has to complete before you delete one, which makes correcting any typos a bear.

content/docs/services/cdn-route.md
jmcarp: This is a lousy answer, but you'd have to either contact support or wait for cloud foundry to give up trying to provision the instance, which iirc takes a week.

content/docs/apps/custom-domains.md
jmcarp: I think this is covered by:

> After the record is created, wait up to 30 minutes for the CloudFront distribution to be provisioned and the DNS changes to propagate. Then visit your custom domain and see whether you have a valid certificate (in other words, that visiting your site in a modern browser doesn't give you a certificate warning).

brittag: Do we need the detail about AWS Route 53 or doing this during low-traffic hours?

jmcarp: I don't think we need route53 here--it's potentially confusing, since not all users are going to use it at all. The low-traffic point is also pretty opinionated--maybe it's useful to say that opening the domain will show an invalid certificate until the changes propagate through aws.

content/docs/services/cdn-route.md
brittag: OK, I added a commit with this info! I'm still confused about whether we need any of the old details from the custom domains page, but if you don't think we need them, I'm ok with that.

brittag (approved):
I consider this done enough from a content perspective - thanks @jmcarp for your patience with my edits and questions! Will let others decide when to merge this. We can close https://github.com/18F/cg-site/issues/679 when we merge this.

jcscottiii (approved):

PR #757, opened 2017-02-10
Clarify top of GovCloud page for new readers

brittag:
Emphasize that new users should use GovCloud instructions.


brittag:
This would help https://github.com/18F/cg-uaa/pull/37 make more sense as well.

sharms (approved):

PR #759, opened 2017-02-10
Fix for /updates layout regression

msecret:
Some cg-style work this week unexpectedly changed some layout on
the updates page. This is a fix for the strange spacing issues
between and above articles.

The fix is to change the `panel-new` class to just the `panel` class
as the `panel-new` class was deprecated and mistakenly removed in
cg-style. There isn't actually a large difference between the two classes,
just some spacing differences:

```
- .panel-new {
-   background-color: $color-white;
-   margin-top: $grid-4;
-   padding: $grid-4;
- }
```

```
.panel {
   background-color: $color-white;
   margin-bottom: $grid-4;
   padding: $grid-3 $grid-3 $grid-4;
}
```

I'm of the opinion that the difference between the two doesn't warrant
having a second class anyway.

Another change will have to come next week when the `panel-new-date`
class is removed from cg-style (expecting this change soon).

image

image

### DIFF
--- a/layouts/updates/li.html
+++ b/layouts/updates/li.html
@@ -1,5 +1,5 @@
 
-   <div class="usa-width-two-thirds usa-end-row panel-new content-text">
+   <div class="usa-width-two-thirds usa-end-row panel content-text">
       <span class="panel-new-date">{{ .Date.Format "January 2, 2006" }}</span>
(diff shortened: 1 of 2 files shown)


jameshupp (approved):

brittag:
Yay thank you @msecret - appreciate the super fast fix!

PR #762, opened 2017-02-13
Document consolidated sandbox layout.

jmcarp:
As we move to ephemeral sandboxes, we're also switching to a single sandbox organization, with spaces named by username. This updates the docs to remove the current "sandbox-${AGENCY}" layout. Should be merged with https://github.com/18F/cg-sandbox-bot/pull/25.

### DIFF
--- a/content/docs/getting-started/setup.md
+++ b/content/docs/getting-started/setup.md
@@ -21,18 +21,18 @@ cloud.gov is based on the Cloud Foundry open source project, so cloud.gov uses t
 {{% govcloud %}}
 **For all accounts in the GovCloud environment:**
 
-Run `cf login -a api.fr.cloud.gov --sso` 
+Run `cf login -a api.fr.cloud.gov --sso`
 
 Then it'll say `One Time Code ( Get one at `[`https://login.fr.cloud.gov/passcode`](https://login.fr.cloud.gov/passcode)` )` -- visit that link in your browser to get your code, copy and paste the code into the command line (no typing indicators will show), and enter it.
 {{% /govcloud %}}
 
 {{% eastwest %}}
-**If you're at GSA or EPA and you log in with your agency account:** 
+**If you're at GSA or EPA and you log in with your agency account:**
 
 Run `cf login -a api.cloud.gov --sso`
 
 Then it'll say `One Time Code ( Get one at `[`https://login.cloud.gov/passcode`](https://login.cloud.gov/passcode)` )` -- visit that link in your browser to get your code, copy and paste the code into the command line (no typing indicators will show), and enter it.
-    
+
 **If you log in with a cloud.gov account that has its own password** (including `ORGNAME_deployer` accounts): run `cf login -a api.cloud.gov`
 {{% /eastwest %}}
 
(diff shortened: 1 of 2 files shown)


content/docs/getting-started/setup.md
```diff
@@ -63,13 +63,12 @@ Start with the following `cf target` command:
 cf target -o <ORG> -s <SPACE>
 ```
 
-Your `<ORG>` is a Cloud Foundry _organization_ named "sandbox-&lt;agencypart&gt;", where &lt;agencypart&gt; is whatever comes right before `.gov` or `.mil` in your
-e-mail address. For example, this may be `sandbox-gsa` or `sandbox-epa`. In most cases, your `<SPACE>` is the part of your email before the `@`, e.g. `firstname.lastname`. Cloud Foundry _spaces_ let applications run independently within an organization.  
+Your `<ORG>` is a Cloud Foundry _organization_ named "sandbox". In most cases, your `<SPACE>` is your email address. Cloud Foundry _spaces_ let applications run independently within an organization.
```
brittag: Are there any cases now in which your `<SPACE>` is not your email address? If it's now all cases, we should say all cases, and if it's still just "most" cases, we should explain the exceptions if we can.

brittag: Also we should probably switch "Cloud Foundry" in these two sentences to "cloud.gov".

brittag:
Yay! The info here looks pretty good (just put in a couple comments), but before we merge this, we need the comms plan prepared for notifying current sandbox users, since this will be a breaking change.

It would also be really helpful to make the standalone sandbox page at the same time (drafted at https://github.com/18F/cg-site/blob/sandbox-content-development/content/overview/overview/free-limited-sandbox.md) and cross-link that here.

content/docs/getting-started/setup.md
jmcarp: Updated. As an atlas person, I prefer saying "cloud foundry" over "cloud.gov" for concepts that aren't specific to cloud.gov, but agreed that "cloud.gov" will be easier for new users to understand.

jmcarp:
@brittag: if it would be easier, we could merge these changes into your WIP branch and close this PR. Also, should we start talking about sandboxes getting wiped out at the end of the month in that patch, or do we want to delay that until we have an announcement drafted?

brittag:
🤔  Sounds like we have three things we should do:

1) Make a docs page about sandboxes for general reference and usefulness.
2) Document new breaking instructions for how to access your sandbox.
3) Document new breaking change that sandboxes are becoming ephemeral.

Let's keep 1 separate (I'll make a WIP PR for it for visibility) and merge it first.

And let's work on 2 and 3 together in a technical sense + docs sense + comms sense, since both require heavily notifying people. Since they're related, it seems better to notify people once than twice.

mogul:
I'm a little confused about what should be happening in this PR now given work happening elsewhere. Can someone summarize the current state?

brittag:
I'm not sure either - I'm guessing the "Document new breaking instructions for how to access your sandbox" task (including reviewing & merging this PR + making sure users are adequately informed of the change) should probably become part of the "Sandboxes are ephemeral" card at https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-2717

brittag:
(Looking at that card, there are links to this in the comments there, but it's not part of the ACs yet.)

jmcarp:
@brittag @mogul: do we still even want to consolidate sandboxes into a single org, or can we close this?

brittag:
@jmcarp I still want us to pick up "Sandboxes are ephemeral" again in the future; I can't remember if consolidating sandboxes is required for that feature - if so, I'd keep it.

brittag:
On second look, the changes here are mostly not relevant anymore because we've shifted these docs around, so we'd have to approach this PR fresh anyway - closing this.

PR #765, opened 2017-02-14
Attempt at defining "system" for readers of the rates page

brittag:
A customer found the mention of "system" confusing on the rates page, so here's an effort at improving this.

Changes:

* Add links to the system definition from the rates page.
* Try to clarify the "System" definition on the terminology page with more user-centered language. 
* Add compliance info to "System" definition. This may make things _more confusing_ for readers who are working with prototyping orgs or who aren't very familiar with compliance, but it should be helpful for some readers.

For your amusement: I was tempted to go in depth with detailed references about how FISMA defines "information system", but the legal definition is very formal and probably not helpful for this specific context: *a discrete set of information resources organized for the collection, processing, maintenance, use, sharing, dissemination, or disposition of information*.

### DIFF
--- a/content/overview/pricing/pricing-terminology.md
+++ b/content/overview/pricing/pricing-terminology.md
@@ -16,9 +16,11 @@ These terms, among others, define a number of products and services offered by c
 
 ### System
 
-A “System” roughly corresponds to a product or project team, including all the individual spaces, apps, and services necessary to realize their service, across as many environments as needed.
+A “system” roughly corresponds to a product or project team, including all the individual spaces, apps, and services necessary to deliver their product to users.
 
-A system maps to an [“org” in Cloud Foundry](http://docs.cloudfoundry.org/concepts/roles.html#orgs).
+A system maps to an [“org” in cloud.gov and Cloud Foundry](http://docs.cloudfoundry.org/concepts/roles.html#orgs).
+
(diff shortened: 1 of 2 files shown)


content/overview/pricing/pricing-terminology.md
```diff
@@ -16,9 +16,11 @@ These terms, among others, define a number of products and services offered by c
...
+A “system” roughly corresponds to a product or project team, including all the individual spaces, apps, and services necessary to deliver their product to users.
 
-A system maps to an [“org” in Cloud Foundry](http://docs.cloudfoundry.org/concepts/roles.html#orgs).
+A system maps to an [“org” in cloud.gov and Cloud Foundry](http://docs.cloudfoundry.org/concepts/roles.html#orgs).
+
+*Recommendation:* If your product needs Authority to Operate (ATO) from your agency, we suggest that your cloud.gov “system” should align with your ATO system boundary. (In other words, your cloud.gov org shouldn't contain two or more products that each need their own ATO, because that would make your compliance process more difficult.) If you have a very complex product with multiple sub-systems, you may choose to build that with multiple cloud.gov orgs/“systems”. For formal guidance on system boundaries, see [NIST Special Publication 800-18, section 2](http://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-18r1.pdf).
```
mogul: I think that this is too narrow a definition. The system boundary for ATO purposes likely won't include all of their staging and development spaces, only their production spaces. But we want them to feel free to include their staging and development spaces in their org! (This is analogous to the cloud.gov SSP not including information about the staging environment, because it's not inside the ATO boundary.)

mogul (changes requested):

content/overview/pricing/pricing-terminology.md
brittag: That makes sense - this would need more discussion and refinement before we could just stick it in the site. I removed this paragraph for now so that it doesn't block the other changes.

brittag:
@mogul I changed this to a more conservative PR, so it could use another review.

mogul (approved):

PR #766, opened 2017-02-15
Update service broker descriptions

adborden:
Part of https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-989

Let's make sure the we're on the same page about the service descriptions in the docs before updating them in the brokers. Once we have them set, I can copy them out to the individual broker repos.

These descriptions are based on the broker descriptions plus edits from @brittag.


brittag:
Do you mean for https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-989 ?

I'd like to see if Skyporter and other people might have a chance to review this as well (to prevent blocking on me for content reviews), but I'm confused right now about the goal of this branch since I'm not sure which card.

Is the task to write new descriptions in the docs pages that we like (using info from: the current descriptions in the CLI/dashboard, the current docs, the pending edits at https://github.com/18F/cg-deploy-s3-broker/pull/9 + https://github.com/18F/aws-broker/pull/27, and our best understanding of how to write good descriptions), which will then be carried to the CLI/dashboard to match?

adborden:
Oops, yes, https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-989 is the correct card.

The task is really to make sure the descriptions in the docs are consistent with the descriptions in the service brokers. And yes, the first part of that is to make sure we have the correct descriptions based on current descriptions in the brokers, the docs, and pending edits. I think having that conversation here makes the most sense. We can make suggested edits directly to the PR.

adborden:
... and correct, once we have the final copy I will submit new PRs for the individual service brokers.

brittag:
I took a quick look at the changes in this PR, and they all look like they'd improve the docs instead of adding confusion, so 👍  from me!

brittag (approved):

brittag:
Approved, but leaving unmerged for now since @adborden might want to update the metadata in the broker repos at the same time.

mogul (approved):

content/docs/services/elasticsearch17.md
```diff
@@ -4,7 +4,7 @@ menu:
     parent: services
 title: Elasticsearch 1.7
 name: "elasticsearch17"
-description: "[Elasticsearch](https://www.elastic.co/) version 1.7"
+description: "[Elasticsearch](https://www.elastic.co/products/elasticsearch) version 1.7 a distributed, RESTful search and analytics engine"
```
brittag: This needs something like a colon after the version number, and I suspect links aren't a good idea since they wouldn't show up properly in the CLI - `Elasticsearch version 2.3: a distributed, RESTful search and analytics engine`

brittag: (Similar for the other Elasticsearch and Redis descriptions)

content/docs/services/cdn-route.md
```diff
@@ -4,7 +4,7 @@ menu:
     parent: services
 title: CDN service
 name: "cdn-route"
-description: "Custom routes with HTTPS certificates using Amazon CloudFront"
+description: "Custom domains, CDN caching, and TLS certificates using Amazon CloudFront"
```
brittag: This should be `Custom domains, CDN caching using Amazon CloudFront, and TLS certificates` since the domains and TLS certificates aren't using CloudFront

brittag: Actually, just `Custom domains, CDN caching, and TLS certificates with automatic renewal` would be fine here - we can leave the CDN details to the documentation

content/docs/services/cdn-route.md
```diff
@@ -18,7 +18,7 @@ This service provides three key elements to support production applications:
 
 Plan Name | Description | Price
 --------- | ----------- | -----
-`cdn-route` | A CDN distribution with custom domain and free TLS certificate with auto renewal | Free
+`cdn-route` | Custom domains, CDN caching, and TLS certificates with auto renewal | Free
```
brittag: `Custom domains, CDN caching, and TLS certificates with automatic renewal`

content/docs/services/relational-database.md
```diff
@@ -13,16 +14,16 @@ If your application uses relational databases for storage, you can use the AWS R
...
 --------- | ----------- | -----
-`shared-psql`            | Shared infrastructure for Postgres DB               | Free
-`medium-psql`            | Dedicated Medium RDS Postgres DB Instance           | $0.115 / hr + storage
-`medium-psql-redundant`  | Dedicated Redundant Medium RDS Postgres DB Instance | $0.230 / hr + storage
-`large-psql`             | Dedicated Large RDS Postgres DB Instance            | $0.230 / hr + storage
-`large-psql-redundant`   | Dedicated Redundant Large RDS Postgres DB Instance  | $0.470 / hr + storage
+`shared-psql`            | Shared infrastructure for PostgreSQL DB               | Free
+`medium-psql`            | Dedicated medium RDS PostgreSQL DB instance           | $0.115 / hr + storage
+`medium-psql-redundant`  | Dedicated redundant medium RDS PostgreSQL DB instance | $0.230 / hr + storage
+`large-psql`             | Dedicated large RDS PostgreSQL DB Instance            | $0.230 / hr + storage
```
brittag: `instance` instead of `Instance`

content/docs/services/relational-database.md
```diff
@@ -13,16 +14,16 @@ If your application uses relational databases for storage, you can use the AWS R
 
 Plan Name | Description | Price
 --------- | ----------- | -----
-`shared-psql`            | Shared infrastructure for Postgres DB               | Free
-`medium-psql`            | Dedicated Medium RDS Postgres DB Instance           | $0.115 / hr + storage
-`medium-psql-redundant`  | Dedicated Redundant Medium RDS Postgres DB Instance | $0.230 / hr + storage
-`large-psql`             | Dedicated Large RDS Postgres DB Instance            | $0.230 / hr + storage
-`large-psql-redundant`   | Dedicated Redundant Large RDS Postgres DB Instance  | $0.470 / hr + storage
+`shared-psql`            | Shared infrastructure for PostgreSQL DB               | Free
```
brittag: Let's spell out "database" instead of "DB" in all of these descriptions. But specifically for this one, `Shared infrastructure for PostgreSQL DB` is a bit vague - what is it really? Is it really `Shared PostgresSQL database for prototyping (no sensitive or production data)` ?

content/docs/services/s3.md
```diff
@@ -4,7 +4,7 @@ menu:
     parent: services
 title: S3
 name: "s3"
-description: "Amazon Simple Storage System (persistent storage)"
+description: "Amazon S3 provides developers with secure, durable, highly-scalable object storage"
```
brittag: Can we make this one match the RDS one better?

brittag: More like `Secure, durable, highly-scalable object storage using Amazon S3`?

data/services/cdn-route.toml
```diff
@@ -1,3 +1,3 @@
 name = "cdn-route"
-description = "Custom routes with HTTPS certificates using Amazon CloudFront"
+description = "Custom domains, CDN caching, and TLS certificates using Amazon CloudFront"
```
brittag: This should match whatever we change above for CDN routes. Should we have to update both manually, or is there a way in Hugo to use this data inside pages?

brittag: Same comment/question for the other `.toml` items (They should match, and can we do it automatically?)

content/docs/services/relational-database.md
```diff
@@ -13,16 +14,16 @@ If your application uses relational databases for storage, you can use the AWS R
...
+`medium-psql`            | Dedicated medium RDS PostgreSQL DB instance           | $0.115 / hr + storage
+`medium-psql-redundant`  | Dedicated redundant medium RDS PostgreSQL DB instance | $0.230 / hr + storage
+`large-psql`             | Dedicated large RDS PostgreSQL DB Instance            | $0.230 / hr + storage
+`large-psql-redundant`   | Dedicated redundant large RDS PostgreSQL DB instance  | $0.470 / hr + storage
 `shared-mysql`           | Shared infrastructure for MySQL DB                  | Free
```
brittag: Same comment as for PostgresSQL - `Shared infrastructure for MySQL DB` is a bit vague - what is it really? Is it really `Shared MySQL database for prototyping (no sensitive or production data)` ?

brittag:
Looks like changes to the `uaa-auth` descriptions aren't in here, or am I missing them? Currently "Protect applications with cloud foundry authentication in the routing path" + "Just add to a route and it will request cloud foundry authentication before proceeding." - these would be good to update as well.

content/docs/services/redis28.md
```diff
@@ -14,7 +14,7 @@ cloud.gov offers [Redis](https://www.redis.io/) 2.8 as a service. **This service
 
 Plan Name | Description | Price
 --------- | ----------- | -----
-`standard` | Redis instance with unlimited storage | Free in Alpha
+`standard` | Redis instance with several GB storage | Free in Alpha
```
brittag, line 17: Is there any way we can make this more specific? How many GB?

content/docs/services/cloud-gov-service-account.md
```diff
@@ -4,7 +4,7 @@ menu:
     parent: services
 title: cloud.gov service account
 name: "cloud-gov-service-account"
-description: "Manage cloud.gov service accounts with access to your organization"
+description: "cloud.gov service accounts with access to your organization"
```
brittag: Thinking that "cloud.gov service accounts with access to your organization" doesn't explain what the service account is _for_ - instead of talking about access to your organization, could this be something that defines what a service account is, such as "cloud.gov service accounts for automated access by programs"?

content/docs/services/cloud-gov-service-account.md
```diff
@@ -16,7 +16,7 @@ To set up your application to be deployed with an automated system, you need a d
 
 Plan Name | Description | Price
 --------- | ----------- | -----
-`space-deployer` | Provision one account with "SpaceDeveloper" access to the space where the service is created | Free
+`space-deployer` | An account for continuous deployment, limited to a single space | Free
```
brittag: Suggest "A service account for continuous deployment, limited to a single space" to emphasize that this is a service account instead of a human account

content/docs/services/cloud-gov-identity-provider.md
```diff
@@ -4,7 +4,7 @@ menu:
     parent: services
 title: cloud.gov identity provider
 name: "cloud-gov-identity-provider"
-description: "Manage client credentials for authenticating cloud.gov users in your app"
+description: "Authenticate cloud.gov users in your app"
```
brittag, line 7: I wonder if we can find a more plain-language way of saying this - not sure if I'd understand "Authenticate cloud.gov users in your app" at first read if I didn't know what this service was. Maybe "Use cloud.gov authentication in your application as a login system for your users"?

jmcarp: We already had a meeting specifically to nail down this text. I'm wondering how much more time this is worth--are we making this clearer or talking in circles?

brittag: Hmm, it is a good question, thanks for asking. I'm honestly a bit lost in the various sources of info here - I'm not sure what the already-nailed-down text is (whether it's this text or other text somewhere else).

Overall I'm totally fine with merging the existing text in this PR - all of my comments should be considered enhancements and shouldn't be considered blockers.

@adborden suggested the PR would benefit from a bit of additional review to prevent future cycles of laborious metadata merges - happy to leave it up to him to decide whether to do more copy work or merge as-is. :)

content/docs/services/elasticsearch17.md
adborden: Right, the links will be stripped out for the actual service broker metadata. The links here will only be displayed on the docs site. I think we should keep them since they are useful in an HTML context.

adborden: BUT, these descriptions are actually for the page metadata (for things like url previews on slack) and are not converted to HTML so should not include links. I'll update that on all the pages.

brittag: Cool, that's ok with me!

content/docs/services/relational-database.md
adborden: > Shared PostgresSQL database for prototyping (no sensitive or production data)

Yeah, good call.

data/services/cdn-route.toml
adborden: The `.toml` files are what is used in the HTML table/pages and can be HTML/markdown content. The `description` in the markdown files are actually page metadata and should **not** contain HTML/markdown content.

The content _could_ be shared in jekyll, so I suspect there's a way to do it in hugo (although I'm not familiar with it). However, given that these fields contain different content types, I think it's better to keep them separate even if it means duplication.

content/docs/services/redis28.md
adborden: [10 GB](https://gsa-tts.slack.com/archives/cloud-gov-atlas/p1487111719009649) but this number was not selected with any meaningful testing or planning, so while it's in Alpha we want to keep it vague.

content/docs/services/cloud-gov-service-account.md
adborden: Yeah, I struggled with this because we only have a single plan, and it's kind of open ended what these could be used for, or what plans might appear in the future. I like your suggestion.

content/docs/services/cloud-gov-identity-provider.md
adborden: > Authenticate cloud.gov users in your app

This came from me. The thing that stuck out for me with the "Manage client credentials for authenticating cloud.gov users in your app" copy was the "manage" part. The service does no managing of credentials. It only provides you with client id/secret so that you can "Use cloud.gov authentication in your application as a login system for your users". Is that right @jmcarp?

Anyway, I'm happy to go ahead with all the suggested changes above. I don't mean to steamroll any previous thought and discussion for the identity-provider copy, but I would like another round to clarify the thinking behind it and maybe iterate on the existing text.

adborden: Hmm, although given an oauth token, potentially you could do much more than just use cloud.gov as a login. You could access any cloud.gov resources on behalf of the user. Do we have examples of what users are building with this service?

adborden:
@brittag for `uaa-auth` you mean the [experimental services](https://github.com/18F/cg-site/blob/master/content/docs/apps/experimental/experimental-services.md#authorization-proxy)? I wasn't going to touch those.

data/services/cdn-route.toml
brittag: Ah interesting! That seems ok to me.

brittag: Might be good to add a note about this bit in the cg-site README, so that we remember to update both next time we go through this process - up to you.

content/docs/services/redis28.md
brittag: Ah ha! I'd suggest explaining that on https://cloud.gov/docs/services/redis28/ - that it's currently 10 GB but may change.

content/docs/services/cloud-gov-identity-provider.md
brittag: I believe Tock currently uses this service - to log into Tock, you go through the cloud.gov auth flow. Not sure what else / what the other capabilities are.

brittag: Actually, I'm not sure whether Tock is using this or something else - I'll leave this question to somebody else.

brittag:
@adborden Yes, the experimental `uaa-auth` service. The other experimental servcies (Elasticsearch and Redis) are covered above, so I'd suggest sticking in updates for `uaa-auth` as well, if we can. Even if we just add "[Experimental]" to the start of each of those descriptions, so that people understand the alpha-ness when they look at the service in the dashboard and CLI.

PR #778, opened 2017-02-17
Suggest trying the CF sample apps as well

brittag:
Referencing https://github.com/cloudfoundry-samples without implying that we've checked or maintain them.


rogeruiz (approved):
:shipit:

PR #783, opened 2017-02-21
Update example tiny app name

brittag:
To reflect https://github.com/18F/cf-hello-worlds

This is currently wrong in the docs and might confuse new users.


mogul (approved):

mogul:
Jumped on the merge here because I saw it too and can confirm the fix.

PR #792, opened 2017-02-24
Clarifying that New Relic is an example service

brittag:
We mention it here since it's commonly used at 18F, but other teams may use other services.

### DIFF
--- a/content/docs/apps/production-ready.md
+++ b/content/docs/apps/production-ready.md
@@ -49,7 +49,7 @@ When running an application for development or testing, it is best to have a sep
 You want to receive alerts about application errors, downtime, and throughput issues.
 
 #### How
-* There are many external services that provide alerting. For example, New Relic provides application performance and availability monitoring with "Insights". You can set up and receive alerts on a variety of metrics.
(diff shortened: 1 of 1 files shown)


adborden (approved):

content/docs/apps/production-ready.md
```diff
@@ -49,7 +49,7 @@ When running an application for development or testing, it is best to have a sep
 You want to receive alerts about application errors, downtime, and throughput issues.
 
 #### How
-* New Relic provides extensive application performance and availability monitoring with "Insights". You can set up and receive alerts on a variety of metrics.
+* There are many external services that provide alerting. For example, New Relic provides application performance and availability monitoring with "Insights". You can set up and receive alerts on a variety of metrics.
```
adborden: FWIW, Insights is a big package containing several monitoring and analytics features. The availability monitoring is specifically called "Synthetics" within Insights. I would consider the performance monitoring a slightly more advanced topic, so we could encourage users to start with Synthetics, and then dig into setting up more advanced monitoring. Just a suggestion.

brittag: @adborden I simplified this suggestion on this page - that makes sense to me!

PR #797, opened 2017-02-27
Better warning about leaving a one-time app in place

mogul:


content/docs/getting-started/one-off-tasks.md
```diff
@@ -73,7 +73,7 @@ Note that this will not work for any command that is interactive.
     cf logs --recent task-runner
     ```
 1. If needed, use [`cf files`][] to collect any artifacts.
-1. Run `cf delete task-runner` to clean it up.
+1. Run `cf delete task-runner` to clean it up. **If you forget to do this, you might find your app runs again due to a restart in future!** (For example, if cloud.gov itself is updated, it may migrate your app to another runner by starting a separate instance there prior to removing the existing one.)
```
brittag: This is a super helpful note, thank you! One bit though is that "For example, if cloud.gov itself is updated, it may migrate your app to another runner by starting a separate instance there prior to removing the existing one." doesn't quite make sense to me (especially if I put myself in the shoes of a non-expert cloud.gov user) - it makes me wonder whether there would be other bad side effects I should take into account.

How about "**If you don't delete it, your short-lived app may automatically run itself again in the future.** cloud.gov sometimes automatically restarts apps as part of routine operations (such as platform updates), which can include restarting this kind of app if it hasn't been deleted."

mogul:
Good suggestion, added. Approve for merge?

brittag (approved):

PR #799, opened 2017-02-28
Release notes 9 3

jameshupp:


jameshupp (commented):
Thank you! My bad.

brittag:
The Concourse build for this hasn't run yet, so we should check with Atlas before merging this (I think that may be related to their current maintenance work) - we should get a green Concourse build before merging.

nikzei (approved):

PR #802, opened 2017-03-01
Document rotation of IAM user access keys

rogeruiz:
Part of secret rotation, this documentation covers versioning the access_key and secret_access_key during a secret rotation.

### DIFF
--- a/content/docs/ops/rotating-iam-users.md
+++ b/content/docs/ops/rotating-iam-users.md
@@ -5,15 +5,15 @@ menu:
 title: Rotating IAM users with Terraform
 ---
 
-### IAM Users infrastructure as a code
+### IAM Users infrastructure as a code with Terraform
 
 In the [18F/cg-provision](https://github.com/18F/cg-provision) contains the
 Terraform files used to maintain our AWS infrastructure as code. User modules
-are in the `terraform/modules/iam_user/*` path. Each module is leveraged in the
-`terraform/stacks/*` path.
+are in the `terraform/modules/iam_user/*` path. Each module is utilized from
+`stack.tf` and `output.tf` files in the `terraform/stacks/*` path.
 
 Outputs are for the `iam_user` modules are versioned with a `_vN` suffix with
-`(N)` representing the version number. When creating new keys, make sure that
+`N` representing the version number. When creating new keys, make sure that
 versions get incremented across all stacks when updating them.
 
 ### Rotating IAM user access key ids and secret access keys
@@ -27,6 +27,8 @@ current named version – the string after `resource "aws_iam_access_key"` – a
 the new named version. The version name is important as Terraform will
 automatically re-create `acccess_key_id` and `secret_access_key` based on the
 version name.
+
+#### Update the rotation iam_user module to have two access key resources
 
 Create a new sequential version of the output in the main
 `module/iam_user/rotation/output.tf` and `module/iam_user/rotation/user.tf` files.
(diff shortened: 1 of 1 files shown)


LinuxBozo:
I have a concern that without proper diff context, all these examples will lose a bit of meaning. Very hard to see from the examples what exactly is going on if you really aren't paying close attention.

rogeruiz:
Yeah very much agree. Maybe having more generic examples _with diffs_  makes more sense. @LinuxBozo 

The `iam_user/base` and `iam_user/rotation` will be two that make sense to include. But if we make the actual modules to update more generic we could even point to this as a start for creating `iam_user` modules. 🤷‍♂️

rogeruiz:
Noticed some typos. So this isn't ready to merge yet.

rogeruiz:
This should now be ready to be reviewed. @jmcarp @rememberlenny

content/docs/ops/rotating-iam-users.md
```diff
@@ -0,0 +1,155 @@
...
++  user = "${aws_iam_user.iam_user.name}"
++}
++
+ resource "aws_iam_user_policy" "iam_policy" {
+   name = "${aws_iam_user.iam_user.name}-policy"
+   user = "${aws_iam_user.iam_user.name}"
+
+```
+
+The example above leverages `_v2` as the now current version by removing the
```
cnelson: Should this be v2 / v3, not v1/v2 (based on your sample diff)

content/docs/ops/rotating-iam-users.md
```diff
@@ -0,0 +1,155 @@
...
++  user = "${aws_iam_user.iam_user.name}"
++}
++
+ resource "aws_iam_user_policy" "iam_policy" {
+   name = "${aws_iam_user.iam_user.name}-policy"
+   user = "${aws_iam_user.iam_user.name}"
+
+```
+
+The example above shows keys being rotated from `_v1` to `_v2`. The example
```
cnelson: Should this be v2 / v3, not v1/v2 (based on your sample diff)

content/docs/ops/rotating-iam-users.md
rogeruiz: good catch. thanks.

rogeruiz:
completely forgot to push up my changes. Should be resolved now @cnelson

cnelson (commented):
LGTM me, but I think @LinuxBozo should approve since he raised the initial issue.  Does this make more sense with the additional context?

rogeruiz:
True. I just wish we had a syntax highlighter for the diffs. 😢

content/docs/ops/rotating-iam-users.md
```diff
@@ -0,0 +1,160 @@
...
+ output "access_key_id_prev" {
+-  value = ""
++  value = "${aws_iam_access_key.iam_access_key_v2.access_key_id}"
+ }
+ output "secret_access_key_prev" {
+-  value = ""
++  value = "${aws_iam_access_key.iam_access_key_v2.secret_access_key}"
+ }
+ output "access_key_id_curr" {
+-  value = "${aws_iam_access_key.iam_access_key.access_key_id}"
```
LinuxBozo: Shouldn't this old value be v2?

content/docs/ops/rotating-iam-users.md
```diff
@@ -0,0 +1,160 @@
...
++  value = "${aws_iam_access_key.iam_access_key_v3.access_key_id}"
+ }
+ output "secret_access_key_curr" {
+-  value = "${aws_iam_access_key.iam_access_key.secret_access_key}"
```
LinuxBozo: This old value should be v2 as well, yeah?

LinuxBozo (changes requested):

jmcarp (approved):

LinuxBozo (approved):

PR #803, opened 2017-03-01
Update font path in `npm run *` scripts

rogeruiz:
This font path was changed as far back as
https://github.com/18F/cg-style/commits/c185b97a43f8f02808bddfc1be913e87f9382c36

Without this, the local building of the site was failing. 😞  But it works now :tada:

cc: @thisisdano, @msecret, @adborden

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "clean": "rm -rf ./public",
     "copy": "npm run copy-img && npm run copy-font",
     "copy-img": "cp -r node_modules/cloudgov-style/img/* ./public/assets/img/",
-    "copy-font": "cp node_modules/cloudgov-style/fonts/* ./public/assets/fonts/",
+    "copy-font": "cp node_modules/cloudgov-style/font/* ./public/assets/fonts/",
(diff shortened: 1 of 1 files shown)


adborden:
Hmm, my build seems to be working fine with the latest cloudgov-style. The npm package includes both `fonts` and `font` but I'm not sure which one is "right", so I'd like to clear that up first.

@rogeruiz what's the error you're seeing?

thisisdano:
The fonts in the `fonts` folder in `cg-style` are the correct faces — the ones in `font` are not. idk why there are two folders, though! (All the Source and Merriweather files could be cleaned out — they are (_should be???_) no longer in use

thisisdano:
Curious that this change would fix things rather than break them...

adborden:
Thanks @thisisdano that's what I'm seeing too. The commit history looks like we had tried to clean this up in Aug/Oct 2016, but missed `font/`, so `fonts` is definitely correct.

rogeruiz:
Oh strange. No wonder I saw weird fonts locally. Thanks for cleaning it up!

PR #810, opened 2017-03-06
Remove vestigial S3 docs

mogul:
With service-keys, it's no longer necessary to create an application just to get access to a raw bucket.


brittag (approved):
Yay, thank you for identifying this problem and fixing it!

PR #811, opened 2017-03-06
Be specific about languages

mogul:
- Mention every language by name so this turns up in search results
- Make sure it's clear that static binaries need to be for a 64-bit Linux kernel ABI
- Distinguish between .NET Core and .NET Framework to assist with evalution


brittag (approved):

brittag:
@mogul This is an improvement! I put in a small edit so I probably shouldn't merge, but I approved it so you can check the edit and merge.

PR #820, opened 2017-03-10
Service disruption guide

brittag:
For review by Atlas and the rest of the team, based on [this doc](https://docs.google.com/document/d/12U3dgnUv73XOSgHRgI2Ndx8zklQ4tFim_9Pktiza_-M/edit).

This puts our basic process into a guide. This overlaps with the earlier info at https://cloud.gov/docs/ops/making-announcements/ - once this page looks good to people, we can figure out what to do with that overlapping page.

We also need to make a process for keeping the list of people in this doc up to date.

### DIFF
--- a/content/docs/ops/service-disruption-guide.md
+++ b/content/docs/ops/service-disruption-guide.md
@@ -11,7 +11,7 @@ Not a guide for:
 
 * Hours outside of East Coast business hours -- out of scope for this document.
 * Security incidents -- switch to [security incident guide]({{< relref "docs/ops/security-ir.md" >}}).
-* Major service disruptions (more than 30 minutes of unexpected downtime or significantly reduced service) -- switch to [contingency plan]({{< relref "docs/ops/contingency-plan.md" >}}).
+* Major service disruptions (more than 7 days of unexpected downtime or significantly reduced service) -- switch to [contingency plan]({{< relref "docs/ops/contingency-plan.md" >}}).
 
 ## Threshold for "service disruption"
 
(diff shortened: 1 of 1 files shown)


content/docs/ops/service-disruption-guide.md
```diff
@@ -0,0 +1,58 @@
...
+title: Service disruption guide
+---
+
+This guide outlines how we handle communications for service disruptions during East Coast business hours.
+
+Not a guide for:
+
+* Hours outside of East Coast business hours -- out of scope for this document.
+* Security incidents -- switch to [security incident guide]({{< relref "docs/ops/security-ir.md" >}}).
+* Major service disruptions (more than 30 minutes of unexpected downtime or significantly reduced service) -- switch to [contingency plan]({{< relref "docs/ops/contingency-plan.md" >}}).
```
mogul, line 14: 30 minutes seems like a low threshold for going from disruption to contingency. We should probably discuss if a bigger number makes sense because I think some of our recent outages would have been over that threshold.

mogul: (That means updating it in the contingency plan too.)

brittag:
Noting for future reference as well: we'll need to add this page to https://github.com/18F/cg-product/blob/master/OnboardingChecklist.md

content/docs/ops/service-disruption-guide.md
brittag: I agree that we should take another look at the contingency plan. That should probably be its own story, since I'd want us to check for compliance effects - we may need to check with our FedRAMP team when making significant changes to it.

mogul (dismissed):
Sorry it's so late... @brittag are you still content for this to go in as-is (with my edit)?

brittag:
@mogul We have to stick with the current Contingency Plan timing for now (we'd have to propose a plan change and check it with the TRs before we can loosen the timing), so I'll put in a commit for that.

mogul (approved):
I'm still 😬 on the 30 minute threshold change, but I don't think it should hold up getting this published.

PR #818, opened 2017-03-09
Cut uaa-auth service from experimental services

brittag:
It's *too* experimental. Also update warning.

PR #823, opened 2017-03-13
Add "Experimental" to alpha service descriptions

adborden:
https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-3602

### DIFF
--- a/data/services/aws-rds.toml
+++ b/data/services/aws-rds.toml
@@ -1,3 +1,5 @@
+# Service metadata lives
+# https://github.com/18F/aws-broker/blob/59f949b/catalog-template.yml#L8
 name = "aws-rds"
 page_name = "relational-database"
 description = "Persistent, relational databases using Amazon RDS"
--- a/data/services/cdn-route.toml
+++ b/data/services/cdn-route.toml
@@ -1,3 +1,5 @@
+# Service metadata lives
+# https://github.com/18F/cf-cdn-service-broker/blob/d44bc1a/catalog.json#L4
 name = "cdn-route"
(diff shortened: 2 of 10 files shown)


data/services/aws-rds.toml
```diff
@@ -1,3 +1,5 @@
+# Service metadata lives
+# https://github.com/18F/aws-broker/blob/59f949b/catalog-template.yml#L8
```
brittag, line 2: Nice idea! Suggest using the branch URLs - such as https://github.com/18F/aws-broker/blob/master/catalog-template.yml#L8

adborden: I usually avoid branch URLs because then the line number can change or file can be moved and then the user is totally in the dark after clicking the link. Users can track a commit through it's history to find where the line moved to if they're looking for recent history.

brittag: Interesting point! I think in general if the file moves, we should update the links in these comments - my hope would be that the editor of this file would be able to hop over to the referenced file and edit the description to match.

brittag: Actually the best resolution for this might be to skip thinking about the "best" strategy, keep these URLs since they definitely serve as a helpful hint, and hope for the glorious future of https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-3375 :)

brittag:
Ah, let's do one more thing - delete the `uaa-auth.toml` file so that it no longer shows up in the services table, since we've discontinued offering that service for now (https://github.com/18F/cg-site/pull/818).

brittag (approved):

PR #824, opened 2017-03-14
Release notes 9 4

jameshupp:
## Changes
Adds new release notes

## Notes
With this release we are going to list only USNs rather than USNs and CVEs, unless there’s something specific in CVEs to call out. The lists can get excessively long, and they are referenced by USNs elsewhere. It’s probably worth reviewing our CVE capture to see if we’re able to hold onto the USNs appropriately, as well, at the end of each sprint.


brittag (approved):
Thank you James!

Issue #563, opened 2016-11-15
ELB documentation needs updating / deleting


cnelson:
We either need to update the [ELB Termination](https://cloud.gov/docs/ops/elb/) page with what we do in GovCloud or delete it.

**Acceptance Criteria**
- [ ] ELB documentation is updated or removed form the site

cnelson:
Does any of this information still apply? If not, can we just specifically mention ELBs in the 
the Terraform section in the [aws onboarding](https://cloud.gov/docs/ops/aws-onboarding/) and delete this page?

afeld:
Since the CDN broker is only available in GovCloud (https://cloud.gov/docs/apps/custom-domains/#managed-service-method), seems like it would be sufficient to add a disclaimer to the top of the page with "this only applies to accounts that can't use GovCloud". That being said, maybe we can create the service instances for any new systems that would need it but can't log into GovCloud, and thus we don't need these instructions anymore.

/cc @jmcarp

PR #826, opened 2017-03-15
Drop outdated ELB docs.

jmcarp:
At this point, the ELB docs do more harm than good, since almost statement in them is out of date, and since we stopped using ELBs for custom customer domains. I think we should just delete these pages.


brittag (approved):
This sounds good to me from a content perspective, but I won't merge yet since probably somebody else should review from a technical perspective.

cnelson:
Fixes https://github.com/18F/cg-site/issues/563

PR #834, opened 2017-03-17
Pricing restructure

jameshupp:
## Changes
- restructures rates around the packages rather than introducing all rate components independently before describing packages
- pushes most of the details about each rate component to a separate page about that component
- increases content on the terminology page
- reorders pricing pages

## Notes
This addresses a number of acceptance criteria but we still need to get purchase-related content into the top nav. We might follow that in this same branch, or we might bring it in a separate PR, depending on how timing moves forward.


jameshupp:
One issue to consider here: the H3 looks almost indistinguishable from bold content. That makes this particular approach to surfacing pricing info harder.

packages

jameshupp:
Another issue: I hacked this button into the markdown, but I would bet there's either a more professional way to do it, or there's a simple fix to the button that would avoid the underline tail you see here.
button_tail

jameshupp:
Update on what's in this PR: I've also changed the top nav label from "Overview" to "Product". It feels less repetitive. I'd like to address `overview.md` in the same way, but I can't find where the title of the page comes from. Would add that if it was discovered before this got merged, but otherwise will figure out and do separately.

nikzei (approved):

nikzei:
@jameshupp "Product" seems only to have replaced "overview" in the nav bar when on the rates page. Otherwise, the nav bar remains the same - I think the change needs to be made somewhere else to make it apply to the whole site . . .

PR #835, opened 2017-03-21
Overview whatis

jameshupp:
## Changes
* change top nav link to `what-is-cloudgov.md`
* incorporate `overview.md` content into what-is
* switch info about building to a team page, pull team info from other pages into this one
* reduce duplication by deleting other pages about building
* incorporate info about why we made this product available into the what-is page, and delete that page
* retitle “why use” and focus on cloud.gov’s _benefits_, since the other aspects of why are now covered on what-is
* reweight pages for ordering
* makes intro to capabilities a bit softer

## notes
It’s probably worth changing the URL tree at some point here from `/overview/overview/` to `/product/overview/`. But I think it’s better to get the rest of this in.


nikzei (approved):

mogul:
Just want to say thanks, this is a big improvement! 🥇

PR #836, opened 2017-03-21
Add snort alert docs.

jmcarp:


content/docs/ops/runbook/troubleshooting-snort.md
```diff
@@ -0,0 +1,29 @@
...
+        cat /var/vcap/sys/log/snort-eth0/snort.log.XXXXX | strings
+        ```
+
+    * Optionally use `u2spewfoo` to decode the snort binary format:
+        
+        ```
+        /var/vcap/packages/snort/bin/u2spewfoo /var/vcap/sys/log/snort-eth0/snort.log.XXXXX
+        ```
+
+* If the alert appears to be a false positive, consider excluding the rule from the snort configuration.
```
cnelson, line 28: Can we link to something that would tell someone how to exclude false positives from the rule set? I wouldn't know where to do that today as https://github.com/18F/cg-snort-boshrelease  / https://github.com/18F/cg-snort-boshrelease/blob/master/jobs/snort/spec is effectively undocumented.

We might also want to touch on what VMs an exclusion should cover...
Should it be done on just the job vms affected? All VMs in the deployment? All of cloud.gov?

jmcarp: I think excludes are always global for now: https://github.com/18F/cg-snort-boshrelease/commit/ac41ca0255073c82078ab3dd36abf3510b15ab1d. We could make them configurable per VM, but then we'd need to take snort out of the runtime config.

jmcarp: Updated with some lame text about how we currently handle exclusions.

cnelson (approved):

PR #841, opened 2017-03-24
Document space auditor plan.

jmcarp:


brittag (approved):
This looks good to me! Added a commit to make placeholder text really obvious, to encourage people to customize it instead of using the default service instance name. Approved from a content perspective; not sure if it needs anyone else to look at it.

mogul (approved):

PR #842, opened 2017-03-24
adding language specifically for auth

LindsayYoung:
Without an auth specific use case, it was a bit confusing to me, because I thought since I already had a team, I was supposed to invite people to the the team. It also makes clear they don't have to set up a sandbox.

Thanks!


mogul:
I'm thinking this might be out of place here, rather than on the auth service page, since people are just getting started and unlikely to know what this might refer to. @brittag what do you think? Worth calling out here or not? Ref [the Slack convo](https://gsa-tts.slack.com/archives/C09CR1Q9Z/p1490385927404687) with Lindsay for background.

mogul:
I merged #843 and figure that will allay confusion until at least the next time someone gets confused, then we'll iterate from there.

PR #847, opened 2017-03-29
Refer more clearly to sandbox limitations

jameshupp:
Changing site content here in a number of places to make clear that limitations are there, including upcoming change in policy

### DIFF
--- a/content/overview/pricing/free-limited-sandbox.md
+++ b/content/overview/pricing/free-limited-sandbox.md
@@ -8,20 +8,18 @@ aliases:
   - /overview/pricing/free-limited-sandbox
 ---
 
-Anyone with a U.S. federal government email address (.gov, .mil, or .fed.us) can evaluate cloud.gov for free using a sandbox space.
-
-A sandbox is a free, limited way to try cloud.gov and see if it might suit your team's needs. The sandbox gives you a sense of how cloud.gov would work for the applications your team builds. From the [setup process]({{< relref "docs/getting-started/setup.md" >}}) through [deploying an app]({{< relref "docs/getting-started/your-first-deploy.md" >}}), it works similarly to using a [paid access package]({{< relref "overview/pricing/rates.md" >}}), with [some limitations](#sandbox-limitations).
+A sandbox is a free cloud.gov account that you can use for a limited time to see if it might suit your team’s needs. You can build and deploy applications to get a sense of how cloud.gov would work for your team. From the [setup process]({{< relref "docs/getting-started/setup.md" >}}) through [deploying an app]({{< relref "docs/getting-started/your-first-deploy.md" >}}), it works similarly to using a [paid access package]({{< relref "overview/pricing/rates.md" >}}), with [some limitations](#sandbox-limitations).
 
 ## Get a sandbox
 
-Anyone with a U.S. federal government email address can [**sign up for a free sandbox space**](https://account.fr.cloud.gov/signup). No paperwork is required from us. (It's up to you to determine whether you may need permission from your organization.) If you have other questions or comments, see [Contact]({{< relref "docs/help.md" >}}).
+Anyone with a U.S. federal government email address (ending in `.gov`, `.mil`, or `.fed.us`) can [**sign up for a free sandbox space**](https://account.fr.cloud.gov/signup). No paperwork is required from us. (It’s up to you to determine whether you may need permission from your organization.) If you have other questions or comments, see [Contact]({{< relref "docs/help.md" >}}).
 
-Then, if you're interested in [purchasing full access]({{< relref "overview/pricing/rates.md" >}}) (whether for **Prototyping** or for production systems at the **Open Data**, **FISMA Low**, or **FISMA Moderate** levels), fill out [this interest form](https://docs.google.com/a/gsa.gov/forms/d/e/1FAIpQLSevZfuJ_4KE-MZlm9gttYfsXQp0PJL7OR6k6LbZ9XnFn-oA6g/viewform) or email us again.
+Then, if you’re interested in [purchasing full access]({{< relref "overview/pricing/rates.md" >}}) (whether for **Prototyping** or for production systems at the **Open Data**, **FISMA Low**, or **FISMA Moderate** levels), fill out [this interest form](https://docs.google.com/a/gsa.gov/forms/d/e/1FAIpQLSevZfuJ_4KE-MZlm9gttYfsXQp0PJL7OR6k6LbZ9XnFn-oA6g/viewform) or email us again.
 
-## To keep in mind before you try your sandbox
+## Keep in mind before you try your sandbox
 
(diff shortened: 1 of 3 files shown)


nikzei (approved):

content/overview/pricing/free-limited-sandbox.md
```diff
@@ -8,44 +8,42 @@ aliases:
   - /overview/pricing/free-limited-sandbox
 ---
 
-Anyone with a U.S. federal government email address (.gov, .mil, or .fed.us) can evaluate cloud.gov for free using a sandbox space.
-
-A sandbox is a free, limited way to try cloud.gov and see if it might suit your team's needs. The sandbox gives you a sense of how cloud.gov would work for the applications your team builds. From the [setup process]({{< relref "docs/getting-started/setup.md" >}}) through [deploying an app]({{< relref "docs/getting-started/your-first-deploy.md" >}}), it works similarly to using a [paid access package]({{< relref "overview/pricing/rates.md" >}}), with [some limitations](#sandbox-limitations).
+A sandbox is a free account that you can use to see if cloud.gov might suit your team’s needs. From the [setup process]({{< relref "docs/getting-started/setup.md" >}}) through [deploying an app]({{< relref "docs/getting-started/your-first-deploy.md" >}}), it works similarly to using a [paid access package]({{< relref "overview/pricing/rates.md" >}}), with [some limitations](#sandbox-limitations).
```
brittag, line 11: I would specify that a sandbox is a free space; accounts don't have costs in general, but orgs and spaces can have costs.

content/overview/pricing/free-limited-sandbox.md
```diff
@@ -8,44 +8,42 @@ aliases:
...
 
 ## Get a sandbox
 
-Anyone with a U.S. federal government email address can [**sign up for a free sandbox space**](https://account.fr.cloud.gov/signup). No paperwork is required from us. (It's up to you to determine whether you may need permission from your organization.) If you have other questions or comments, see [Contact]({{< relref "docs/help.md" >}}).
+Anyone with a U.S. federal government email address (ending in `.gov`, `.mil`, or `.fed.us`) can [**sign up for a free sandbox space**](https://account.fr.cloud.gov/signup). No paperwork is required from us. (It’s up to you to determine whether you may need permission from your organization.) If you have other questions or comments, see [Contact]({{< relref "docs/help.md" >}}).
```
brittag, line 15: Good to know that some .gov email addresses aren't federal government email addresses (they can be state email addresses for example), but that may be too specific for this context.

content/overview/pricing/free-limited-sandbox.md
```diff
@@ -8,44 +8,42 @@ aliases:
...
+* If your organization prohibits installing the [cloud.gov command line interface]({{< relref "docs/getting-started/setup.md#set-up-the-command-line" >}}) on your computer, you won’t be able to deploy applications in your sandbox. (For paid access packages, we can coordinate with your agency to help them approve this tool.)
 
 ## A few things you can try in your sandbox
 
-On a technical level, a sandbox is a specially-limited ["space"]({{< relref "docs/getting-started/concepts.md#spaces" >}}) within a sandbox-only ["organization"]({{< relref "docs/getting-started/concepts.md#organizations" >}}) that is managed by cloud.gov.
+On a technical level, a sandbox is a specially-limited ["space"]({{< relref "docs/getting-started/concepts.md#spaces" >}}) within a sandbox-only ["organization"]({{< relref "docs/getting-started/concepts.md#organizations" >}}) that is managed by cloud.gov. In paid accounts, you would manage your own organization and set up spaces for different applications. With a sandbox, you are managing one space and the applications and services within it.
```
brittag, line 26: An account can have both access to a free sandbox space and a paid org, but their access to a paid org might not be "Org Manager" access - they might just have "Space Developer" access to spaces there. Also often an org has spaces for same or similar versions of the same application (such as a development and a production space) rather than different applications.

content/overview/pricing/rates.md
```diff
@@ -70,7 +70,7 @@ Interested in a FISMA Moderate account? [Fill out our initial interest form.](ht
 
 ### Sandbox
 
-Free of charge and without any paperwork, anyone with a government email address can use a limited version of cloud.gov for a limited time. Before the time on any sandbox expires, we will reach out to alert you and discuss whether moving forward on another package makes sense for your system. You can’t use sandboxes for production.
+Free of charge and without any paperwork, anyone with a government email address can use a [limited version of cloud.gov]({{< relref "overview/pricing/free-limited-sandbox.md" >}}). The use of sandboxes is [limited]({{< relref "overview/pricing/free-limited-sandbox.md" >}}#sandbox-limitations), and the contents of all sandboxes are cleared every 30 days. But we will also reach out to alert you before any deletion point to remind you to save your data, and to discuss whether moving forward on another package makes sense for your system.
```
brittag, line 73: Saying "the contents of all sandboxes are cleared every 30 days" here may be confusing since we aren't yet making that actually happen.

content/overview/pricing/free-limited-sandbox.md
nikzei: I don't understand the change here: the fact that sandboxes are free is very clearly communicated already (for each package description on the Packages and Rates page, which leads to this sandbox page, the price is listed under the package description and we list sandboxes as FREE. On this page, free is in the header. Sandboxes are discussed as free in the first line of the previous text . . .) to me, if anything, I feel like we could discuss it *less* rather than more . . .

Is it that you want to communicate that there *could* be costs? If so, I don't get that so clearly in this revision. Perhaps that's what you might want to emphasize?

content/overview/pricing/rates.md
nikzei: That's an interesting point. Is there a clear plan around when we will start doing this? If so, perhaps the tweak here is just to state that we will begin to do this on [x date], and that the clock for all existing accounts will start on that date (or whatever you all decide). 

James has been revising communications to ensure that users understand sandboxes as ephemeral. If the communication needs to be different here, does it need to be different in all of his revisions elsewhere while we are still in this transitional phase?

content/overview/pricing/free-limited-sandbox.md
nikzei: Ok, I think my initial emphasis was on *free* when it should have been on *space.* I think I understand better now. It's a great point that might also apply for access package customers? My two cents:

I'd assume that many prospective sandbox users will have entry level knowledge so a really clear and direct statement here would probably be appropriate ("as with all access packages, if you want [x], you'll have to purchase that separately."). Laying this out a little bit more clearly on the page (without relying on hyperlink) could also help us educate prospective customers about the way that pricing works and will work if they purchase an access package in the future.

brittag: Ah yeah, I meant my comment with emphasis on _space_ rather than _free_. I agree that "free" is already well-communicated in general and could even be repeated less. The problem is that calling a sandbox a free _account_ is confusing. The limitations on sandbox spaces apply whether or not your account also has access to paid orgs.

Specifically I mean that the change from "A sandbox is a free, limited way" to "A sandbox is a free account" adds some confusion because a sandbox is not an "account", it's a "space".

content/overview/pricing/rates.md
brittag: There's no specific date for that switchover yet - it's waiting on users being properly informed that the change will happen. We just shouldn't tell them it's currently-happening until it's actually currently-happening.

brittag: (It may make sense for James + Atlas to work together on determining a target date for the switchover.)

content/overview/pricing/free-limited-sandbox.md
jameshupp: I'm leaving this one as-is, since we lead into the parenthetical with `federal`, unless we've actually experienced state emails as a problem.

PR #848, opened 2017-03-29
stating that sandboxes get cleared every 30 days

nikzei:
Added this in the sandbox description paragraph so that it is clear and up front.


jameshupp:
I just put an explicit reference to clearing every 30 days into #847. @nikzei if you can review that and it looks good to merge, we can close this one rather than merging. (Merging both is likely to create merge conflicts, I think.)

jameshupp:
Handled by #847

PR #854, opened 2017-04-03
Rates MB pricing

jameshupp:
This attempts to surface the daily MB rate via the GB rate, since listing the daily MBs on the packages seemed extra confusing. I’m not wild about this but would love to get feedback. @brittag

### DIFF
--- a/content/overview/pricing/rates.md
+++ b/content/overview/pricing/rates.md
@@ -33,7 +33,7 @@ Once you fill out the [initial interest form](https://docs.google.com/forms/d/e/
 Our most basic paid package lets many teams build and deploy apps, with access control delegated to teams. Prototyping accounts cannot host production data; all apps are limited to the `*.app.cloud.gov` domain. Whole agencies or departments often purchase a single prototyping package.
 
 **Annual fee:** $15K<br>
-**Resource usage quota:** $99 per GB per month, up to any limit set by user<br>
+**Resource usage quota:** ~$99 per GB per month[*](#quota-pricing), up to any limit set by user<br>
 **Managed services:** Full access
 
 Interested in a prototyping account? [Fill out our initial interest form.](https://docs.google.com/forms/d/e/1FAIpQLSevZfuJ_4KE-MZlm9gttYfsXQp0PJL7OR6k6LbZ9XnFn-oA6g/viewform)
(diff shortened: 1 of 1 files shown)


content/overview/pricing/rates.md
```diff
@@ -78,6 +78,16 @@ Free of charge and without any paperwork, anyone with a government email address
 
 [Learn more about sandboxes.]({{< relref "overview/pricing/free-limited-sandbox.md" >}})
 
+## Quota pricing
+
+Quotas are priced at $0.0033 per MB reserved per day and assessed monthly. In practice, this works out to $99 per GB per month, based on a 30-day month:
+
+`$0.0033 * 1000 MB * 30 days = $99 per GB per month`
+
+The monthly cost will vary slightly from $92.40 per GB in a 28-day month to $102.30 per GB in a 31-day month.
```
khandelwal, line 87: Do our users really care about this level of precision? I can't imagine we need to specify the cents, but additionally the difference is really only $10. That's like not going to make or break an agency's budget.

content/overview/pricing/rates.md
```diff
@@ -78,6 +78,16 @@ Free of charge and without any paperwork, anyone with a government email address
 
 [Learn more about sandboxes.]({{< relref "overview/pricing/free-limited-sandbox.md" >}})
 
+## Quota pricing
+
+Quotas are priced at $0.0033 per MB reserved per day and assessed monthly. In practice, this works out to $99 per GB per month, based on a 30-day month:
```
khandelwal, line 83: What does 'reserved per day' mean here?

content/overview/pricing/rates.md
jameshupp: The issue we're really parsing here is whether we can say "$99 per GB per month" vs. "$0.0033 per MB per day". The former is easier to comprehend, but the latter is more accurate. What I want to do is lead with the former but provide the latter so people understand the technicality and/or don't accidentally misrepresent the price.

So the question is: is it easier to just provide the $0.0033 figure? Or do we accept the risk of uncertainty from just providing the $99 figure? That's where we started out here, but in conversation, @brittag I were discussing that it isn't really correct so it might be wrong to say it that way. 

It's also technically more like $100.38 per GB per month. $99 is a great estimate for understanding how it looks over the course of 30 days, but there are still 5 extra days in the year after that. You consider this stuff long enough and leaving it at $99 starts to feel insufficient. But again, maybe we're way out in front of our customers.

content/overview/pricing/rates.md
jameshupp: The quota is a reservation, not what you actually use. What's unused isn't refunded. So if you reserve, say, 5GB, your price is $16.50 per day. Repeated over a 30-day month, that's $495 per month. You pay the quota price for every single day, and then we assess and bill monthly.

Maybe it'd be better to say something like:
`Quotas are set in MB of memory you reserve. The price is $0.0033 per MB, and the price is considered per day. We assess the usage monthly, which in practice works out to about $99 per GB for a 30-day month.`

Then maybe pull all the calculation stuff over to the dedicated page for this (`quotas.md`)?

content/overview/pricing/rates.md
brittag: Yes, we should have this detail available so that everyone can make accurate estimates and check their bills. $10 isn't meaningful for an IT budget, but it can be meaningful for making everyone's bookkeeping line up precisely, and that can be important as well.

brittag (approved):

PR #859, opened 2017-04-13
Try clarifying ssh environment setup

adborden:
Debugging an issue with @laurenancona and neither of us picked up on the fact that after logging into an app with `cf ssh` you still need to run some setup commands in order to run your app. This is my weak attempt at trying to make that more clear, short of copy/pasting commands from the Cloud Foundry docs.

Feel free to improve :)


brittag:
@adborden Just to verify, the only addition here is "in order to debug your application"? It sounds like we should make a stronger statement here in general - something like replacing "You'll want to" with "You need to"!

mogul (approved):
Made Britta's suggested change

PR #860, opened 2017-04-14
Tell people about seeing guid in space

LindsayYoung:
Tell people there will be extra accounts in the space that is deploys from. That way they don't have to confirm that is what it is every time they do a space audit.

Thanks!


brittag (approved):
Thank you @LindsayYoung, this is helpful!

PR #863, opened 2017-04-18
Add screenshots of ELK 5.x dashboards [Don't merge until new version is in production]

jmcarp:
Because they changed the colors.


brittag:
Added a commit to make the page use the new images and update the instructions a little bit. Looks fine, but I'll let somebody else approve.

jmcarp:
Looks like I forgot to check in those changes--thanks for fixing, @brittag 😓.

cnelson (approved):
LGTM, but we should wait to merge this until the new version is in production.

PR #866, opened 2017-04-20
Add policy for resetting second factor for cloud.gov IDP

brittag:
This is based on [the discussion here](https://docs.google.com/document/d/1LvPb33vmZGA7zWEpFevWhBv_76nJCXMXe4jqFFX-TsY/edit). Requesting review from both @rogeruiz as Cloud Ops representative and @khandelwal as System Owner; we shouldn't merge this without approval from @khandelwal.

Changes:
* Consolidate existing instructions about MFA into one paragraph.
* Add paragraph about what happens if you lose your MFA token.

### DIFF
--- a/content/docs/getting-started/accounts.md
+++ b/content/docs/getting-started/accounts.md
@@ -27,11 +27,10 @@ If you were invited with an email address that isn't part of an agency with sing
 
 Follow [these instructions to log in on the command line (CLI) and web UI (dashboard)]({{< relref "setup.md" >}}). When you log in to cloud.gov via a web browser, select the **cloud.gov** option.
 
-<!--<img src="/img/cloud-gov-idp-button-cropped.png" alt="cloud.gov provider button to login page is highlighted in blue" width="373" height="187" />-->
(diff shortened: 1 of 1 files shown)


rogeruiz:
> to recap from the [Google Doc](https://docs.google.com/document/d/1LvPb33vmZGA7zWEpFevWhBv_76nJCXMXe4jqFFX-TsY/edit?ts=58f80cc4) 🔒 , I _think_ we should mention the risk associated with users that have access to several `sandbox-*` spaces, all users should know that any apps that another user requesting their MFA token reset that had `SpaceDeveloper` roles in their space will **also** get apps deleted in their `sandbox--*` spaces.

The idea is that an attacker who compromised a user's email address would be prevented from gaining access to the contents of what is in an another user's apps. We don't want to take the risk of compromising someone's codebase just because the attacker's user previously had access to it, considering the possibility of any sensitive environment information that we do/don't provide. If we just remove the access, then they can possibly just get access via that compromised email and email the `SpaceDeveloper` in that space. Then the risk would fall on our users. We should make that point clear.

Also, @brittag brought up a good point about a compromised email address getting access on to our system. But I feel like that's a different scenario with the same possible outcome, that's more of a targeted attack. Still a concern, but from our perspective we probably wouldn't want to be traced back as the people responsible to reset the compromised user's access in the recent past.

It would also reinforce the idea that _sandboxes are ephemeral_ ☁️.

🤔 thoughts @18F/cloud-gov-ops  ??

brittag:
Thank you @rogeruiz! We talked about this internally a bit more and will take a different direction; closing this PR.

PR #868, opened 2017-04-20
Second iteration of policy for resetting second factor for cloud.gov IDP

brittag:
This is a followup to [this earlier PR about MFA reset policies](https://github.com/18F/cg-site/pull/866), providing a simpler procedure.

This is based on [the additional discussion here](https://docs.google.com/document/d/1LvPb33vmZGA7zWEpFevWhBv_76nJCXMXe4jqFFX-TsY/edit#heading=h.afmvcdkdd868).

Similarly, requesting review from both @rogeruiz as Cloud Ops representative and @khandelwal as System Owner; we shouldn't merge this without approval from @khandelwal.

### DIFF
--- a/content/docs/getting-started/accounts.md
+++ b/content/docs/getting-started/accounts.md
@@ -29,9 +29,9 @@ Follow [these instructions to log in on the command line (CLI) and web UI (dashb
 
 <!--<img src="/img/cloud-gov-idp-button-cropped.png" alt="cloud.gov provider button to login page is highlighted in blue" width="373" height="187" />-->
 
-Your cloud.gov account requires multi-factor authentication. To log into the system, you need two "factors" -- something you know (your password) and something you have on your person (your smartphone).
+Your cloud.gov account requires multi-factor authentication: something you know (your password) and something you have on your person (your smartphone). In order to perform multi-factor authentication, you need an authentication application that generates time-based one-time passcodes, such as [Google Authenticator](https://support.google.com/accounts/answer/1066447?hl=en) or [Authy](https://www.authy.com/app/mobile). When you log into cloud.gov for the first time, follow the instructions to set this up.
 
-In order to perform multi-factor authentication, you need an authentication application that generates time-based one-time passcodes, such as [Google Authenticator](https://support.google.com/accounts/answer/1066447?hl=en) or [Authy](https://www.authy.com/app/mobile). Download the app on your mobile device. When you log into cloud.gov for the first time, follow the instructions to store the `cloud.gov` key in your application.
+
 {{% /govcloud %}}
 
 {{% eastwest %}}
--- a/content/docs/getting-started/accounts.md
+++ b/content/docs/getting-started/accounts.md
@@ -31,6 +31,13 @@ Follow [these instructions to log in on the command line (CLI) and web UI (dashb
 
 Your cloud.gov account requires multi-factor authentication: something you know (your password) and something you have on your person (your smartphone). In order to perform multi-factor authentication, you need an authentication application that generates time-based one-time passcodes, such as [Google Authenticator](https://support.google.com/accounts/answer/1066447?hl=en) or [Authy](https://www.authy.com/app/mobile). When you log into cloud.gov for the first time, follow the instructions to set this up.
 
+If you lose access to your authentication application (such as if you lose your phone), email [cloud.gov support](/help/) so that we can allow you to set up a new one. We'll follow this process to mitigate the risk of requests from compromised email addresses:
+
+1. Delete the contents of your sandbox space (if you have one).
+2. Remove your permissions to any other spaces and orgs.
+3. For those spaces and orgs, notify the Space Managers and Org Managers that we've removed your access because of your request to reset your second factor.
+4. Reset your second factor.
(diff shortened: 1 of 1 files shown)


content/docs/getting-started/accounts.md
```diff
@@ -29,9 +29,16 @@ Follow [these instructions to log in on the command line (CLI) and web UI (dashb
 
 <!--<img src="/img/cloud-gov-idp-button-cropped.png" alt="cloud.gov provider button to login page is highlighted in blue" width="373" height="187" />-->
 
-Your cloud.gov account requires multi-factor authentication. To log into the system, you need two "factors" -- something you know (your password) and something you have on your person (your smartphone).
+Your cloud.gov account requires multi-factor authentication: something you know (your password) and something you have on your person (your smartphone). In order to perform multi-factor authentication, you need an authentication application that generates time-based one-time passcodes, such as [Google Authenticator](https://support.google.com/accounts/answer/1066447?hl=en) or [Authy](https://www.authy.com/app/mobile). When you log into cloud.gov for the first time, follow the instructions to set this up.
```
khandelwal: This first sentence might need a little clarity. Are we defining multi-factor authentication, or are we defining what MFA means in the cloud.gov context?

Most of our audience will be familiar with MFA (and increasingly so) as time goes on.

khandelwal: "such as Google Authenticator ..." doesn't make it crystal clear that those are the two choices. I'm assuming users aren't free to choose which MFA application they want.

content/docs/getting-started/accounts.md
```diff
@@ -29,9 +29,16 @@ Follow [these instructions to log in on the command line (CLI) and web UI (dashb
...
+
+If you lose access to your authentication application (such as if you lose your phone), email [cloud.gov support](/help/) so that we can allow you to set up a new one. We'll follow this process to mitigate the risk of requests from compromised email addresses:
```
khandelwal: Putting a title of sorts above this section might help readability. Something like "What happens if I lose my token?"

content/docs/getting-started/accounts.md
```diff
@@ -29,9 +29,16 @@ Follow [these instructions to log in on the command line (CLI) and web UI (dashb
...
+
+1. Delete the contents of your sandbox space (if you have one).
+2. Remove your permissions to any other spaces and orgs.
+3. For those spaces and orgs, notify the Space Managers and Org Managers that we've removed your access because of your request to reset your second factor.
```
khandelwal: I think this is the first time we mention the phrase 'second factor'. Do we need to clarify what this means?

khandelwal (dismissed):

rogeruiz (commented):
It looks good from a technical perspective, based on what we covered yesterday after standup. I'll defer to @khandelwal's review for the final approval.

content/docs/getting-started/accounts.md
rogeruiz: We had this explanation in our documentation before. The above defines MFA as what it means in both contexts. Though it might make sense to consider leveraging the same words we use on the login screens from 18F/Shibboleth-IdP3-TOTP-Auth#15. e.g. `MFA app` and `token code`.

rogeruiz: It might make sense to stretch this out to include any [Time-based One-time Password Algorithm](https://en.wikipedia.org/wiki/Time-based_One-time_Password_Algorithm) (e.g.`TOTP`) application.  We had this question come in through support because we recommend `smartphone apps` and they wanted to use `a Google Chrome extension`. They're free to use any TOTP application and there are several Desktop applications as well. [Authy has most platforms covered](https://authy.com/features/multiple-devices/). 

Maybe we can just call it _something you know (your password)_ and _something you have (an MFA/2FA application)_? or do we need to keep the phone part as part of the SSP?

content/docs/getting-started/accounts.md
rogeruiz: Meh, I'd say it's similar enough to  multi-factor / second-factor. Maybe when we explain MFA we can say `first factor ( password )` and `second factor ( MFA/2FA app )` ?? 🤷‍♂️

brittag:
@rogeruiz I added a new commit with language adjustments to address Shashank's comments - up for taking another look? If you think it's ok, we can merge it.

rogeruiz (approved):
looks great! :shipit:

content/docs/getting-started/accounts.md
```diff
@@ -38,7 +38,7 @@ If you lose access to your authentication application (such as if you lose your
 1. Delete the contents of your sandbox space (if you have one).
 2. Remove your permissions to any other spaces and orgs.
 3. For those spaces and orgs, notify the Space Managers and Org Managers that we've removed your access because of your request to reset your account's authentication application.
-4. Reset your second factor.
+4. Reset your account's authentication application.
```
rogeruiz: Good catch. Are we waiting on @khandelwal approval? or should I dismiss his previous one since it looks like you've addressed everything. wdyt? @brittag

brittag: @rogeruiz Go ahead and dismiss them - I approve this in his place! We don't have to wait for him to return from OOO; I have confidence we have addressed his suggestions.

PR #869, opened 2017-04-21
Add section on CDN caching and authentication.

cmc333333:
After being bit by an incorrect mental model of the CDN's caching behavior, I
figured we should document it. At roughly the same time, someone asked about
authentication, so I added that as well.

Maintainers should be able to edit this PR directly -- feel free to rewrite everything :)

### DIFF
--- a/content/docs/services/cdn-route.md
+++ b/content/docs/services/cdn-route.md
@@ -123,6 +123,27 @@ field. If you do, follow ["How to set up DNS"](#how-to-set-up-dns) again.
 
 After you delete an existing service instance, you may need to wait up to 30 minutes for the changes to complete and propagate.
 
+## Caching
+
+CloudFront [uses](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Expiration.html)
+your application's `Cache-Control` or `Expires` HTTP headers to determine how
+long to cache content. If your application does not provide these headers,
+CloudFront will use a default timeout of **24 hours**. This can be
+particularly confusing as different requests might be routed to different
+CloudFront Edge endpoints.
+
+While there is no mechanism for cloud.gov users to trigger a cache clear, the
+cloud.gov support team can. Note however, that cache invalidation is not
+instantaneous; Amazon expecting a lag time of 10-15 minutes (more if there are
+many distinct endpoints.)
+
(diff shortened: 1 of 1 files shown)


brittag:
Thank you so much for this addition - yes, I think this will be really helpful and prevent other people from running into problems!

The one simple change I'd make is moving these new "Caching" and "Authentication" sections together underneath something like a "More about how this service works" section, and then sticking that section after the "External services and applications" section - so that you get the three sets of basic creation/update instructions first, and then additional info about more advanced aspects. I'll wait for a double check on the technical info by Atlas before mucking around with content though. :)

content/docs/services/cdn-route.md
```diff
@@ -123,6 +123,27 @@ field. If you do, follow ["How to set up DNS"](#how-to-set-up-dns) again.
...
+CloudFront [uses](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/Expiration.html)
+your application's `Cache-Control` or `Expires` HTTP headers to determine how
+long to cache content. If your application does not provide these headers,
+CloudFront will use a default timeout of **24 hours**. This can be
+particularly confusing as different requests might be routed to different
+CloudFront Edge endpoints.
+
+While there is no mechanism for cloud.gov users to trigger a cache clear, the
+cloud.gov support team can. Note however, that cache invalidation is not
+instantaneous; Amazon expecting a lag time of 10-15 minutes (more if there are
```
fureigh: @cmc333333 Should `Amazon expecting` simply be `expect`?

cmc333333: Sounds good. Or "Amazon recommends expecting".

cmc333333:
Hey @brittag and @fureigh, would you mind making these changes? I'm going to be bogged down in other work for a bit.

brittag:
Thanks @cmc333333, we'll take this forward!

mogul:
@jmcarp can you take a look at this one?

jmcarp (dismissed):

jmcarp:
Technical content looks good to me.

brittag:
@jmcarp Now ready from a content perspective! Feel free to approve again and merge.

jmcarp (approved):

PR #872, opened 2017-04-25
adding spreadsheet

jameshupp:
## Changes
- adds both xlsx format and ods format. xlsx is far more likely to be useful to our prospective customers, but I think it’s good to also provide an open standard if we can.
- links to both on rates page, including date of last update.

## Notes
- Because these are static files, we’ll need to replace each time. At some point it would be awesome to figure out a web-based way to do this. IP sprint thought!
- Note that last-updated date is both on the page and the filename. One more assurance for people about accuracy/recency of pricing info. But that will also need to be changed with any updates to the documents themselves.


jameshupp:
Sorry, created a pretty long review list, basically everyone who had worked on this plus @austinhernandez for visibility as he gets ramped up.

jameshupp:
One question for the group: what's the right way to note where the Google Sheet we're working from is located? Doesn't matter to users, so I think it wouldn't help much to list it publicly, but it's very useful for our own team going forward.

rememberlenny (commented):
This looks good!

We might want to link to an S3 upload, so we can change out the file without deployments.

We can make a assets folder for this time of thing.

brittag:
I'd suggest putting the URL of the Google Spreadsheet in an HTML comment in the .md page: `<!-- link -->`

That way it's easy to find in context for other team members who are editing the page. It'd be visible to readers if they go look at the source of the rendered page, which seems fine.

brittag:
@jameshupp This exported spreadsheet renders a bit oddly when I open it up in Excel:

screen shot 2017-04-26 at 9 02 43 am

suprenant (approved):

PR #879, opened 2017-05-03
Secrets rotations Runbooks

rogeruiz:
Pairing with @wjwoodson ~~and @commit-dkp~~ on the ~~initial~~ work done here. @18F/cloud-gov-ops, please comment or add to this PR as needed.

### DIFF
--- a/content/docs/ops/runbook/rotating-bosh.md
+++ b/content/docs/ops/runbook/rotating-bosh.md
@@ -1,0 +1,18 @@
+---
+menu:
+  docs:
+    parent: runbook
+
+title: Rotating Bosh
+---
+
+## As part of secrets rotation
+
+## Deploying updated secrets
+
+You can rotate everything here. You'll be [rotating IAM users]({{< relref "ops/runbook/rotating-iam-users.md" >}})
+as well.
+
+## Monitoring cloud check VMs
+
+Deployments won't work
--- a/content/docs/ops/runbook/rotating-bosh.md
+++ b/content/docs/ops/runbook/rotating-bosh.md
@@ -57,7 +57,16 @@ ID.
 bosh-cli tasks -a -r | head
 ```
 
-## Coordinate & monitor deployments
+Once you have a list of task IDs, you can launch into it by attaching to the
+task.
+
+```sh
+bosh-cli task ${task_id}
+```
+
+You can detach from the task without canceling it with `Ctrl-C`.
+
+### Coordinate & monitor deployments
 
 There are two types of deployments that need to be coordinated here. All of the
 BOSH directors deploy using the previous BOSH director in the pipeline with the
@@ -66,35 +75,117 @@ exception of the environment BOSH deployments (e.g. `development`, `staging`,
 `deploy-bosh` pipeline which use the BOSH credentials in either their secrets or
 credential files.
 
-### BOSH deployments
+#### External deployments
 
-#### Deploying Master BOSH
+There are various pipelines that require BOSH credentials which will immediately
+be out of date once the BOSH deployment it targets and cause deployment failures
+and potentially cause downtime. All pipelines depending on BOSH must be paused
+and updated before unpausing them again.
+
+##### Finding deployments with BOSH dependencies
+
+Search GitHub for the `BOSH_TARGET` and `BOSH_ENVIRONMENT` with a `YAML`
+filetype to find all the different deployments that will need to be paused and
+updated after all the Bosh deployments are
+
+```bash
+open \
+  'https://github.com/search?l=YAML&q=user%3A18F+BOSH_TARGET&type=Code&utf8=%E2%9C%93' \
+  'https://github.com/search?l=YAML&q=user%3A18F+BOSH_ENVIRONMENT&type=Code&utf8=%E2%9C%93' \
+  'https://github.com/search?l=YAML&q=user%3A18F+upload-release&type=Code' \
+  'https://github.com/search?l=YAML&q=user%3A18F+bosh-errand&type=Code'
+```
+
+Once these pipelines are paused, update the values after all BOSH deployments
+have successfully deployed. Then unpause pipelines after you have successfully
+uploaded secrets or reflown pipelines.
+
+#### BOSH deployments
+
+There are five BOSH deployments which are used to deploy cloud.gov. Master BOSH
+deploys Tooling BOSH which deploys Development, Staging, and Production BOSH.
+During a deployment of the BOSH director, BOSH will be unresponsive and will
+lead to errors in [any deployments that interact with the director](#external-deployments).
+Verify that these external deployments have been paused before proceeding with
+these steps.
+
+Besides certificates, all secrets in each BOSH deployment can be generated and
+replaced or can be updated from [rotated IAM users]({{< relref "docs/ops/runbook/rotating-iam-users.md" >}}).
+
+##### Validating BOSH deployments
+
+Each BOSH deployment contains a set a releases that is uploaded to it from the
+`deploy-bosh` pipeline. After a successful deployment of BOSH in a given
+environment, run the `common-releases-${bosh_environment}` job in Concourse.
+
+```sh
+fly --target fr \
+    trigger-job \
+    --job deploy-bosh/common-releases-${bosh_environment} \
+    --watch
+```
+
+Run this step after each successful BOSH rotation deployment. If it completes
+without errors, BOSH has been successfully rotated and deployed.
+
+##### Deploying Master BOSH
 
 Start with the Master bosh deployment and secrets. This will require that you
 create a new root CA certificate for this deployment and to sign certificates
 for subsequent BOSH deployments (e.g. Tooling, Development, Staging,
 Production).
 
-#### Deploying Tooling BOSH
+In the [`cg-deploy-bosh`](https://github.com/18F/cg-deploy-bosh) repository, you
+will find a `generate-master-bosh-certs.sh`. Create a `./tmp` directory
+and run that script in there with the IP address for Master BOSH. The IP address
+can be found in the secrets file you're rotating as either `${BOSH_TARGET}` or
+`${BOSH_ENVIRONMENT}`.
 
-After Master BOSH has been deployed and validated, it's time to update Tooling
-BOSH.
+```sh
+./generate-master-bosh-certs.sh ${master_bosh_ip_address}
+```
 
-#### Deploying other BOSH environments
+Once the certificates have been generated, update the values in the Master BOSH
+secrets. See the [secret key management]({{< relref "docs/ops/secrets.md" >}})
+and substitute the file you're downloading for the Master BOSH file found in the
+`deploy-bosh` pipeline.
 
-This sections covers deploy development, staging, and production BOSH. These
-deployments are deployed with the Tooling BOSH director. These deployments can
-happen serially or
+###### Encrypting private key
 
-### External deployments
+Use the same secrets passphrase you generated for the encrypting both the
+secrets file and the private key. Encrypt and upload the private key using the
+same technique for [secret key management]({{< relref "docs/ops/secrets.md" >}}).
 
-Enumerate dependencies in CF pipelines. These pipelines should be paused before rotating BOSH secrets,
-as they will fail until corresponding values have been rotated in each of their secrets files.
+```sh
+INPUT_FILE=master-bosh.key \
+OUTPUT_FILE=master-bosh.pem \
+PASSPHRASE=${master_bosh_secrets_passphrase} \
+$CG_SCRIPT/encrypt.sh
+```
 
-#### Finding deployments with BOSH dependencies
+##### Deploying Tooling BOSH
 
-```bash
-open \
-  'https://github.com/search?l=YAML&q=user%3A18F+BOSH_TARGET&type=Code&utf8=%E2%9C%93' \
-  'https://github.com/search?l=YAML&q=user%3A18F+BOSH_ENVIRONMENT&type=Code&utf8=%E2%9C%93'
+In the [`cg-deploy-bosh`](https://github.com/18F/cg-deploy-bosh) repository, you
+will find a `generate-bosh-certs.sh`. Create a `./tmp` directory and run that
+script in there with the IP address for Tooling BOSH. The IP address can be
+found in the secrets file you're rotating as either `${BOSH_TARGET}` or
+`${BOSH_ENVIRONMENT}`.
+
+```sh
+./generate-bosh-certs.sh Tooling ${tooling_bosh_ip_address}
+```
+
+##### Deploying other BOSH environments
+
+These deployments are deployed with the Tooling BOSH director.  In the
+[`cg-deploy-bosh`](https://github.com/18F/cg-deploy-bosh) repository, you will
+find a `generate-bosh-certs.sh`. Create a `./tmp` directory and run that script
+in there with the IP address for each BOSH (e.g. Development, Staging,
+Production). The IP address can be found in the secrets file you're rotating as
+either `${BOSH_TARGET}` or `${BOSH_ENVIRONMENT}`.
+
+```sh
+./generate-bosh-certs.sh Development ${development_bosh_ip_address}
+./generate-bosh-certs.sh Staging ${staging_bosh_ip_address}
+./generate-bosh-certs.sh Production ${production_bosh_ip_address}
 ```
--- a/content/docs/ops/runbook/rotating-bosh.md
+++ b/content/docs/ops/runbook/rotating-bosh.md
@@ -18,7 +18,7 @@ repository along with the rest of this guide.
 ### Agent limitations when updating BOSH
 
 The following should be referenced throughout the various deployments. [According
-to the BOSH development team](https://cloudfoundry.slack.com/archives/C02HPPYQ2/p1492784585438034),
+to the BOSH development team](https://github.com/cloudfoundry/bosh/issues/1140),
 there is currently no way to update the BOSH password without recreating the
 virtual machines (VMs). This is easy enough to do with the BOSH CLI. It is a
 time-consuming process which recreates all of the VMs for a particular
--- a/content/docs/ops/runbook/rotating-bosh.md
+++ b/content/docs/ops/runbook/rotating-bosh.md
@@ -6,7 +6,7 @@ menu:
 title: Rotating Secrets - Bosh
 ---
 
-### Introduction
+## Introduction
 
 Rotating Bosh secrets are pretty straight forward and should be done first.
 There are known limitations of the Bosh agent when updating the main Bosh
@@ -14,11 +14,15 @@ password for each deployment. Deployments of Bosh should be done serially due to
 the dependencies of where these Bosh directors are being deployed to. Please
 reference the
 
-
-Enumerate dependencies in CF pipelines. These pipelines should be paused before rotating BOSH secrets, 
+Enumerate dependencies in CF pipelines. These pipelines should be paused before rotating BOSH secrets,
 as they will fail until corresponding values have been rotated in each of their secrets files.
 ```bash
 https://github.com/search?l=YAML&q=user%3A18F+BOSH_TARGET&type=Code&utf8=%E2%9C%93
 https://github.com/search?l=YAML&q=user%3A18F+BOSH_ENVIRONMENT&type=Code&utf8=%E2%9C%93
 ```
 
+## Update Bosh secrets
+
+### Follow along with the Bosh documentation
+
+## Coordinate & monitor external deployments
--- a/content/docs/ops/runbook/rotating-bosh.md
+++ b/content/docs/ops/runbook/rotating-bosh.md
@@ -84,9 +84,11 @@ and updated before unpausing them again.
 
 #### Finding deployments with BOSH dependencies
 
-Search GitHub for the `BOSH_TARGET` and `BOSH_ENVIRONMENT` with a `YAML`
-filetype to find all the different deployments that will need to be paused and
-updated after all the Bosh deployments are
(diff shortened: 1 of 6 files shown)


content/docs/ops/runbook/rotating-secrets.md
```diff
@@ -0,0 +1,44 @@
...
+title: Rotating Secrets
+---
+
+# Start by rotating access to various services and IAM accounts.
+
+- Access AWS
+- Rotate these specific IAM Users first
+    - `bosh_user`
+    - `concourse_user`
+    - `cf_user`
```
jmcarp: We dropped `cf_user` in favor of instance profiles recently, and for some reason we've never actually provisioned the read-only admin user, so there aren't currently any credentials rotate there. Maybe it's time to either drop that user or add it to a stack--wdyt @LinuxBozo ?

rogeruiz: yeah good call on both of these. I wonder if the runbook is the best place for the actual names? or if it's just better to say something like. **Look through `cg-provision`, and in AWS, and make sure any IAM users with Policies attached to them that can modify contents of buckets or the infrastructure need to be rotated first?

jmcarp: By the end of the quarter, we're supposed to get rid of all IAM users, so this shouldn't matter for long. But for now, 👍 to "rotate all the users" or similar.

rogeruiz:
Looks like this is the last part left for secrets rotation. Should have this in a state to review today.

rogeruiz:
Moving this back to ready. I'm adding notes to it but not really working on it full-force until all secrets are finished rotating.

content/docs/ops/runbook/rotating-cloudfoundry-diego.md
```diff
@@ -0,0 +1,202 @@
...
+External secrets fall into a few categories which range from _rotatable_ with
+coordination with other cloud.gov deployments on one end and _non-rotatable_
+without data loss.
+
+#### Properties used in other deployments
+
+1. `uaa.clients.*.secrets` properties need to be coordinated across various Bosh
+   deployments and cloud.gov apps or service brokers.
+1. `collectd.riemann_server` property for Riemann server address.
+1. `nessus-agent.*` properties for Nessus.
```
cnelson: This should be removed, riemann/collectd/nessus config is handled by runtime-config now

content/docs/ops/runbook/rotating-cloudfoundry-diego.md
```diff
@@ -0,0 +1,202 @@
...
+1. `login.smtp.*` properties for Mandrill.
+
+#### Metadata and properties for external SAML providers
+
+1. `login.providers.*` metadata along with `login.saml.*` properties for
+   certificates and keys.
+    - The metadata comes from our external SAML providers. In order to rotate
+      these coordinate with them to exchange new metadata. The properties for
+      `login.saml.*` can be generate with the following documentation, although
```
cnelson: login.saml cannot currently be rotated without breaking authentication with properly-configured IDPs.

We are waiting on https://www.pivotaltracker.com/n/projects/997278/stories/142082611 to complete to allow rotation without downtime.  Once that stories done, rotating this will be a multi-step process similar to the consul certs.

content/docs/ops/runbook/rotating-cloudfoundry-diego.md
```diff
@@ -0,0 +1,202 @@
...
+      this data will also need to be coordinated and exchanged with all
+      providers.
+
+#### Credentials for RDS instances
```
cnelson: Shouldn't this just get merged in from terraform and not be copy/pasted into secrets?

content/docs/ops/runbook/rotating-secrets.md
```diff
@@ -0,0 +1,187 @@
...
+1. Check the services, if one exists for this deployer renamed it with a `_prev`
+   suffix for later deletion.
+   - `cf rename-service ${name_of_deployer} ${name_of_deployer}_prev`
+1. Create a new service instance.
+  - `cf create-service cloud-gov-service-account space-deployer "${name_of_deployer}"`
+1. Read the service instance details for the deployer.
+  - `cf service "${name_of_deployer}"`
+1. Open the Fugacious link and save the credentials locally in the `credentials.yml`
+  - `cf service ${name_of_deployer} | grep Dashboard | awk '{ print $2 }'`
+1. If the deployer user is used for more than a `cf push`, give it the proper
```
cnelson: I don't think we should do this;  If a deployer account needs special perms, then we should use an account that's defined in the cf manifest with those perms so it's documented, rather than hand modifying brokered deployer accounts.

cnelson (commented):
Dang, that's a lot of documentation @rogeruiz;  Well done 👏 

Just a few things I think are incorrect / should be handed differently, thoughts?

content/docs/ops/runbook/rotating-cloudfoundry-diego.md
rogeruiz: true that.

content/docs/ops/runbook/rotating-cloudfoundry-diego.md
rogeruiz: Ah perfect. thanks.

content/docs/ops/runbook/rotating-secrets.md
rogeruiz: yeah good point. are we worried at all about have those users extending the CF deployment? I'm thinking since most of these deployers exist in the same `org` and `space` we would only need a handful of the ones we currently have deployed.

cnelson: I'd rather have a long, repeatable deployment than a fast inconsistent one :)  Also we can set override: false on all our users once created so UAA won't spend time recreating them on every startup

https://github.com/cloudfoundry/uaa-release/blob/develop/jobs/uaa/spec#L461-L463

content/docs/ops/runbook/rotating-bosh.md
```diff
@@ -0,0 +1,194 @@
...
+There are known limitations of the BOSH agent when updating the main BOSH
+password for each deployment. Deployments of BOSH should be done serially due to
+the dependencies of where these BOSH directors are being deployed to. Please
+reference the [18F/cg-deploy-bosh](https://github.com/18F/cg-deploy-bosh)
+repository along with the rest of this guide.
+
+### Agent limitations when updating BOSH
+
+The following should be referenced throughout the various deployments. [According
+to the BOSH development team](https://cloudfoundry.slack.com/archives/C02HPPYQ2/p1492784585438034),
```
jmcarp: WDYT about filing a ticket at github.com/cloudfoundry/bosh for tracking purposes? Not sure they're paying slack to keep messages for long.

content/docs/ops/runbook/rotating-bosh.md
```diff
@@ -0,0 +1,194 @@
...
+### External deployments
+
+There are various pipelines that require BOSH credentials which will immediately
+be out of date once the BOSH deployment it targets and cause deployment failures
+and potentially cause downtime. All pipelines depending on BOSH must be paused
+and updated before unpausing them again.
+
+#### Finding deployments with BOSH dependencies
+
+Search GitHub for the `BOSH_TARGET` and `BOSH_ENVIRONMENT` with a `YAML`
```
jmcarp: Could we also look at `bosh-cli deployments`?

content/docs/ops/runbook/rotating-bosh.md
```diff
@@ -0,0 +1,194 @@
...
+deploys Tooling BOSH which deploys Development, Staging, and Production BOSH.
+During a deployment of the BOSH director, BOSH will be unresponsive and will
+lead to errors in [any deployments that interact with the director](#external-deployments).
+Verify that these external deployments have been paused before proceeding with
+these steps.
+
+Besides certificates, all secrets in each BOSH deployment can be generated and
+replaced or can be updated from [rotated IAM users]({{< relref "docs/ops/runbook/rotating-iam-users.md" >}}).
+
+#### Validating BOSH deployments
```
jmcarp: Do we need this section? At the moment, deploying a bosh instance also updates its cloud config, and if the director isn't available, updating cloud config will fail.

content/docs/ops/runbook/rotating-bosh.md
```diff
@@ -0,0 +1,194 @@
...
+#### Deploying Master BOSH
+
+Start with the Master bosh deployment and secrets. This will require that you
+create a new root CA certificate for this deployment and to sign certificates
+for subsequent BOSH deployments (e.g. Tooling, Development, Staging,
+Production).
+
+In the [`cg-deploy-bosh`](https://github.com/18F/cg-deploy-bosh) repository, you
+will find a `generate-master-bosh-certs.sh`. Create a `./tmp` directory
+and run that script in there with the IP address for Master BOSH. The IP address
```
jmcarp, line 126: This address shouldn't be in the secrets anymore--now it comes from terraform.

content/docs/ops/runbook/rotating-bosh.md
```diff
@@ -0,0 +1,194 @@
...
+OUTPUT_FILE=master-bosh.pem \
+PASSPHRASE=${master_bosh_secrets_passphrase} \
+$CG_SCRIPT/encrypt.sh
+```
+
+#### Deploying Tooling BOSH
+
+In the [`cg-deploy-bosh`](https://github.com/18F/cg-deploy-bosh) repository, you
+will find a `generate-bosh-certs.sh`. Create a `./tmp` directory and run that
+script in there with the IP address for Tooling BOSH. The IP address can be
```
jmcarp, line 155: Same point as above--these addresses shouldn't be in secrets now.

content/docs/ops/runbook/rotating-cloudfoundry-diego.md
```diff
@@ -0,0 +1,206 @@
...
+
+## Prepare a local directory for rotating secret files
+
+Create a directory you can use to generate all the various artifacts that need
+to be rotated for the secrets files. Create a `./tmp` directory inside the
+`cg-deploy-cf` repository. Delete this directory at the end of secrets rotation.
+
+## Dealing with multiple secret files
+
+The Cloud Foundry manifest is split into two parts, a `main.yml` and
```
jmcarp: s/cloud foundry manifest/cloud foundry secrets/ ?

content/docs/ops/runbook/rotating-cloudfoundry-diego.md
```diff
@@ -0,0 +1,206 @@
...
+`external.yml`. Internal secrets can be rotated either all at once or on a
+rolling schedule of deployments. You can pull the information needed to decrypt
+these from the `cg-deploy-cf` pipeline. You can easily grab the information
+needed to [decrypt and encrypt secrets]({{< relref "docs/ops/secrets.md" >}}) from
+Concourse using the `fly-cli`.
+
+```sh
+# note the target name `fr` here is the output of `fly targets`
+$ fly -t fr get-pipeline --pipeline deploy-cf | \
+  spurce json | \
```
jmcarp: s/spurce/spruce

content/docs/ops/runbook/rotating-cloudfoundry-diego.md
```diff
@@ -0,0 +1,206 @@
...
+  jq -r '.resources | select(.name | test("common-"))'
+```
+
+### Internal secrets (main.yml)
+
+1. Rotate all UAA user passwords with passwords matching the UAA restrictions
+   set in [the spec](https://github.com/18F/uaa-customized-boshrelease/) for our
+   `uaa-customized-boshrelease`.
```
jmcarp: Up to you, but I don't know that listing these requirements here is useful, since they might change in the future. More to the point, uaa will reject bad passwords, so operators might not need a list of rules here.

content/docs/ops/runbook/rotating-cloudfoundry-diego.md
```diff
@@ -0,0 +1,206 @@
...
+From inside the `./tmp` directory, run the following commands:
+
+```sh
+$ ci_env=fr ../generate-all-certificates.sh --grab-cert && \
+  mv all-cf-certs staging-cf-diego-certs
+$ ci_env=fr ../generate-all-certificates.sh --grab-cert && \
+  mv all-cert-certs production-cf-diego-certs
+```
+
+This script will generate all 25 of the certificates for both Diego and
```
jmcarp: Let's not hard-code certificate counts--the number doesn't really matter and will probably change over time.

content/docs/ops/runbook/rotating-cloudfoundry-diego.md
```diff
@@ -0,0 +1,206 @@
...
+    - Deploy the new encrypt key before the old one for the first deployment.
+      And remove the old encrypt key for the last deployment.
+
+Make sure that Diego deploys after each rotation step for the certificate
+updates from _Security Configuration for Consul_ in order to prevent downtime.
+
+#### Updating certificates in secrets
+
+Replace all the certificates corresponding to the filename and property matching
+in the secrets file, e.g. `certs_dir/hm9000_server.crt` maps to
```
jmcarp: hm9000 is gone now!

content/docs/ops/runbook/rotating-cloudfoundry-diego.md
```diff
@@ -0,0 +1,206 @@
...
+
+External secrets fall into a few categories which range from _rotatable_ with
+coordination with other cloud.gov deployments on one end and _non-rotatable_
+without data loss.
+
+#### Properties used in other deployments
+
+1. `uaa.clients.*.secrets` properties need to be coordinated across various Bosh
+   deployments and cloud.gov apps or service brokers.
+1. `collectd.riemann_server` property for Riemann server address.
```
jmcarp: This should get merged from terraform.

content/docs/ops/runbook/rotating-secrets.md
```diff
@@ -0,0 +1,190 @@
...
+1. [Rotating Cloud Foundry / Diego]({{< relref "docs/ops/runbook/rotating-cloudfoundry-diego.md" >}})
+
+## Rotate CF deployment users
+
+Some of our Concourse pipelines deploy to Cloud Foundry using cloud.gov service
+account deployers. In order to rotate these, you need to target the correct
+org and space in order to create the cloud.gov service deployer account
+instance.
+
+### Auditing all CF deployment users
```
jmcarp: Can we drop this now that we're adding the shared broker deployer, or is it still useful?

content/docs/ops/runbook/rotating-secrets.md
```diff
@@ -0,0 +1,190 @@
...
+1. If the deployer user is used for more than a `cf push`, give it the proper
+  `cloud_controller.admin` scope manually using `uaac`. Otherwise it will fail
+  in the pipeline job in Concourse. Make sure you authenticate with the correct
+  UAA server when escalating the scopes of the deployer user.
+     - `uaac member add cloud_controller.admin ${guid_of_deployer}`
+1. Refly the pipeline with updated credentials and start a new job.
+1. Delete the `${name_of_deployer}_prev` service account deployer.
+   - `cf delete-service ${name_of_deployer}_prev -f`
+
+## Rotate secrets passphrases
```
jmcarp: IMO this *has* to include adding passphrases back to keys in `concourse-credentials`, otherwise we have too much random cleanup to do afterwards.

content/docs/ops/secrets.md
```diff
@@ -66,21 +66,40 @@ All UAA clients and users and associated credentials should be created via the C
...
         ```
 
-    1. Grab the `secrets_passphrase` under the `common-prod` section.
 1. Decrypt the secrets file.
 
     ```bash
-    INPUT_FILE=tmp/cf.yml.enc OUTPUT_FILE=tmp/cf.yml PASSPHRASE=... ./decrypt.sh
+    INPUT_FILE=tmp/cf.main.yml.enc
+    OUTPUT_FILE=tmp/cf.main.yml
+    PASSPHRASE=$(fly --target <target> get-pipeline --pipeline deploy-cf | spruce json | jq -r '.resources[] | select(.name | test("common-prod")) | .sources.secrets_passphrase')
```
jmcarp, line 101: Is this a good time to standardize on `common-production`?

content/docs/ops/runbook/rotating-cloudfoundry-diego.md
rogeruiz: yeah the idea here is that since it's in a deployment, you won't get the _reject bad credentials bit_ until you deploy the thing, and it's not the best feedback loop. ➿

content/docs/ops/runbook/rotating-cloudfoundry-diego.md
rogeruiz: 😅

content/docs/ops/secrets.md
rogeruiz: _yes_ 😉

content/docs/ops/runbook/rotating-bosh.md
rogeruiz: https://github.com/cloudfoundry/bosh/issues/1140 Found an oldie related to changing nats passwords. Will link to that.

rogeruiz:
Running through all the feedback @cnelson and @jmcarp today. After that I'll ping for another review and tighten up any loose ends. 👍 thanks!

content/docs/ops/runbook/rotating-secrets.md
rogeruiz: I'll add these bits. @jmcarp added a new deployer user to the cf manifest, so I'll update this with that new process.

content/docs/ops/runbook/rotating-bosh.md
rogeruiz: I was thinking this would at least be good to run just ensure that nothing is broken with the pipeline. Running the `common-releases-*` shouldn't be that big of a deal? wdyt?

jmcarp: Agreed that it's not a big deal, but what does it buy us? My point is that the deploy-*-bosh tasks already include a step that exercises the director by updating its cloud config.

wjwoodson:
>Migrate UAA Client IDs out of secrets into public manifests to facilitate rotating CF external secrets

External client IDs *should* all be completed and awaiting review:
https://github.com/18F/cg-deploy-logsearch/pull/68
https://github.com/18F/cg-deploy-riemann-firehose-nozzle/pull/4
https://github.com/18F/cg-uaa-extras/pull/98
https://github.com/18F/cg-deploy-monitoring/pull/129
https://github.com/18F/cg-sandbox-bot/pull/34
https://github.com/18F/cg-deploy-uaa-guard-broker/pull/9
https://github.com/18F/concourse-broker/pull/16
https://github.com/18F/cg-deploy-kubernetes/pull/72
https://github.com/18F/cg-deploy-cf/pull/235

content/docs/ops/runbook/rotating-secrets.md
rogeruiz: Actually, I think so... Just for the kinds of users that only require a `cf push` 👍 I included this below it.

> Make sure that cloud.gov Service Account users are only used for basic `cf push`
deployments. If you're deploying service brokers, read the [Rotate Secrets V CF
external secrets section]({{< relref "docs/ops/runbook/rotating-]rotating-cloudfoundry-diego.md" >}}).

content/docs/ops/runbook/rotating-bosh.md
rogeruiz: this is fixed now in 8b4036b55e728e15dd82f7cca9263f9d69fc6eb7

content/docs/ops/runbook/rotating-bosh.md
rogeruiz: this is fixed now in 8b4036b55e728e15dd82f7cca9263f9d69fc6eb7

content/docs/ops/runbook/rotating-bosh.md
rogeruiz: Fixed in d46aababa2ce033ee9f047dbf53fc5f5fb2b8abe

content/docs/ops/runbook/rotating-secrets.md
rogeruiz: Addressed in 3e20ce203c1fc18c59a1d9340f0db199f1da3e98

content/docs/ops/secrets.md
rogeruiz: I think this would be a good thing to update after the fact. I only touched on this file a bit to address `main` / `external` files.

rogeruiz:
pinging @jcscottiii as well for a copy / technical review. But everyone else is also invited to give comments too. 👍 _I'm encouraging operators that haven't done secrets rotation._

content/docs/ops/runbook/rotating-cloudfoundry-diego.md
```diff
@@ -0,0 +1,183 @@
...
+1. `login.providers.*` metadata along with `login.saml.*` properties for
+   certificates and keys.
+    - The metadata comes from our external SAML providers. [These properties
+      cannot _currently_ be rotated without downtime](https://www.pivotaltracker.com/n/projects/997278/stories/142082611).
+
+#### Rotating external UAA client secrets
+
+You will need to coordinate these changes with secrets and passphrase changes
+for various other deployments. You can [reference the list of external
+clients](https://github.com/18F/cg-deploy-cf/blob/master/cf-secrets-example.external.yml#L60-L87)
```
jcscottiii: I wouldn't hardcode the lines for this section of YML. maybe put comments in the YML for which subset of clients you are talking about. (unless you mean all of the clients there, then just reference items under the `uaa.clients` key)

jcscottiii (approved):

PR #880, opened 2017-05-04
Noting DISA impact level two letter

brittag:
Small update to mention the DISA letter based on the FedRAMP P-ATO letter.


mogul (approved):
Probably warrants its own page or explanation at some point, but this is good for now!

PR #885, opened 2017-05-10
Add more advice about deciphering log messages

brittag:
@LinuxBozo gave [useful advice in this support ticket](https://cloud-gov.zendesk.com/agent/tickets/445), so here's a quick effort at throwing that info into the docs.

This needs technical review to make sure I didn't mess something up while reformatting from support ticket to docs. This also isn't the greatest information architecture, but incremental improvements can be more helpful than nothing.


LinuxBozo (approved):
LGTM

mogul (approved):

PR #886, opened 2017-05-10
Rename channels for easier linking

rogeruiz:
n/t

### DIFF
--- a/content/docs/ops/maintenance-list.md
+++ b/content/docs/ops/maintenance-list.md
@@ -9,9 +9,9 @@ The platform requires regular support and maintenance activities to remain in a 
 
 # Is it your first day of support?
 
-- Update the `#cloud-gov-atlas` topic to include your name as the support contact.
+- Update the [`#cg-platform`](https://gsa-tts.slack.com/messages/cg-platform/) topic to include your name as the support contact.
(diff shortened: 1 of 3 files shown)


content/docs/ops/maintenance-list.md
```diff
@@ -29,7 +29,7 @@ The tasks on this checklist should be performed each day.
 
 ## Review and respond to open alerts
 
-Review all recent alerts and notifications delivered to `cg-notifications` and `#cloud-gov-atlas-news`.  
+Review all recent alerts and notifications delivered to `cg-notifications` and [`#cg-platform-news`](https://gsa-tts.slack.com/messages/cg-platform-news/).
```
brittag: Should cg-notifications be linked somewhere too?

rogeruiz: hmm, I don't see that channel anywhere @brittag 

screen shot 2017-05-12 at 12 50 22 pm

rogeruiz: whoops. sorry this was super unclear to me at first too. `cg-notifications` is the Google Group we use for emailed notifications in place of PagerDuty. @brittag clearly this needs to be a little more clear.

brittag (approved):

PR #898, opened 2017-05-22
matching release notes dates

jameshupp:
This is so the email and web dates match.


content/updates/2017-05-22-release-notes.md
```diff
@@ -0,0 +1,48 @@
++++
+date = "2017-05-22"
+title = "Platform Release Notes: May 22, 2017"
++++
+
+Curious what’s new that you might find helpful as a cloud.gov application developer? Here are highlights from our platform updates over the past two weeks.
+<!--more-->
+
+### Researching
```
khandelwal: I know it doesn't fit with the style of the other headings but renaming this to "Volunteers needed" or something like that it will help highlight the call to action.

content/updates/2017-05-22-release-notes.md
```diff
@@ -0,0 +1,48 @@
...
+- The Defense Information Systems Agency (DISA) has issued a [provisional authorization for Department of Defense teams]({{< relref "overview/security/fedramp-tracker.md#how-this-p-ato-helps-your-team" >}}) to use cloud.gov’s FedRAMP P-ATO for systems at the DISA level 2 impact level. This is a followup to the P-ATO; Level 2 is equivalent to FedRAMP Moderate.
+- The latest version of the PHP buildpack supports PHP 7.1.4 and 7.0.18.
+
+### Fixed
+We’ve improved our automated process for updating the part of cloud.gov that routes traffic to your application. Previously, you might have seen occasional 502 errors in your application when we made updates.
+
+### Removed
+The latest version of the PHP buildpack removes support for PHP 7.1.2 and 7.0.16. If your applications rely on one of these versions, you need to update your application to use a supported version of PHP.
```
khandelwal: Here's another call to action that should probably be better highlighted.

content/updates/2017-05-22-release-notes.md
```diff
@@ -0,0 +1,48 @@
...
+
+### Platform releases
+We upgraded the Cloud Foundry deployment to [v258](https://github.com/cloudfoundry/cf-release/releases/tag/v258). This upgrade addresses these security vulnerabilities:
+- [CVE-2017-4972: Blind SQL Injection in UAA](https://www.cloudfoundry.org/cve-2017-4972/)
+- [CVE-2017-4973: Privilege Escalation in UAA](https://www.cloudfoundry.org/cve-2017-4973/)
+
+You should [restage your application](http://cli.cloudfoundry.org/en-US/cf/restage.html) to [incorporate fixes in the base filesystem](https://docs.cloudfoundry.org/devguide/deploy-apps/stacks.html#cli-commands) and ensure you’re running the most recent language version supported by your buildpack.
```
khandelwal: And here too, is another call to action.

khandelwal (changes requested):

content/updates/2017-05-22-release-notes.md
jameshupp: I think I'm going to go with `Help make cloud.gov easier to use` since it's pretty directly for users who have a vested interest in that.

jameshupp: n/m looks like @mogul added the other version, which is sufficient for me

mogul (dismissed):

content/updates/2017-05-22-release-notes.md
jameshupp: Unlike the volunteer ask, this is something we've done before and used this style. I think it's fine to work on this—I'll add it to the release notes template—but I don't want to hold up publishing this edition for it unless you think it's essential.

content/updates/2017-05-22-release-notes.md
mogul: Only because I was faster on the draw... Make it say what you think it should say!

content/updates/2017-05-22-release-notes.md
jameshupp: Same as above. One thing we could do, though, is **use strong text on the CTAs**

mogul: That was my first thought for how to address it, and seems like a good interim step to unblock shipping this one.

khandelwal (approved):

PR #899, opened 2017-05-22
correct reference to access package subtotal

jfredrickson5:
Cell C29 refers to cell C12; should refer to cell F12 instead.


jfredrickson5:
Since these are binary files with no diff to compare to make sure I didn't break anything, maybe a committer wants to just make the changes themselves and close this PR. I won't be offended. 😃

suprenant:
@jfredrickson5 thanks for the catch, this is really helpful! I made the change you suggested on our end and opened this PR: 
https://github.com/18F/cg-site/pull/938/

PR #902, opened 2017-05-24
removes duplicate release notes

jameshupp:
Somehow ended up with both a 5-19 filename and a 5-22 filename.


gboone (approved):
This looks good to me.

brittag (approved):

PR #907, opened 2017-05-31
update to include mention of service account

suprenant:
So we don't confuse users. 

Slack reference: https://gsa-tts.slack.com/conversation/C0ENP71UG/p1495486951589104


brittag (approved):

brittag:
Thanks @suprenant! Made it a little more specific about what users will see (since they won't see that email address right away in their application event logs - they'll only see it there when the service instance starts takes actions). Will leave it unmerged in case you want to take another look before merging.

PR #908, opened 2017-05-31
add new release notes

jameshupp:
@gboone @brittag Would love to knock this out today, but I have time tomorrow if it takes a little longer to give this a proper review.

### DIFF
--- a/content/updates/2017-05-31-release-notes.md
+++ b/content/updates/2017-05-31-release-notes.md
@@ -6,10 +6,12 @@ title = "Platform Release Notes: May 31, 2017"
 Curious what’s new that you might find helpful as a cloud.gov application developer? Here are highlights from our platform updates over the past week and a half.
 <!--more-->
 
-### Changed
-- cloud.gov document now covers only managing apps in the GovCloud environment. This follows the migration of all apps that had been in the East/West environment into GovCloud.
-- Different organizations in cloud.gov are now handled within separate accounts for purposes of encryption calls. Previously, all apps hosted on cloud.gov were handled through a single account with the encryption provider, and rate-limiting was causing TLS certificates to expire. Using separate accounts resolves this issue.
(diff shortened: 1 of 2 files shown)


gboone (dismissed):

brittag:
Thanks James! Sorry I didn't get to this earlier.

I put in a suggested edit, since I know you're rolling off, and I know Greg has a time-sensitive thing to do:

* For the E/W docs removal, that makes more sense to me as a “Removed” item instead of a “Changed” item (we removed stuff), and the service improvements seem like “Fixed” items (we fixed problems).
* For the improvements, I tried describing them in terms that users are more familiar with.

The ordering of the info under **Platform releases** is a bit confusing to me as a reader - I missed this in the last update post, but I’m seeing it now. It sort of looks like we’re telling people that they need to restage to benefit from the fix to the UAA problem, but I don't think that's what we mean - I believe we mean they should restage to benefit from the “Additional upgrades”.

I don't know if that's confusing to anyone else though, or even if I'm understanding this correctly.

Also, in this case, the link to the dashboard release notes may be confusing since it talks about improvements that we temporarily rolled back; we may want to omit the dashboard release notes link this time?

jameshupp:
I see what you mean. In my head "base filesystem" referred to CF specifically, but a) that may not be correct and b) you're right that it appears to modify only fixes to the CVE. I'd put it at the very end, where I think it makes sense, but I'm a little afraid of it getting lost after the long list of USNs.

We could put it at the beginning and rewrite it as something like `cloud.gov now includes the following releases and upgrades. You will need to restage your application to incorporate fixes in the base filesystem and ensure you’re running the most recent language version supported by your buildpack.` (with all appropriate linking included there) Then follow that with the CF upgrade and go on to the additional upgrades from there.

brittag:
Yes, I like that structure! That puts the part most relevant to the reader first (it's a good time to restage your apps), and follows it up with the details about why.

brittag (approved):

PR #910, opened 2017-06-01
Tracking quota usage

pburkholder:
From the pinned message: https://gsa-tts.slack.com/archives/C09CR1Q9Z/p1480527207005021

### DIFF
--- a/content/overview/pricing/quotas.md
+++ b/content/overview/pricing/quotas.md
@@ -48,3 +48,14 @@ In this situation you have three options:
 1. Delete existing resources with `cf delete`, `delete-service`, `delete-route` or similar.
 2. Reconfigure individual existing [Application Quotas]({{< relref "docs/apps/limits.md" >}}) and redeploy.
 3. Request a quota change by [asking support](/help/).
+
+
+## Tracking quota usage
+
+There are two useful `cf` CLI plugins that address common needs around right-sizing your instance count and memory limits. There are both available from the _CF-Community_ plugin site when you run: `cf add-plugin-repo CF-Community http://plugins.cloudfoundry.org/`
+
+- The Usage Report plugin gives you a report of how your quota is used across visible organizations and spaces (admins, :warning: it takes forever and may crash on you due to lots of requests/data): https://github.com/krujos/usagereport-plugin. To install: `cf install-plugin 'Usage Report' -r CF-Community`
+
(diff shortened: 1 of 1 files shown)


brittag:
Ahh, this brings up something really interesting: the "product"/overview section is mostly for prospective users learning about cloud.gov capabilities but often not using it yet, and the "docs" section is mostly for current users who need practical info about using it in practice - but we don't yet have a docs page where this would fit really well.

We just have https://cloud.gov/docs/getting-started/concepts/ and https://cloud.gov/docs/apps/limits/ but not really a page for "hey Org Manager, here is info to help you manage your org, such as how to monitor your quota use". So we might need a new page for this!

brittag:
We might want to list https://github.com/18F/cf-route-lookup in that kind of Org Manager page as well, and cross-link to https://cloud.gov/docs/apps/plugins/

pburkholder:
I think an org manager section would be great. Do want me to tackle in the scope of this PR, with new sections so we have this:

* Getting Started
* Developing Apps \[Rename from "Managing apps"\]
  * ("Services" should move into here)
* Managing Orgs \[New\]
  * Essentials (new content, including these plugins)
  * Managing teammates (move from Managing applications)
* Experimental features

?

OR - should I shim this into the Application Limits page and create a new issue for site re-org?

pburkholder:
Oops - wrong button. Reopening.

brittag:
To keep it tiny white-noise work, I'd suggest:

(1) Update https://cloud.gov/docs/apps/limits/ to be called something like "Managing quotas and limits"
(2) Split it into three sections: org, space, app.
(3) Move existing app stuff into app section.
(4) Add one line in space section linking to CF docs about space quotas
(5) Put this new stuff under the org section.

But if it starts getting larger than white noise, we can ask Customer squad to take a look as a story for later.

fureigh:
@pburkholder Do you have capacity to move this forward?

fureigh:
I added a Favro card, [Create org manager section of the docs](https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-5865), for the documentation reorganization work @brittag's identified here and in #926.

jmcarp:
Is this patch still active @pburkholder @fureigh ? Let's finish it if so or close if not.

brittag:
I'll take a look at finishing this PR according to the five steps listed above - basically moving this good addition into the current-user docs instead of the prospective-user docs.

brittag:
Consolidated this info into this PR: https://github.com/18F/cg-site/pull/996

PR #911, opened 2017-06-01
Remove East-West-specific styles

fureigh:
...now that we're shutting down East-West.

Noticed while working on https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-4826.

### DIFF
--- a/static/assets/css/main.css
+++ b/static/assets/css/main.css
@@ -50,15 +50,6 @@ blockquote li,
   margin-top: 8px; /* $grid-1 */
 }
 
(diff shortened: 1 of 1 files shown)


rogeruiz:
@fureigh we've already removed all the EW content right? if so I'll approved this. 👍

suprenant:
@rogeruiz following up here – yes, we removed the EW content, so we're good to go here if you want to review. cc @fureigh

brittag (approved):

fureigh:
Thank you!

PR #915, opened 2017-06-04
Show index on 404.

jmcarp:
So that users don't see potentially confusing nginx errors on urls like https://cloud.gov/lol. Alternatively, we could add a custom 404 page: https://gohugo.io/templates/404.


mogul:
Seems worth doing it the Hugo way. 🤷‍♂️

cnelson:
Can we move the custom nginx config for this app into the `location_include` now that it is a thing, rather than overwriting the whole nginx.conf ?

jmcarp:
Updated @cnelson. By the way, the `.cfignore` ignores `public`, so I'm not exactly sure how this is working at the moment--I pushed to staging manually but had to drop that ignore to get it to work.

nginx.conf
```diff
@@ -37,6 +37,7 @@ http {
       root <%= ENV["APP_ROOT"] %>/public;
 
       index index.html index.htm Default.htm;
+      error_page 404 /index.html;
```
cnelson: Can this file be deleted?

cnelson:
It does?  I don't see public in [master](https://github.com/18F/cg-site/blob/master/.cfignore) or [this PR ](https://github.com/18F/cg-site/blob/404-page/.cfignore )

jmcarp:
I think so--it's a symlink to `.gitignore`, which ignores `public`.

rogeruiz:
I'm all for a simple little 404 based on the homepage, shouldn't require much design but then again those are _famous last words of mine every time_.

cnelson:
Apparently I don't know how to read Github 😁 ; I'm reluctant to merge this if we don't know how it's working.

jmcarp:
I experimented with this a bit, and it looks like setting the path to `public` in the manifest avoids skipping that directory via `.cfignore`.

jmcarp:
WDYT about adding a staging deploy for this @cnelson ? We can deploy manually for now to verify that it works, but might be nice to automate.

layouts/404.html
```diff
@@ -0,0 +1,31 @@
...
+        <div class="usa-width-whole billboard section-bg">
+          <h1 class="billboard-display_text lh-hed">404 Page not found</h1>
+        </div>
+      </div>
+    </section>
+
+    <section class="usa-content section-middle section-light">
+      <div class="usa-grid">
+        <h2 class="section-content usa-content">
+          Looks like that page doesn't exist! If you think it should, send us a <a href="https://github.com/18F/cg-site/pulls", target="_blank">pull request</a>, or
```
brittag, line 21: Probably this should be "send us a pull request or [contact us](https://cloud.gov/docs/help/)" since likely many of our users aren't allowed to use GitHub because of agency restrictions.

jcscottiii (approved):

PR #919, opened 2017-06-06
ByeBye E/W, Hello support boilerplate

pburkholder:


content/docs/ops/aws-accounts.md
```diff
@@ -5,10 +5,22 @@ menu:
...
+## AWS account
 
-1. 18F E/W cloud.gov installation: all cloud.gov apps and services run under the same account but this is the 18F account we are moving out of.
-1. GovCloud cloud.gov main account: this is the "new" AWS account where BOSH/CF API and most services live. This account is where we are migrating tenants to.
+cloud.gov currently (June 2017) uses 2 AWS accounts:
+
+1. GovCloud cloud.gov main account: this is the main AWS account where BOSH/CF API and most services live. This account is where all our tenants reside, including those that used to be hosted in the retired 18F E/W account.
 1. cloud.gov E/W: We only use this for CloudFront right now since it is not available in GovCloud. **This account is linked to the GovCloud one so any support tickets should be filled here**.
 
 Note: DNS for `cloud.gov` is managed through [https://github.com/18f/dns](https://github.com/18f/dns), which is deployed to a Route 53 service in yet another account. cloud.gov does not have access to that account.
```
rogeruiz: Is this currently true? I thought that `18F/dns` changes come through in the cloud.gov E/W account that's tied to GovCloud? I mean, this is where I've seen the updates to cloud.gov related Route 53 pop over in there.

mogul: Yeah, Route 53 is in there too.

mogul: Oh actually I misread... Not sure if it's in our E/W account, or another 18f E/W account. I just know Route 53 isn't in GovCloud. So it's not just Cloudfront that we rely on E/W for.

rogeruiz: confirmed @cnelson this in our overflow for standup today. 18F/dns lives in our cloud.gov EW account. I also confirmed it with the pipeline values that talk about showing the use of the `dns` IAM user in cloud.gov EW. @pburkholder

PR #920, opened 2017-06-06
Document our egress addresses

cnelson:
Do I need to say more than this for the purposes of https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-5505 ?

### DIFF
--- a/content/docs/apps/static-egress.md
+++ b/content/docs/apps/static-egress.md
@@ -5,21 +5,21 @@ menu:
 title: Communicating with external services
 ---
 
-If your cloud.gov-hosted application needs access to an external service it must be Internet accessible. All traffic from cloud.gov applications will transit the public Internet.
+If your cloud.gov-hosted application needs access to an external service, that service must be accessible over the internet. All traffic from cloud.gov applications will transit the internet.
 
 ## Securing communications with external services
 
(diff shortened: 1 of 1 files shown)


jmcarp (dismissed):

brittag:
I _think_ the first sentence might be more clear as the following sentence, but I'm not sure I've fully preserved the meaning:

> If your cloud.gov-hosted application needs access to an external service, that service must be accessible over the internet. All traffic from cloud.gov applications will transit the internet.

Under "Securing communications with external services", is there a reasonable way to structure this page so that we put our recommendation first and the IP option as a backup second item, so that the reader would read something more like the following structure?

> You can set up your cloud.gov application to communicate with any external service available over the internet. To ensure _only_ your application can communicate with the external service, the service should provide application-level authentication.

> If the external service also requires IP filtering, you can do the following. Caveat about this allowing access from any cloud.gov app, which is why the application-level auth is important.

If that's not reasonable, it's fine to leave the page as-is.

cnelson:
@brittag I swapped the order as you suggested.  I still think it reads ok.  WDYT?

brittag (approved):
This makes sense to me, thanks @cnelson!

PR #929, opened 2017-06-15
Update rundown of AWS accounts

fureigh:
See commit messages for details.

One outstanding question, not created by this PR but not yet addressed in it: what/where is the "here" referenced in "any support tickets should be filed here"?


jcscottiii:
Looks good! Let's wait until E/W is shutdown completely then we can merge this.

mogul (approved):

PR #931, opened 2017-06-15
Update configuration management plan

wjwoodson:
Update change workflow to allow a committer to merge an approved PR. Based on discussion in quarterly policy review.


brittag:
This makes sense to me. I think we need to also add something about how we handle commits made after approval - for example, whether all of our repos should be configured with "Dismiss stale pull request approvals when new commits are pushed" enabled.

mogul (approved):
Gonna leave that question for a separate PR. (Britta, feel free to make that one if you feel strongly that we should state the policy now.)

brittag:
Agree with not blocking on that; we may need to do a more comprehensive review of the plan soon anyway.

PR #932, opened 2017-06-15
Update old pre-IDP + E/W bits in managing users

brittag:
The main change is that `ORGNAME_deployer` accounts should no longer be a thing since E/W is deprecated. The other bits are just updates to fix a link and accommodate the existence of the cloud.gov IDP.


mogul (dismissed):
Approved, assuming the conflicts are cleaned up.

brittag:
@mogul Can you take another look at this? Cleaned up the conflicts.

mogul (approved):

PR #934, opened 2017-06-17
Update k8s based on what I see actually works

pburkholder:
The section on NodePorts still needs some help. The commands

```shell
curl http://0.master.default.kubernetes.bosh
curl http://1.master.default.kubernetes.bosh
```

don't work on jumpbox or on the k8 master node.


content/docs/ops/runbook/troubleshooting-kubernetes.md
```diff
@@ -92,14 +96,21 @@ rescheduled:
...
 
-  ## Both of these return the same host via iptables:
-  ```
-  curl http://0.master.default.kubernetes.bosh
-  curl http://1.master.default.kubernetes.bosh
-  ```
\ No newline at end of file
+Both of these return the same host via iptables:
+```
+curl http://0.master.default.kubernetes.bosh
```
jmcarp: I see that this is just an indentation change, but:

* These DNS names are wrong
* The kubernetes machines list aws dns before bosh dns, so these names don't resolve
* If we query bosh dns, the names don't resolve to the same address.

WDYT about deleting this block?

#writeonlydocs

content/docs/ops/runbook/troubleshooting-kubernetes.md
```diff
@@ -14,25 +14,34 @@ Alerts are generated whenever a pod's status is not `Running`. Alerts contains t
...
+### Other useful Kubernetes `kubectl` commands 
 
-### Other Useful Kubernetes commands
-
-#### Set Path
-Kubernetes binaries are shipped in a BOSH package:
-```shell
-export PATH=$PATH:/var/vcap/packages/kubernetes/bin
-```
+All of these assume your are logged into a Kubernetes master:
```
jmcarp: s/your are/you are/

jmcarp:
@pburkholder: I left some comments on the diffs. We can now also shorten `bosh-cli` to `bosh`. Have a few minutes to address comments and fix merge conflicts?

pburkholder:
Thanks for the comments, @jmcarp. I indeed just removed the whole section at the end, as the nodeports bit wasn't useful.

content/docs/ops/runbook/troubleshooting-kubernetes.md
```diff
@@ -14,25 +14,36 @@ Alerts are generated whenever a pod's status is not `Running`. Alerts contains t
...
+sudo bash
 ```
 
-The `Events` section should indicate why the pod cannot be started.  Resolve the underlying issue and the pod should transition into a `Running` state.  If you need to force the pod to be restarted you can delete it which will cause kubernetes to immediately recreate it.
-```sh
-/var/vcap/packages/kubernetes/bin/kubectl --namespace :namespace delete pod :pod-name
+#### Fixing non-running pods
+Login to a kubernetes master, find the namespaces, and describe the pods:
+```bash
+kubeclt get namespaces
```
jmcarp: s/kubeclt/kubectl/

content/docs/ops/runbook/troubleshooting-kubernetes.md
```diff
@@ -14,25 +14,36 @@ Alerts are generated whenever a pod's status is not `Running`. Alerts contains t
...
-The `Events` section should indicate why the pod cannot be started.  Resolve the underlying issue and the pod should transition into a `Running` state.  If you need to force the pod to be restarted you can delete it which will cause kubernetes to immediately recreate it.
-```sh
-/var/vcap/packages/kubernetes/bin/kubectl --namespace :namespace delete pod :pod-name
+#### Fixing non-running pods
+Login to a kubernetes master, find the namespaces, and describe the pods:
+```bash
+kubeclt get namespaces 
+kubectl --namespace :namespace describe pod :pod-name
 ```
+where you provide values for :namespace and :pod-name, where a pod-name is
```
jmcarp: Should use "is" or "will be", not both.

content/docs/ops/runbook/troubleshooting-kubernetes.md
```diff
@@ -14,25 +14,36 @@ Alerts are generated whenever a pod's status is not `Running`. Alerts contains t
...
+will be something like `xc956b1d94dd64-master-0`
 
+The `Events` section should indicate why the pod cannot be started.  Resolve the underlying issue and the pod should transition into a `Running` state.  If you need to force the pod to be restarted you can delete it which will cause kubernetes to immediately recreate it:
```
jmcarp: This is potentially misleading. Kubernetes will recreated deleted pods that are part of a statefulset / deployment / daemonset / etc, but it's also possible to run pods that aren't managed by one of these resources. May want to add that caveat or drop this note.

jmcarp:
I noticed a few minor issues. I'm approving now--feel free to merge when you're happy with this @pburkholder.

jmcarp (dismissed):

content/docs/ops/runbook/troubleshooting-kubernetes.md
```diff
@@ -14,26 +14,39 @@ Alerts are generated whenever a pod's status is not `Running`. Alerts contains t
...
+From a jumpbox in the appropriate environment:
+- ssh to a master
+- update the PATH to find the kubernetes binaries
+- sudo to root for all the `kubectl` commands
+
 ```sh
-/var/vcap/packages/kubernetes/bin/kubectl --namespace :namespace describe pod :pod-name
+bosh -d kubernetes ssh master/0
+PATH=$PATH:/var/vcap/packages/kubernetes/bin
+sudo bash
```
rogeruiz: Shouldn't these two commands be reversed because the sudo command will run in a new shell with the updated variable? Also we need to `export` it as well right?

```sh
$ sudo bash # this is the same as `sudo -i` right?
$ export PATH=$PATH:/var/vcap/packages/kubernetes/bin
```

rogeruiz (changes requested):
Looks good to me too, but I think the beginning needs to be reordered. @pburkholder

content/docs/ops/runbook/troubleshooting-kubernetes.md
pburkholder: When I reviewed this I'd thought I'd gotten them out of order too.... but it worked and I just triplechecked:

screen shot 2017-08-11 at 12 25 04 pm

However, I'll stick to the "principle of least surprise" reorder those.

rogeruiz: wow. that's cool. but awesome. thanks!

rogeruiz (approved):
:shipit:

PR #935, opened 2017-06-17
More on expected CloudTrail events

pburkholder:


content/docs/ops/maintenance-list.md
```diff
@@ -71,26 +71,44 @@ that cards to fix alerts are prioritized properly.
...
+aws cloudtrail lookup-events --lookup-attributes AttributeKey=EventName,AttributeValue=$event_name
+```
+
+These `EventNames` should be attributed to human individuals on the cloud.gov team:
 - ConsoleLogin
-- CreatePolicy
-- CreateSecurityGroup
+
+All of human-generated events should be mapped to named users, e.g. `firstname.lastname`, and NOT to `Administrator`.
+The source IPs should be within the GSA CIDR range, and the dates and times should correspond to users' working hours.
```
jmcarp: Do we require operators to connect to vpn before using the aws cli / console?

wjwoodson (approved):
:shipit:

PR #936, opened 2017-06-17
Check that users have MFA

pburkholder:


cnelson (approved):

jmcarp:
@pburkholder: wdyt about automating this check instead of adding to the list of manual support tasks? I know this will just take a second to run manually, but it's still one more thing to remember to do.

pburkholder:
Or implement this?

http://docs.aws.amazon.com/IAM/latest/UserGuide/tutorial_users-self-manage-mfa-and-creds.html

PR #937, opened 2017-06-19
Use cf-service-connect plugin instead of random cli binaries.

jmcarp:
Because we don't want to encourage users to use unmaintained cli binaries from a random s3 bucket, and we also don't want to maintain those binaries.

### DIFF
--- a/content/docs/services/relational-database.md
+++ b/content/docs/services/relational-database.md
@@ -81,41 +81,28 @@ The instructions below are for PostgreSQL, but should be similar for MySQL or ot
 
 #### Using cf ssh
 
-To access a service database, use the [cf ssh]({{< relref "docs/getting-started/one-off-tasks.md#cf-ssh" >}}) CLI command to login to an instance that is bound to the service and download the `psql` binary to that instance:
-
-```sh
-cf ssh APP_NAME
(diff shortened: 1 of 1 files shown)


jcscottiii (approved):

brittag:
What about Windows users though, since they can't use cf-service-connect? :/ It's likely that a large number of our users outside 18F are on Windows.

brittag:
Quick Google Analytics check says 48% Windows, 33% Mac OS for cloud.gov visitors.

cnelson:
Well we can fix https://github.com/18F/cf-service-connect/issues/13 or add a note to the docs telling windows users they need to use the `-no-client` option and connect their client manually.

Either way we we can't rely on hosting binaries for a user to download into their container for this use case any more.  As soon as we enable support for MS-SQL and Oracle (currently in-flight) we won't have clients for those databases as there's no MS-SQL client for 14.04, and I don't think we can redistribute oracle's instant client.

This means the user will have to use a client on their local machine to connect to them.

brittag:
👍 Not tied to any particular implementation of support for Windows, just want to make sure they have instructions for doing what they need to do.

cnelson:
Verified the plugin works fine on windows and cut a new release with windows binaries: https://github.com/18F/cf-service-connect/releases/tag/1.1.0

PR #939, opened 2017-06-21
Drop unused and outdated operator docs.

jmcarp:
Read these for a nostalgic trip down memory lane :trollface:


cnelson (dismissed):

content/docs/ops/configuration-management.md
```diff
@@ -47,8 +47,6 @@ is encouraged and qualifies as a review. Review should include assessment of arc
 
 ![Pipeline Example](/img/pipeline-example.png)
 
-A more detailed example of this process can be seen in [Updating Cloud Foundry]({{< relref "updating-cf.md" >}}).
```
cnelson: @brittag Do we need to worry about any :compliance: issues when dropping these old, out of date docs since they are mentioned in our config management plan?

brittag: Ah thanks, good to check. 

Process I use for checking: Generally it's fine to make changes that are compatible with the requirements in the CM (Configuration Management) family and compatible with our SSP statements in the CM family. Usually this means it's fine to make configuration management _more_ rigorous (or to make changes that have no impact on rigorousness).

In this case, looking at the SSP, CM-8 (1) says "If there are components to update, remove or test, Cloud Operations follows the update procedures listed here: https://cloud.gov/docs/ops/updating-cf/ "

So I'd hesitate to completely dump that page, especially since readers of the SSP (such as current and prospective customers) may click that link. It'd be better to cut that page to the valid basics of the procedure, or redirect it to something that covers the basics.

mogul (changes requested):
Let's make sure that link still results in something valid, even if it's just a pointer elsewhere.

jmcarp:
Updated.

brittag:
Added a link to https://github.com/18F/cg-deploy-cf so that anyone interested has a bit more to chew on; this seems adequate to me now.

jmcarp:
@mogul: does this need more revisions?

mogul (approved):

PR #944, opened 2017-06-27
changing name from Fugacious to Toaster

suprenant:


suprenant:
FYI, the service is still located at fugacious.18f.gov, so we may need to change this in the future.

brittag (approved):

PR #947, opened 2017-06-28
Document oracle access on request.

jmcarp:
We'll flip oracle to be available to all organizations after govcloud releases se2, hopefully in July.

### DIFF
--- a/content/docs/services/relational-database.md
+++ b/content/docs/services/relational-database.md
@@ -26,6 +26,8 @@ Plan Name | Description | Price
 `large-mysql-redundant`  | Dedicated redundant large RDS MySQL DB instance                               | Will be paid per hour + storage
 `medium-oracle-se1`      | Dedicated medium RDS Oracle SE1 DB; available by request via [cloud.gov support](mailto:cloud-gov-support@gsa.gov) | Will be paid per hour + storage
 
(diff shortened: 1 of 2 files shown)


brittag:
Oh nice! I think this needs a couple more things:

* We need to also update ["Not a good fit"](https://cloud.gov/overview/overview/who-can-use-cloudgov/#not-a-good-fit) ("Your application requires Oracle, SQL Server, or proprietary databases.") - it's a bit more complex now. Not sure how to rephrase this. Maybe we just add that we have some support for Oracle, and link to the docs page?
* This docs page needs a little more info about Oracle, maybe as a small section later in the page, since it's a special case - explaining a tiny bit about why it requires asking us (basically, SE1 vs SE2).

content/docs/services/relational-database.md
```diff
@@ -139,7 +142,7 @@ sftp> quit
 
 Load the dump into your local database using the [pg_restore](https://www.postgresql.org/docs/current/static/app-pgrestore.html) tool. If objects exist in a
 local copy of the database already, you might run into inconsistencies when doing a
-`pg_restore`. This Pg_restore invocation does not drop all of the objects in the database when loading the
+`pg_restore`. This pg_restore invocation does not drop all of the objects in the database when loading the
```
rogeruiz, line 145: 🎯

rogeruiz (dismissed):
:shipit:

cnelson:
@rogeruiz I think the link to amazon is more appropriate than the oracle pdf for explaining why this service isn't available to everyone

mogul:
Doesn't the change also include adding an SQL Server plan? Or did that get broken out separately?

mogul:
[Looks like it does](https://github.com/18F/aws-broker/pull/37/files#diff-c4466882786d0b6479386ae641a8d4d3R422)... Is there a separate PR for the docs on that?

jmcarp:
@mogul: the favro story didn't include sql server, so we left it unfinished and undocumented.

LinuxBozo (approved):

PR #951, opened 2017-06-30
Make the s3 service key example work.

jmcarp:
* Fix broken `cf service-key` usage
* Don't export json; breaks in zsh


LinuxBozo (approved):

mogul:
🤦‍♂️ (90% certain I'm responsible for that.)

PR #952, opened 2017-06-30
Update & standardize code blocks in docs

wjwoodson:


content/docs/ops/maintenance-list.md
```diff
@@ -29,7 +29,7 @@ that week.
 
 - Confirm that all AWS users have MFA configured. The following should return `[]`
 or the users that don't have MFA enabled:
-```shell
+```bash
```
jmcarp: It looks like we use `shell`, `sh`, and `bash` a bunch throughout this repo. WDYT about picking one and standardizing?

LinuxBozo: +1 I always err on the side of `sh` and trying to keep things posix compliant so it's truly portable.

jmcarp (approved):
🚢

brittag:
Non-blocking comment: suggest also adding a short note to https://github.com/18F/cg-site#18f-cloudgov-website documenting the standard that we're using for this.

PR #953, opened 2017-06-30
Document fugacious-free credentials broker interface.

jmcarp:
Document interface proposed in https://github.com/cloudfoundry-community/uaa-credentials-broker/pull/29. Questions for reviewers:

* Is this accurate? (I'm pretty sure it is)
* Is this clear?
* Do you think this proposed interface makes sense?

### DIFF
--- a/content/docs/apps/continuous-deployment.md
+++ b/content/docs/apps/continuous-deployment.md
@@ -21,6 +21,7 @@ Continuous deployment systems require credentials for use in pushing new version
 
 ### Deployer account broker
 
+<<<<<<< HEAD
 You can provision a deployer account with permission to deploy to a single space using the [cloud.gov service account]({{< relref "docs/services/cloud-gov-service-account.md" >}}) service broker:
 
 1. Target the org and space for which you want credentials
@@ -49,10 +50,7 @@ You can provision a deployer account with permission to deploy to a single space
 1. Retrieve your credentials from the dashboard link. Be sure to retrieve your credentials right away, since the link will only work for a brief length of time. Keep these credentials secure. If they’re compromised, the way to invalidate the credentials is to delete the service instance (you can create another, and it will have a fresh set of credentials). <!-- this advice should match on /docs/apps/continuous-deployment/ + /docs/services/cloud-gov-service-account/ + /docs/services/cloud-gov-identity-provider/ -->
 
 To delete your deployer account, delete the service instance:
-
-```sh
-$ cf delete-service my-service-account
-```
+You can provision a deployer account with permission to deploy to a single space using the [cloud.gov service account]({{< relref "docs/services/cloud-gov-service-account.md" >}}) service broker.
 
 ## Continuous integration services
 
@@ -162,7 +160,5 @@ Change `APP` and `DOMAIN` to match your application, and set up the following en
 | CF_PASS | deployer password  |
 | CF_ORG  | target organization|
 | CF_SPACE| target space       |
-
-
 
 You can also add the `alt_appname` attribute to do [Blue-Green deploys](https://docs.cloudfoundry.org/devguide/deploy-apps/blue-green.html).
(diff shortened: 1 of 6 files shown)


rogeruiz:
[I like the usage updates from the broker readme more](https://github.com/cloudfoundry-community/uaa-credentials-broker/tree/no-fugacious#usage) than what's in the docs here. wdyt about just linking to that readme instead?

jmcarp:
I'll defer to @brittag on which docs are clearer.

brittag:
There's also a third set of instructions here: https://cloud.gov/docs/apps/continuous-deployment/#deployer-account-broker

I'm not sure whether it makes more sense from a user-experience perspective to maintain that third set or delete it and ask the reader to jump to https://cloud.gov/docs/services/cloud-gov-service-account/ for the provisioning process.

I believe it's important for readers to be able to stay within the cloud.gov docs (instead of having to cross-reference readmes and docs), but it can be ok to ask people to switch between two docs pages as long as the steps are clear.

brittag:
Cross-linking: [Google doc with text for a post for https://cloud.gov/updates/ about the change](https://docs.google.com/document/d/1kP-_HZODpn3-2ul9uxyyMvP51829Fj59xBdtYQCeHow/edit) - good to review as well.

brittag:
Took another look at this and cleaned up a couple things. I think the last bit is just figuring out how to communicate this to users - the "note" on the two pages should ideally link to an /updates/ post (with the draft text in the Google Doc).

We should probably wait for the next round of release notes, so that we'll have an /updates/ post and email people a notification.

rogeruiz:
I'll take care of the conflicts, but what should we do about the Google Doc link? Should we add that note in the comments so we can merge this in? Or I guess maybe we could get started even though releases notes aren't for another week. @brittag

brittag:
@rogeruiz Basically this requires us to make a judgment call about how disruptive/confusing we think this change will be for users:

* If we think it's somewhat disruptive/confusing: we should finish drafting update text and wait for release notes, since that way everyone will get an email notification about the change around the same time it happens.
* If we think it's not terribly confusing: within this PR we should set up an /updates/ post with our update text (a file in https://github.com/18F/cg-site/tree/master/content/updates) and cross-link it from the "note" in the docs. Then we merge this PR whenever we like, and a future release notes post (and email) can link to it.

rogeruiz:
Had to rebase this because of #952.

brittag:
@rogeruiz I updated your update to remove the details that users aren't familiar with (such as info about UAA) and to remove the commands (since they're in the docs). I also cross-linked the update from the note on docs pages, and I updated the date on everything to Wednesday. :)

This all looks reasonable to me, but if we have bandwidth, I'd like us to get an opinion from a user about whether the docs and update post make sense - we can ask in #cg-support for a volunteer for a 10-minute screenshare.

rogeruiz:
Awesome sounds good. I'll move this story to blocked. We could always wait until the end of the sprint if we find that it's too confusing. I think it'll be okay and a welcomed update, but it is still not a small change.

rogeruiz:
Ran through this documentation as a user and was able to properly find all the necessary bits leading the crumbs from `/docs`. I also posted in #cg-support to see if anyone has time to walk through the new changes from an actual user's perspective. Looks good to :shipit: after that.

```
https://cloud.gov/docs/
https://cloud.gov/docs/getting-started/your-first-deploy/
https://cloud.gov/docs/apps/deployment/
https://cloud.gov/docs/apps/continuous-deployment/
https://cloud.gov/docs/apps/production-ready/
https://cloud.gov/docs/services/cloud-gov-service-account/
https://cloud.gov/updates/2017-07-05-changes-to-credentials-broker/
```

rogeruiz (dismissed):
:shipit:

rogeruiz (approved):
:shipit:

PR #956, opened 2017-07-07
Document credential rotation for services.

jmcarp:
As suggested by @brittag after manual credential rotation for an rds service instance today.

### DIFF
--- a/content/docs/services/elasticsearch24.md
+++ b/content/docs/services/elasticsearch24.md
@@ -37,7 +37,7 @@ that [here](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/dots-in-
 
 ### Rotating credentials
 
-The elasticsearch24 service does not support self-service credential rotation. To rotate credentials, file a ticket with [cloud.gov support](mailto:cloud-gov-support@gsa.gov).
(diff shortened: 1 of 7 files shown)


content/docs/services/elasticsearch24.md
```diff
@@ -35,6 +35,10 @@ ElasticSearch 2.4 allows for dots in field names. This is a feature that existed
 pre-ElasticSearch 2.0 but was disabled in versions 2.0 to 2.3. Learn more about
 that [here](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/dots-in-names.html).
 
+### Rotating credentials
+
+The elasticsearch24 service does not support self-service credential rotation. To rotate credentials, file a ticket with [cloud.gov support](mailto:cloud-gov-support@gsa.gov).
```
brittag: For these ones that would require emailing us, should we also note that an option is to delete and recreate the service (understanding that this may cause service disruption / data loss)? I'm guessing that in some incident response cases, the cost-benefit analysis for delete + recreate would be better than emailing us, and in a tense situation it's easy to forget the simple option. 🤔

brittag (approved):
Looks good to me, thanks @jmcarp!

PR #959, opened 2017-07-07
Move Support For Customers To Top Of Page

commit-dkp:
Folks who need support need it faster than folks who aren't using cloud.gov yet, so move content for them to the top of the page.

### DIFF
--- a/content/docs/help.md
+++ b/content/docs/help.md
@@ -7,7 +7,7 @@ aliases:
 
 ### Support for people who use cloud.gov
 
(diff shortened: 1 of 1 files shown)


brittag:
This is a tricky one, since we also want to make sure that people who don't yet use cloud.gov have an easy way to talk to us. I believe a better solution here is to cut the number of words in the first section, instead of rearranging the order, so that both options are visible together.

commit-dkp:
Could we split the difference and have both options visible together but prioritize the content order for the audience that is likely experiencing an emergency?

brittag:
Closing since we integrated this need into https://github.com/18F/cg-site/pull/963

PR #960, opened 2017-07-07
Clarify IR Starts With 18F Handbook

commit-dkp:
Modified formatting to clarify that IR begins with the steps in the 18F Handbook and then calls back into the steps listed here.


rogeruiz (approved):
:shipit:

PR #961, opened 2017-07-07
Clarify Real Incident Definition

commit-dkp:
Time spent investigating the root cause of an incident inhibits reporting to the required stakeholders than an incident has occurred, so we clarify that incidents are real even if they may not be in scope for cloud.gov.


rogeruiz (approved):
:shipit:

PR #962, opened 2017-07-07
Add IR Strategy

commit-dkp:
IR can get stuck when trying to trace backwards, so describe a method for tracing forwards, too.


rogeruiz (approved):
:shipit:

PR #958, opened 2017-07-07
Add Links For Security Issues

commit-dkp:
Folks who are looking to contact about a security issue are going to search for "security" on the page before they settle for "contact".

PR #963, opened 2017-07-07
Alternate improvements for security issue reporting

brittag:
@commit-dkp Here's an alternate way of helping with the user needs addressed by https://github.com/18F/cg-site/pull/958 and https://github.com/18F/cg-site/pull/959 - curious what you think:

* Cut inquiries contact information to four lines instead of seven. This means that people looking for support info can find it faster (underneath this section). This is better than moving the support info first, since the support info needs to be longer than the inquiries info (it's more complicated than just "email us").
* Cut extra words from customer support info, to help people find what they need faster.
* Add "If you need help with an application security incident, the request should come from the System Owner or Org Manager, to help us validate the request." - we shouldn't require this for initial reports, since we haven't communicated to everyone that this is a requirement (and also we don't have a solid procedure yet for helping all customers keep their System Owner info up to date, so we need to be able to fall back on Org Manager, which everyone has and knows how to keep up to date).
* Add "report a security issue" info to help page with info about vulnerability reporting.
* Change footer "Contact" link to "Get help for customer issues" (with a direct link to support section of contact page), so that if customers don't think the "contact" header link is for them, this link will catch them.


commit-dkp:
@brittag Looks good to me!

jmcarp:
What's the status of this work @brittag @commit-dkp ?

commit-dkp:
@jmcarp Appears to be pending review by someone with write access.

brittag:
Fixed merge conflict - ready for review. I'd be happy to talk through these changes with somebody since they're not super-trivial changes.

brittag:
@femmebot This might be interesting to take a couple minutes to look at together, if you're interested - it's a few changes that try to improve the experience of requesting help in a few different situations that have different needs.

brittag:
@femmebot Ah I missed this note, sorry! I don't see GitHub messages right away. I'll put 20 minutes on our calendars tomorrow. :)

femmebot (commented):
LGTM! 

(note to self: look into layout / typographic styles to better support varying degrees of emphasizing content to help users quickly find what they need and act upon it)

brittag:
@femmebot Thanks! If you want to merge this, use the "approve" option when reviewing, and then it'll give you the "merge pull request" option. After merging, it'll auto-deploy to the live site.

femmebot (approved):

PR #966, opened 2017-07-10
Add clear documentation for the Staticfile buildpack configuration

rememberlenny:


brittag:
@rememberlenny Thanks, ran this locally and it looks good! Here's how it looks, which helps explain why we don't need an initial "Overview" heading:

screen shot 2017-07-10 at 12 17 47 pm

brittag (approved):

PR #976, opened 2017-07-19
Try to make RDS page more helpful

brittag:
Changes:

* I removed "Note that for a Rails app, `bind-service` will [overwrite your `database.yml`](http://docs.cloudfoundry.org/buildpacks/ruby/ruby-service-bindings.html#rails-applications-have-autoconfigured-database-yml)." because [that information is no longer on the linked docs page](https://docs.cloudfoundry.org/buildpacks/ruby/ruby-service-bindings.html).
* For "manually accessing the database", I removed "The only type of database supported via this method is PostgreSQL" because the cf-service-connect readme explains [support for MySQL and Redis](https://github.com/18F/cf-service-connect#support) as well. I moved "The instructions below are for PostgreSQL, but should be similar for MySQL or others." to what looked like a more appropriate spot for the same reason.
* I toned down the cg-migrate-db description because it may not be as helpful as cf-service-connect for many use cases.
* I added a simple note about RDS instances being encrypted at rest, since people ask about that. We should also explain [answers to this question](https://gsa-tts.slack.com/archives/C09CR1Q9Z/p1490293989829692) for the Redis, Elasticsearch, and S3 pages, but that can be another PR.

### DIFF
--- a/content/docs/services/relational-database.md
+++ b/content/docs/services/relational-database.md
@@ -146,7 +146,11 @@ dump.
 $ pg_restore --clean --no-owner --no-acl --dbname={database name} backup.pg
 ```
 
-### Rotating credentials
+## Encryption
+
+Every RDS instance configured through cloud.gov is [encrypted at rest](http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Overview.Encryption.html).
+
+## Rotating credentials
 
 You can rotate credentials by creating a new instance and deleting the existing instance. If this is not an option, email [cloud.gov support](mailto:cloud-gov-support@gsa.gov) to request rotating the credentials manually.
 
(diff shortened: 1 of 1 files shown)


content/docs/services/relational-database.md
```diff
@@ -149,7 +146,11 @@ dump.
 $ pg_restore --clean --no-owner --no-acl --dbname={database name} backup.pg
 ```
 
-### Rotating credentials
+## Encryption
+
+Every RDS instance configured through cloud.gov is [encrypted at rest](http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Overview.Encryption.html).
```
mogul, line 151: Do you think we should link to the place in our broker config where this happens, @brittag? So that customers can use that in the course of gathering "evidence" for their auditors that we do in fact do this?

mogul (approved):

content/docs/services/relational-database.md
brittag: @mogul Interesting idea, but I think this need would be better served by adding a note saying something like "For details, see SC-28 in the cloud.gov SSP in our [P-ATO documentation package](https://cloud.gov/overview/security/fedramp-tracker/#how-you-can-use-this-p-ato)." Checking whether a config setting does what it's supposed to do is really hard for an outside assessor, but assessors should be able to trust the P-ATO documentation package since it has the 3PAO and JAB review.

mogul:
Good point!

PR #983, opened 2017-07-26
Remove manual method for domains

brittag:
The CDN service method is better, and we haven't received any requests for the manual method since we made the CDN service available to everyone, so we don't need this in the docs.


jmcarp (approved):

jmcarp:
Thanks @brittag !

PR #987, opened 2017-07-26
Org Managers / System Owners should request memory quota changes + explain sandbox case

brittag:
We should make sure memory quota change requests come from somebody who can likely authorize billing changes.

### DIFF
--- a/content/docs/ops/quotas.md
+++ b/content/docs/ops/quotas.md
@@ -7,6 +7,8 @@ title: Managing quotas
 
 Quota modification is most commonly necessary for increasing their resource allocation or enabling paid services. To modify the quota:
 
+1. Check whether the requester is an Org Manager or System Owner for that org. If they're not, ask them to get an Org Manager or System Owner to confirm the request.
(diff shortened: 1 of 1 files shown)


fureigh (dismissed):
Thanks, for this, @brittag!

One question: where would I check whether the requester is a System Owner? The only roles I see in Cloud Foundry are `Org Manager`, `Billing Manager` and `Org Auditor`.

And a minor copy editing nitpick: I'd change `since` to `as` because this isn't time-related.

mogul:
@fureigh: Info about the System Owner is captured as part of the customer info in https://customers.fr.cloud.gov

mogul:
Just found out that the system owner isn't recorded in that service yet. Will edit to point to the right place.

mogul:
@fureigh care to approve?

el-mapache (approved):
This is great, super helpful!

brittag:
@el-mapache Thanks! Go ahead and merge if it looks good to you; it doesn't need additional review unless you think somebody else needs to look at it.

fureigh:
@mogul @brittag LGTM! Thanks!

PR #990, opened 2017-07-28
Oracle SE2 is generally available.

jmcarp:
Does this need an entry in `/updates`?


brittag:
@jmcarp It's not a breaking change for anyone, so we can just put this in the next release notes. :D

mogul (approved):

PR #992, opened 2017-07-28
[WIP] Document high-availability elasticsearch plans.

jmcarp:
Not ready for merge until we finalize the details of the HA plans.

This probably needs a post in `/updates`, and we might want to announce this at the same time as the HA redis service.

cc @cnelson @rogeruiz @mogul @brittag

### DIFF
--- a/content/docs/services/elasticsearch24.md
+++ b/content/docs/services/elasticsearch24.md
@@ -5,19 +5,23 @@ menu:
 title: Elasticsearch 2.4
 name: "elasticsearch24"
 description: "Elasticsearch version 2.4: a distributed, RESTful search and analytics engine"
-status: "Alpha"
+status: "Beta"
 ---
 
-cloud.gov offers [Elasticsearch](https://www.elastic.co/) 2.4 as a service. **This service is still in Alpha mode**: downtime and data loss is possible.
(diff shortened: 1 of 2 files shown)


brittag:
This looks pretty good! For the note up top about backups, do you think it needs to be super up front? (In other words, does this break people's normal expectations of what an Elasticsearch service should do, so they need to know about it first?)

If it's relatively normal information, I'd recommend moving that information below into the "body" of the documentation (probably under "Managing backups"), so that people can read it in context when they're learning about the service capabilities and options.

cnelson:
I think we should call this `Beta` and maybe still make it free.  We should have some customers running these in production for a few months without issue before we call it GA.

mogul:
:thumbsup:

jmcarp:
Updated to call the service beta, and to move the note about backups to the backups section. This is still `[WIP]` until we sort out plans.

brittag:
@jmcarp Should this PR also remove Elasticsearch from https://cloud.gov/docs/apps/experimental/services/ ?

brittag:
I see a couple copyediting things here but not sure if this is ready for copy-level edits or if there's more to be done here first; feel free to reach out when ready!

jmcarp:
Ready for edits @brittag !

cnelson (approved):

PR #993, opened 2017-07-28
FEC case study

gboone:
This would publish the case study we did on FEC in a pretty basic way.

For some reason this isn't building right and consequently I don't really know how it'll look. I thought it would end up at /case-studies/fec/ but whenever I build the site locally I get infinite redirects from that url to itself. This is my first time using Hugo so I'm _sure_ I did something wrong. Any advice is much appreciated!


brittag:
@gboone When I run this locally, I get a 404 at `/case-studies/fec/` until I locally put something in `/layouts/case-studies/single.html`, and then I see a page as expected. Happy to show how the layouts work, or another customer team member can as well.

gboone:
Ooooh, that makes sense. I'll keep working on this, thanks for the tip.

gboone:
OK, I added a layout in that directory and now the page isn't building. Good news is I'm not seeing the infinite redirects thing anymore but I'm a little confused since it seems like I have everything set up right. 

I'll try and pair with someone on the  team next week and figure out what I'm doing wrong. In any case, if someone can review this and get the page to build, I think this is ready to merge.

khandelwal:
Are these answered somewhere else:
Where on the site will this live? 
How will it be linked to? 
How will users navigate to it? 

Just wondering if I'm missing a PR that I should also be reading.

gboone:
Good questions and you're not missing a separate PR. 

This will publish it at `/case-studies/fec/` and I can rename it or add aliases for other URLs if we want. We'll have a link to it in the quarterly newsletter and could send it to individuals as needed.

I'm not sure how we want to link to it or have users navigate to it organically yet. We could add something to the homepage for it but I wouldn't want to do that without some design opinions on where and how. I think once we have more than one case study we could think about how we want to feature them on cloud.gov. 

For now, this quietly publishes the case study and gives us a canonical URL we can send to people.

mogul:
It's a common pattern to see summaries or pull-quotes from case-studies on the front of a product's site, which you can click on to "learn more". @femmebot do you have thoughts?

mogul:
(And yes, I agree, that we can publish this and give people the URL as a way to get it out there more quickly, but the sooner we can get it working for us "on its own" by being discoverable without talking to us first, the better!)

brittag:
It seems ok to me for now to make another sub-section of the product/overview section with something like "Customer stories" and have this as the only page there - it'd be a simple way to make it public, and we could rearrange later as we add more info.

gboone:
OK, I changed up the publishing on this so it publishes under "customer stories" and added a menu section for it nested inside "overview." I think we're ready to go here?

brittag (approved):
Looks good to me! Added a link to the current site.

gboone:
Strange: I'm not seeing it where I expect to https://cloud.gov/overview/overview/what-is-cloudgov/ Does the site take a bit to build?

brittag:
Followed up in #cg-customer! In case anyone else is interested: usually the site takes about 5 minutes to build - if it's taking longer than that for changes to show up, check on whether the build didn't work. :)

PR #994, opened 2017-07-30
Removing multi-buildpack reference

mogul:
Now multi-buildpacks are officially supported, and we'll have them as of the deployment of [the new release](https://github.com/cloudfoundry/capi-release/releases/tag/1.36.0) (any day now)! So we may as well stop recommending this path to people to avoid them having to migrate later.


brittag (approved):
This makes sense to me! Added another commit to remove a reference to `cf-buildpack-multi`, so I'll let somebody else merge this.

Issue #253, opened 2016-05-11
Check and fix broken links


afeld:
We check internal links as part of the Wercker build, but I just cam across two broken external ones in a row. We should run [HTML Proofer](https://github.com/gjtorikian/html-proofer) locally and fix them.

suprenant:
@afeld we're doing some board cleanup. Is this still relevant?

brittag:
Switched this to a SkyPorter label since this is about user-facing content. It would still be good to come up with a way to consistently check for broken external links in our content.

brittag:
I ran a link checker on the site and found only one legitimately broken link, "[security reasons](http://www.twobotechnologies.com/blog/2014/02/importance-of-state-in-oauth2.html)" on https://cloud.gov/docs/apps/leveraging-authentication/

I'll fix that, but I think we're doing ok without the automated version.

brittag:
Eh, reopening - this was an important catch actually!

suprenant:
We had a recent update to the Github issue that didn't come through here.

mogul:
Now handled as part of the deployment pipeline

PR #997, opened 2017-08-02
On leveraging auth page, fix info about state parameter

brittag:
https://cloud.gov/docs/apps/leveraging-authentication/ has a [broken link](https://github.com/18F/cg-site/issues/253), so I tried to fix it:

* Replace broken link to security info with a link that might not be super helpful but emphasizes the importance of the parameter and probably won't break
* Add best practice to example
* Give a little more info about best practices ([borrowing from here](https://developers.google.com/identity/protocols/OpenIDConnect?hl=fr#createxsrftoken))

This definitely needs technical review.

### DIFF
--- a/content/docs/apps/leveraging-authentication.md
+++ b/content/docs/apps/leveraging-authentication.md
@@ -36,14 +36,14 @@ query parameters:
 * `response_type=code`
 * (optional) `state=<ANYTHING>`
 
-You can optionally set a `state` parameter with any value you'd like.
-It will be returned to you in a later step. While optional, it's highly
-recommended that you use it for [security reasons](http://www.twobotechnologies.com/blog/2014/02/importance-of-state-in-oauth2.html).
+You can set a `state` parameter with any value you'd like.
+It will be returned to you in a later step. While optional, we *strongly*
+recommend that you use it with a high-quality random number or a hash generated with a secret key, because it [protects against cross-site request forgery attacks](https://tools.ietf.org/id/draft-bradley-oauth-jwt-encoded-state-07.html).
 
 Here is an example:
 
 ```html
-<a href="https://login.fr.cloud.gov/oauth/authorize?client_id=NAME&response_type=code">
+<a href="https://login.fr.cloud.gov/oauth/authorize?client_id=NAME&response_type=code&state=9ab894ad91d99eb9ee4b30ea7f02b9d8e43eb15db58ff93e4894f3b49817d7ab">
   Log in
 </a>
 ```


rogeruiz (approved):
:shipit: 🔏

content/docs/apps/leveraging-authentication.md
```diff
@@ -36,14 +36,14 @@ query parameters:
 * `response_type=code`
 * (optional) `state=<ANYTHING>`
 
-You can optionally set a `state` parameter with any value you'd like.
-It will be returned to you in a later step. While optional, it's highly
-recommended that you use it for [security reasons](http://www.twobotechnologies.com/blog/2014/02/importance-of-state-in-oauth2.html).
+You can set a `state` parameter with any value you'd like.
+It will be returned to you in a later step. While optional, we *strongly*
+recommend that you use it with a high-quality random number or a hash generated with a secret key, because it [protects against cross-site request forgery attacks](https://tools.ietf.org/id/draft-bradley-oauth-jwt-encoded-state-07.html).
```
jmcarp, line 41: @rogeruiz: did you check out this link? I'm not sure JWTs are relevant to this guidance.

rogeruiz: I did. I was originally going to link to [this](https://tools.ietf.org/id/draft-bradley-oauth-jwt-encoded-state-07.html#rfc.section.4.1 or [this](https://tools.ietf.org/id/draft-bradley-oauth-jwt-encoded-state-07.html#rfc.section.4), but felt like it was overkill.

jmcarp: My question is why we're linking to a doc about JWTs at all. We're talking about adding a random `state` parameter to requests to prevent CSRF, right?

rogeruiz: oh dang. yeah you're right. I'll patch it up without that link for this one [http://www.thread-safe.com/2014/05/the-correct-use-of-state-parameter-in.html](http://www.thread-safe.com/2014/05/the-correct-use-of-state-parameter-in.html)

cnelson: The existing link seems valid to me.  It's suggesting one use JWT for state rather than a random string.  Maybe overkill but seems like a good idea to me :)

brittag: Follow-up PR with new link here: https://github.com/18F/cg-site/pull/999 (either link is fine with me)

PR #999, opened 2017-08-02
Use a better link

rogeruiz:
does what it says on the tin.


brittag (approved):
Approved based on the sensible discussion at https://github.com/18F/cg-site/pull/997#event-1188840459

PR #1001, opened 2017-08-03
Add minimal redis backup and restore docs.

jmcarp:
We don't currently offer automated redis backups, so users who need backups have to dump and restore keys themselves. This patch adds minimal docs linking to the relevant redis commands. As we discussed in slack, there are open-source tools that would simplify this process, but we don't have the capacity to vet and maintain those tools (for reference, the tool I'm talking about is https://github.com/pampa/rutil).


brittag:
I added a missing space - looks fine to me otherwise from a copyediting perspective, but I don't know enough about Redis to really make sense of it as a reader. :)

cnelson (approved):

PR #1006, opened 2017-08-06
Add term "SLA" for discoverability

pburkholder:
I met someone at ATARC symposium who said she couldn't find our SLA.

### DIFF
--- a/content/overview/pricing/pricing-terminology.md
+++ b/content/overview/pricing/pricing-terminology.md
@@ -23,7 +23,7 @@ A system maps to an [“org” in cloud.gov and Cloud Foundry](http://docs.cloud
 
 ### Support {#support}
 
-“Support” is support to use the platform as intended, on a best-effort basis. We have no existing service-level agreement. We plan to always publish metrics that give agencies the ability to make an informed choice about whether to use cloud.gov or an alternative PaaS solution based on our actual track record.
+“Support” is support to use the platform as intended, on a best-effort basis. We have no existing service-level agreement (SLA). We plan to always publish metrics that give agencies the ability to make an informed choice about whether to use cloud.gov or an alternative PaaS solution based on our actual track record.
 
 ### Consulting
 


content/overview/pricing/pricing-terminology.md
```diff
@@ -23,7 +23,7 @@ A system maps to an [“org” in cloud.gov and Cloud Foundry](http://docs.cloud
 
 ### Support {#support}
 
-“Support” is support to use the platform as intended, on a best-effort basis. We have no existing service-level agreement. We plan to always publish metrics that give agencies the ability to make an informed choice about whether to use cloud.gov or an alternative PaaS solution based on our actual track record.
+“Support” is support to use the platform as intended, on a best-effort basis. We have no existing service-level agreement (SLA). We plan to always publish metrics that give agencies the ability to make an informed choice about whether to use cloud.gov or an alternative PaaS solution based on our actual track record.
```
khandelwal, line 26: Agreed on the addition of the abbreviation. However, now that I'm looking at this - shouldn't this be phrased differently? 'We have no existing SLA' doesn't seem quite right. Shouldn't it be something more along the lines of: We currently can not offer Service Level Agreements? @brittag

pburkholder: I think this could use a revamp. If we could also be more specific about what historical data we have (or when we'll publish it), that would provide better guidance to our potential customers.

brittag: Oh boy, it looks like this whole page really needs reworking. It's not very helpful to explain complicated topics by giving a list of terms and defining them, instead of centering the problem the user is trying to solve and explaining what they need to know. I'll give it a shot as part of working on the pricing info. :/

pburkholder: How about we merge this 5-character change, then let the rework be a new thing?

brittag (approved):
Approving incremental change!

PR #1010, opened 2017-08-10
Link to live service-level stats

mogul:
There are probably other places where we can or should surface this link, but at least this closely associates the one search hit for `SLA` with a link to view our uptime stats!


khandelwal (approved):
I think this page needs to change, but adding this link here is a good reminder to keep it next to where ever we eventually talk about SLAs and uptime.

PR #1012, opened 2017-08-12
Update using-ssh.md

pburkholder:
Based on the two ssh questions on Friday.

### DIFF
--- a/content/docs/apps/using-ssh.md
+++ b/content/docs/apps/using-ssh.md
@@ -20,3 +20,9 @@ in order to debug your application.
 You can interact directly with the services bound to your application via [port-forwarding](https://docs.cloudfoundry.org/devguide/deploy-apps/ssh-services.html). This allows you to access those services using native clients on your local machine. The [Service Connect plugin](https://github.com/18F/cf-service-connect#readme) makes this even easier.
 
 `cf ssh` uses port 2222. If your network blocks port 2222, you may receive an error message such as `Error opening SSH connection`.
+
+
+## SSH-Related Questions
+
+*Q: What version of SSH does cloud.gov use to access application containers?
+
--- a/content/docs/apps/using-ssh.md
+++ b/content/docs/apps/using-ssh.md
@@ -34,6 +34,6 @@ You can interact directly with the services bound to your application via [port-
 
 ### SSH version information 
 
-The cloud.gov application containers use the SSH-2.0 protocol. The SSH service uses the [CloudFoundry SSH implementation](https://github.com/cloudfoundry/diego-ssh). For more on how CloudFoundry implements SSH, refer to CloudFoundry's documentation on [Understanding Application SSH](https://docs.cloudfoundry.org/concepts/diego/ssh-conceptual.html).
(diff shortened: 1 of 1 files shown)


brittag:
This is good information. I recommend restructuring this as organized declarative information rather than Q&A, so that the page structure is easier to read as it expands! Happy to explain more what I mean by that if helpful. cc @fureigh

(Against FAQs: [18F content guide](https://content-guide.18f.gov/structure-the-content/), [GOV.UK content guide](https://www.gov.uk/guidance/content-design/writing-for-gov-uk#dont-use-faqs).)

content/docs/apps/using-ssh.md
```diff
@@ -20,3 +20,24 @@ in order to debug your application.
 You can interact directly with the services bound to your application via [port-forwarding](https://docs.cloudfoundry.org/devguide/deploy-apps/ssh-services.html). This allows you to access those services using native clients on your local machine. The [Service Connect plugin](https://github.com/18F/cf-service-connect#readme) makes this even easier.
 
 `cf ssh` uses port 2222. If your network blocks port 2222, you may receive an error message such as `Error opening SSH connection`.
+
+
+## SSH-Related Questions
+
+*Q*: When I SSH to my application and I try to run `python`  I get `ImportError: No module named site`. How should I set that up?
```
fureigh: Thanks for writing this up, @pburkholder! With @brittag's recommendation in mind, this could be rephrased as something like `If when you SSH to your application and try to run 'python' you get [...]`. I don't share the opposition to FAQs when they're questions that users have actually asked, as is the case here, but I'm trying to learn from the approach @brittag's recommending.

content/docs/apps/using-ssh.md
```diff
@@ -20,3 +20,24 @@ in order to debug your application.
...
+for f in /home/vcap/app/.profile.d/*.sh; 
+  do source "$f"; 
+done
+export LD_LIBRARY_PATH=/home/vcap/deps/0/python/lib/
+export PYTHONHOME=/home/vcap/deps/0/python
+```
+
+*Q*: What version of SSH does cloud.gov use to access application containers?
+
+*A*: The cloud.gov application containers use the SSH-2.0 protocol. The SSH service uses the CloudFoundry SSH implementation, https://github.com/cloudfoundry/diego-ssh. For more on how CloudFoundry implements SSH, refer to https://docs.cloudfoundry.org/concepts/diego/ssh-conceptual.html.
```
fureigh: I'm inclined to change `the CloudFoundry SSH implementation, https://github.com/cloudfoundry/diego-ssh` to `the [CloudFoundry SSH implementation](https://github.com/cloudfoundry/diego-ssh)`. If you think there's more value in having the URL be explicitly visible, though, I could be convinced otherwise. :)

fureigh (dismissed):

content/docs/apps/using-ssh.md
brittag: On https://cloud.gov/docs/apps/logs/#error-messages we have a section for error messages, and this page could have something similar - it already has one error case (If your network blocks port 2222, you may receive an error message such as Error opening SSH connection.) so the two error cases could be together in a section.

(On that Logs page it's not as effective as it could be since the error descriptions are kind of far down the page from the point where you put in the command, but ideally the error cases are easily found in context at the point of "ok I tried this thing and it didn't work. now what?")

content/docs/apps/using-ssh.md
mogul: I'd definitely make it an implicit link on the text rather than an explicit link-in-text.

pburkholder:
Yep, entirely agree. How now?

brittag (approved):
Looks good to me, thanks Peter!

PR #1014, opened 2017-08-14
Move login and platform status, use button

el-mapache:
Sets the platform status link and login button on their own row. 

**Screenshot**: 

screen shot 2017-08-22 at 9 26 31 am

### DIFF
--- a/layouts/partials/header.html
+++ b/layouts/partials/header.html
@@ -8,9 +8,17 @@
 </div>
 <header class="header">
   <div class="header-wrap">
+    <ul class="nav mt0 mb6">
+      <li class="nav-link row-center-y">
+        <a href="https://cloudgov.statuspage.io/" class="external-link" style="align-self: center">Platform Status</a>
+      </li>
+      <li class="nav-link">
+        <a href="https://login.fr.cloud.gov/login" class="action action-outline inline-block">Log in</a>
+      </li>
+    </ul>
 
     <!-- start logo -->
-    <h2 class="header-title">
+    <h2 class="header-title my0">
       <a href="/" class="logo" title="Home">
         <svg class="logo-img">
           <use xlink:href="/assets/img/cloudgov-sprite.svg#logo"/>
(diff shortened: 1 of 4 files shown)


layouts/partials/header.html
```diff
@@ -8,9 +8,17 @@
 </div>
 <header class="header">
   <div class="header-wrap">
+    <ul class="nav mt0 mb6">
+      <li class="nav-link row-center-y">
+        <a href="https://cloudgov.statuspage.io/" style="align-self: center">Platform Status</a>
+      </li>
+      <li class="nav-link">
+        <a href="https://login.fr.cloud.gov/login" class="action action-outline inline-block">Log in</a>
```
el-mapache: Does this link need to be dynamic? I'm not sure if we have different environments for staging, etc.

jcscottiii: this is fine.

jcscottiii:
I think if we move the login button above, we should get rid of the old one so we don't have two in the header.

jcscottiii:
also adding @brittag @femmebot

el-mapache:
@jcscottiii Oh yes definitely. I thought I moved both that and the status link.

el-mapache:
Sounds good!

mogul:
I'm loving the idea that the status and login links will appear in the same place whether you're looking at docs or looking at the dashboard... Is that where this is heading?

el-mapache:
@mogul Actually I'm not sure if that was intended, but that's a great point! We might want to think about making those headers match in a future PR.

el-mapache:
@femmebot thanks! Getting the `cg-style` update in first works for me!

layouts/partials/header.html
```diff
@@ -8,9 +8,17 @@
 </div>
 <header class="header">
   <div class="header-wrap">
+    <ul class="nav mt0 mb6">
+      <li class="nav-link row-center-y">
+        <a href="https://cloudgov.statuspage.io/" style="align-self: center">Platform Status</a>
```
femmebot: Note: need to add class `.external-link` to include the icon here

el-mapache:
@femmebot @jcscottiii This ticket has been updated and is good to go!

suprenant:
Decided from our standup followup: let's remove the "How to log in" button and keep the "Log in" button.

el-mapache:
@suprenant @femmebot updated!

femmebot (approved):
LGTM! Thanks for removing the "How to log in" link. 

Is there a way to preview the login page with the CLI blurb or is that a separate task?

el-mapache:
@femmebot There is, but you have to pull down a few repos and build/run them locally.

PR #1015, opened 2017-08-14
Redis docs for HA plans

rogeruiz:
This patch also includes some changes to the docs around experimental services and collapses Redis into a single page rather than a versioned page.


brittag:
Yay! Pretending to be a reader, I have a few questions (I don't need answers in comments, just thoughts to consider for improving the docs page if they make sense):

* `redis28` still says "downtime and data loss is possible", the same as the current Redis service, but this version is marked Beta and the old page says Alpha. Is that plan upgraded in some way? Should the description explain what's improved?
* Is there any reason why I should use `redis28` still, or should I use `redis32` for all new work and migrate old work to it when possible? (If we're only keeping `redis28` for backwards-compatibility, should we explain that in some way so that people don't keep using it?)
* If I'm building an application, are there any major differences in using `redis28` and `redis32` that I should know about? If there are any major differences, where should I look in the Redis docs for a description of them? (I don't think we need to do a ton of work on re-explaining things that are already in Redis docs, just seems helpful to give a link to a list of breaking changes if one is available.)
* For `redis32`, "connections may be dropped during platform updates or server failover", what does this really mean for me as an application developer? Will I see brief downtime and then automatic recovery? Also I might be confused as a reader since the intro to the page told me there's a HA option and then I see a note saying it might not be HA.

cnelson (dismissed):
LGTM.  Not merging only so that we can push this when we enable the service

jmcarp (approved):

PR #1020, opened 2017-08-15
Create a Tab based UI element for content that has "alternative" solutions

rememberlenny:
Created a new shortcode for `alternatives`. The `alternatives` uses a custom build of the jquery-ui package, with only the `alternatives` functionality. The `alternatives` shortcode takes two parameters (`dashboard` and `cli`) for the IDs used to target anchored alternatives. 

The parameters allows the tab buttons to know what tab content to connect.

Note: I need to pull out the CSS into cg-style, but Im not quite sure what the protocol is around the jquery-ui based CSS. Any tips @msecret ?


jcscottiii:
i'm kinda leaning towards having them both visible at the same time like we did for east-west vs govcloud https://github.com/18F/cg-site/pull/544

also, marco had a similar comment about it here: https://github.com/18F/cg-site/pull/544#issuecomment-259828661

regarding the css, we can leave it here for now. this styling of switching wouldn't be put in cg-style (we didn't do it for the east-west vs govcloud distinction either).

rememberlenny:
@jcscottiii  Im for it either way! 

The one consideration is that the page is getting really long and difficult to grok.

@mogul mentioned this approach and it made sense as I added the Space related changes.

Thoughts @brittag ?

jcscottiii:
i'd be interested to know if our pages are running long do we need to break them up into smaller pages?

rememberlenny:
@jcscottiii I think in most cases we should. This page is tricky because most of the content is being repeated 4 times. As a result, the only variation is in the word Organization/Space or in the CLI/Dashboard instructions.

brittag:
Sounds like we should look at refactoring the page to be less repetitive!

rememberlenny:
That's definitely one solution. I appreciate when I use documentation and I don't need to read the entire page for context. It does make sense that things are repetitive from the perspective that a user can scroll to the place they need to see, and find heir answer.

mogul:
The root of the problem is that often our docs need to describe two methods for accomplishing an action... Via the CLI, or via the Dashboard. Users are usually interested in one or the other. For example, project managers are almost certainly going to use the dashboard to manipulate user roles, while developers are almost certain to be using the CLI. Since we're listing both methods in sequence it's hard to pick out which is which, and scan past the part that's not relevant to you, particularly if there's more than one section on a page with such alternatives.

I made the suggestion that we could follow the example of others, who differentiate in their docs based on the client type. For example Google:
https://support.google.com/photos/answer/6128838?hl=en&ref_topic=6128856&co=GENIE.Platform%3DDesktop&oco=1
image

And another example in Slack:
https://get.slack.help/hc/en-us/articles/204092246-Edit-your-profile
image

This PR is Lenny's attempt at enabling that kind of consolidation by client-type.

mogul:
(The Slack example has multiple sets of tabs in the single page of docs, keeping it clean and short.)

rememberlenny:
Currently, I'm only thinking about the Command line vs Dashboard based options in doing user management related actions.

brittag:
@femmebot Here are a few additional places in the docs where we could/should offer both CLI and dashboard instructions for the same tasks (I don't have a strong opinion on whether tabs are the right solution):

* Listing available services and creating (some) service instances - https://cloud.gov/docs/apps/managed-services/
  * Possibly: creating service instances for https://cloud.gov/docs/services/elasticsearch24/ + https://cloud.gov/docs/services/redis28/ + https://cloud.gov/docs/services/relational-database/ + https://cloud.gov/docs/services/s3/ + https://cloud.gov/docs/services/cloud-gov-service-account/
* Configuring multiple instances - https://cloud.gov/docs/apps/multiple-instances/
* Configuring application limits - https://cloud.gov/docs/apps/limits/
* Viewing recent logs - https://cloud.gov/docs/apps/logs/

jcscottiii (approved):

PR #1030, opened 2017-08-17
Update to replace entire block with just s3

sharms:


content/docs/ops/runbook/troubleshooting-logsearch.md
```diff
@@ -48,15 +48,19 @@ cp /var/vcap/jobs/ingestor_syslog/config/logstash.conf /tmp/logstash-restore.con
...
 Edit the `/tmp/logstash-restore.conf` and make the following changes:
 
-#### Remove the `redis` input and replace it with an `s3` input
+#### Update the "input" block to only contain s3
 ```
+input {
   s3 {
     bucket => ":bucket:"
     region => ":region:"
+    access_key_id => ":access_key_id:"
```
LinuxBozo, line 57: Will it not use instance profiles?

jmcarp: The s3 plugin should understand instance profiles, but the ingestors are currently using the default instance profile and can't read from these buckets. We talked about having the ingestors use an instance profile that grants read-only access to firehose backup buckets, but didn't think it made sense to give those machines permissions they'll only need in the exceptional case of a manual restore. WDYT?

LinuxBozo: Actually, torn. I clearly see value in having an instance profile. A restore is likely going to happen during the course of some event. Depending on stress levels, an operator could possibly leak credentials during this process if forced to inject them here, instead of just relying on the fact that the permissions needed to carry out the operation are already in place. Might be too paranoid a situation though.

jmcarp (approved):

PR #1031, opened 2017-08-17
another text tweak

suprenant:
Added a little tweak ("So,") for flow.


khandelwal:
The sentence is stronger without the 'So'.

PR #1034, opened 2017-08-26
Add useful CF docs to troubleshooting page

brittag:
The page at https://docs.cloudfoundry.org/devguide/deploy-apps/troubleshoot-app-health.html has some good user-centered advice - each section start with a description of the problem and then provides ways to solve it. I added a link to it!

Our STG / DEA tips are organized by stage of deployment - I'm not sure this really works well for user needs, but we can work on that another time. I tried at least revising the titles a little bit.

### DIFF
--- a/content/docs/apps/troubleshooting.md
+++ b/content/docs/apps/troubleshooting.md
@@ -42,9 +42,9 @@ Cloud Foundry will attempt to detect the buildpack to use with your app by exami
 
 If you're seeing errors installing any Python dependencies, check whether you have a `vendor/` directory in your app's root. The Python buildpack won't use PyPI if you have a `vendor/` directory, so you'll need to rename that directory to something else.
 
-## Potential causes of issues during Droplet Execution Phase [DEA]
+## Potential causes of issues during app starting/stopping [CELL]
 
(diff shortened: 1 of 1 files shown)


content/docs/apps/troubleshooting.md
```diff
@@ -42,9 +42,9 @@ Cloud Foundry will attempt to detect the buildpack to use with your app by exami
 
 If you're seeing errors installing any Python dependencies, check whether you have a `vendor/` directory in your app's root. The Python buildpack won't use PyPI if you have a `vendor/` directory, so you'll need to rename that directory to something else.
 
-## [DEA] Droplet Execution Phase
+## Potential causes of issues during Droplet Execution Phase [DEA]
```
jmcarp: Any docs that mention the droplet execution agent (dea) are outdated, since we haven't used that backend since the switch to govcloud (and with it diego). We don't have dea logs anymore--the current set of log types is at https://docs.cloudfoundry.org/devguide/deploy-apps/streaming-logs.html#format. I'm thinking we can reference `[CELL]` logs instead of `[DEA]` logs here.

brittag:
@jmcarp Fixed the problem - up for reviewing again?

jmcarp (approved):

PR #1035, opened 2017-08-28
Add header to ops docs.

### LINKED ISSUE #893: Distinguish operations docs from general user docs: Troubleshooting SMTP edition
**User stories:**
* As a cloud.gov user, I want to be able to follow the docs' instructions, so that I can accomplish what I came to do without accidentally following false paths.
* As a cloud.gov team member doing support, I want a reliable set of instructions that are reliably applicable to users, so that I can give users accurate guidance.

**Context:**
There's a larger question about whether operations docs should be separate from the general user docs, but it'd take a megaticket to execute on that. :) We can start small, by clearly stating when something is 18F-specific.

Prompted by the references to #infrastructure, which outsiders can't access, and Mandrill, which we're not currently recommending, on the [Troubleshooting SMTP](https://github.com/18F/cg-site/blob/f996eb07a430c082cf618ecdc2c65a203182f156/content/docs/ops/runbook/troubleshooting-smtp.md) page.

jmcarp:
Note: Hugo doesn't support layouts for nested sections yet, but once it
does, we can drop the `layout` stanza for each `ops` page:
https://discourse.gohugo.io/t/layout-for-nested-section/7292.

[Resolves #893]

### DIFF
--- a/content/docs/ops/runbook/rotating-bosh.md
+++ b/content/docs/ops/runbook/rotating-bosh.md
@@ -2,6 +2,7 @@
 menu:
   docs:
     parent: runbook
+layout: ops
 
 title: Rotating Secrets III - BOSH
 ---
--- a/layouts/docs/ops.html
+++ b/layouts/docs/ops.html
@@ -1,9 +1,11 @@
 {{ partial "docs-header.html" . }}
 
-<h3>
-  Note: these docs are primarily intended for cloud.gov operators; check out
-  our <a href="/docs">user documentation</a> for help with using the platform.
-</h3>
+<p>
+<em>
+  This page is primarily for the cloud.gov team, and it's public so that you can learn 
+  from it. For help with using cloud.gov, see the <a href="/docs/">user docs</a>.
+</em>
+</p>
 
 {{ .Content }}
 
--- a/layouts/docs/ops.html
+++ b/layouts/docs/ops.html
@@ -2,8 +2,8 @@
 
 <p>
 <em>
-  This page is primarily for the cloud.gov team, and it's public so that you can learn 
-  from it. For help with using cloud.gov, see the <a href="/docs/">user docs</a>.
+  This page is primarily for the cloud.gov team. It's public so that you can learn 
+  from it. For help using cloud.gov, see the <a href="/docs/">user docs</a>.
 </em>
 </p>
 
--- a/layouts/docs/ops.html
+++ b/layouts/docs/ops.html
@@ -1,0 +1,12 @@
+{{ partial "docs-header.html" . }}
+
+<h3>
+  Note: these docs are primarily intended for cloud.gov operators; check out
+  our <a href="/docs">user documentation</a> for help with using the platform.
+</h3>
+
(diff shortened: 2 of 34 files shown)


brittag:
Oh nice, this is a great way to make this happen - thanks @jmcarp! I updated the template a bit. @fureigh would you be up for reviewing this to see if it delivers what you had in mind?

content/docs/ops/runbook/rotating-bosh.md
```diff
@@ -2,6 +2,7 @@
 menu:
   docs:
     parent: runbook
+layout: ops
```
fureigh, line 6: From this file through `content/docs/ops/runbook/troubleshooting-snort.md`, there's an extra linebreak between the `layout` line and the `title` line. It makes no functional difference, but I mention it for style consistency.

layouts/docs/ops.html
```diff
@@ -0,0 +1,14 @@
+{{ partial "docs-header.html" . }}
+
+<p>
+<em>
+  This page is primarily for the cloud.gov team, and it's public so that you can learn
```
fureigh: I suggest changing `, and it's` to `. It's` — flows better, to my ear.

layouts/docs/ops.html
```diff
@@ -0,0 +1,14 @@
+{{ partial "docs-header.html" . }}
+
+<p>
+<em>
+  This page is primarily for the cloud.gov team, and it's public so that you can learn 
+  from it. For help with using cloud.gov, see the <a href="/docs/">user docs</a>.
```
fureigh: I suggest removing `with` for the sake of brevity. Also not a big deal.

layouts/docs/ops.html
```diff
@@ -0,0 +1,14 @@
+{{ partial "docs-header.html" . }}
+
+<p>
+<em>
```
fureigh, line 4: Suggest indenting as in `layouts/docs/services.html`.

fureigh (approved):
Thanks for doing this, @jmcarp, and for looping me in, @brittag! No blockers, though I made a few suggestions. If you'd like me to go ahead and implement them, feel free to tap me.

fureigh:
@jmcarp I did notice one issue: the vertical spacing between this new header and the start of the content is inconsistent throughout the pages.

For example, compare `/docs/ops/dns/`:

screen shot 2017-08-31 at 12 06 02 pm

to `/docs/ops/runbook/rotating-bosh/`:

screen shot 2017-08-31 at 12 05 37 pm

Haven't yet determined whether that's due to `layouts/docs/ops.html`, but the issue isn't currently replicable on production.

fureigh (dismissed):
(See more recent note.)

jmcarp:
@fureigh: I'm not seeing the issue with whitespace. It looks like there's a bigger margin between the header and content on some pages because headings have a bigger top margin than paragraphs, and the main content starts with a heading on some pages and a paragraph on others.

brittag (approved):

PR #1040, opened 2017-08-31
Add the word `encrypt` for search purposes

fureigh:
We had `encryption` but searches for `encrypt` weren't getting any hits. This solves that.

Addresses https://gsa-tts.slack.com/archives/C09CR1Q9Z/p1504192405000181.

### DIFF
--- a/content/docs/services/relational-database.md
+++ b/content/docs/services/relational-database.md
@@ -146,7 +146,7 @@ $ pg_restore --clean --no-owner --no-acl --dbname={database name} backup.pg
 
 ## Encryption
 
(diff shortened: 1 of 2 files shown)


brittag:
How about we cross-link to the assessor details from the docs page instead of repeating the info? I'm ok with this if you want this as-is, just an option if it seems helpful.

brittag (approved):

fureigh:
@brittag Usually I'm in favor of deduplication, but if the purpose of this ticket is to point people in the right direction when they're looking for information about cloud.gov's encryption, I think it behooves us to include the word `encrypt` in both places. (The info is all in the link that's already provided, but that doesn't come through in search results, of course.)

fureigh:
@brittag Thank you for the suggestion, though!

PR #1046, opened 2017-09-06
Update the US Federal Gov site notice on cg-site

rememberlenny:
open-close


el-mapache (dismissed):
One nitpick but looks fine otherwise.

static/assets/js/index.js
```diff
@@ -9,12 +12,37 @@
...
+      }
+    });
+    $govBarWrapper.on('click', function(){
+      toggleGovBar();
+    })
     ajaxifyContactForm();
   });
 
+  function toggleGovBar() {
+    if ($('#gov-banner').hasClass('hide')) {
```
el-mapache: You can use the `$govBarContent` variable here I think.

rememberlenny: Updated.

el-mapache (approved):
all the thumbs up

PR #1051, opened 2017-09-12
Change "fewer" to "less"

femmebot:
Less is used with numbers when they are on their own and with expressions of measurement or time. https://en.oxforddictionaries.com/usage/less-or-fewer


khandelwal (approved):
Looks good to me.

PR #1057, opened 2017-09-13
Add info to SSH page about disabling SSH

brittag:
Changes:
* Note that SSH is enabled by default and how to disable it.
* ~Mention the alternate SSH ports available ([cloud.gov team: as documented here](https://docs.google.com/document/d/1-yWK83jESYB-hUQVe_J65jR65UOy0MImDh5GkX3uHlk/edit)). I don't know if this description accurately represents the user experience - does it automatically work like this, or does the user need to do something special to use the alternate ports?~

This needs technical review.

### DIFF
--- a/content/docs/apps/using-ssh.md
+++ b/content/docs/apps/using-ssh.md
@@ -21,7 +21,7 @@ You can interact directly with the services bound to your application via [port-
 
 ### Error messages
 
-* `cf ssh` uses port 2222, 22, or 443. If your network blocks all of these ports, you may receive an error message such as `Error opening SSH connection`.
+* `cf ssh` uses port 2222. If your network blocks port 2222, you may receive an error message such as `Error opening SSH connection`.
 * Running `python` within an SSH session results in `ImportError: No module named site`.  
   * To resolve this, configure your SSH session environment with these commands:
     ```
--- a/content/docs/apps/using-ssh.md
+++ b/content/docs/apps/using-ssh.md
@@ -32,6 +32,10 @@ You can interact directly with the services bound to your application via [port-
     export PYTHONHOME=/home/vcap/deps/0/python
     ```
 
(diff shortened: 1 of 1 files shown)


content/docs/apps/using-ssh.md
```diff
@@ -21,7 +21,7 @@ You can interact directly with the services bound to your application via [port-
 
 ### Error messages
 
-* `cf ssh` uses port 2222. If your network blocks port 2222, you may receive an error message such as `Error opening SSH connection`.
+* `cf ssh` uses port 2222, 22, or 443. If your network blocks all of these ports, you may receive an error message such as `Error opening SSH connection`.
```
mogul: By default `cf ssh` uses 2222, and there's no way to change that when running `cf ssh`. If that port's blocked you can use a different SSH client to connect and specify one of the alternate ports; see where `-p` appears [in this example](https://docs.cloudfoundry.org/devguide/deploy-apps/ssh-apps.html#other-ssh-access). I'm on my phone so I can't check if the other two ports are advertised when you `cf curl /v2/info`; if they're not, we should explicitly document that they're supported and point to the fact that you can use them with `-p` (or the equivalent in other SSH clients).

mogul (changes requested):

content/docs/apps/using-ssh.md
brittag: Running `cf curl /v2/info` only advertises `"app_ssh_endpoint": "ssh.fr.cloud.gov:2222"`.

We can wait on the [review of this doc](https://docs.google.com/document/d/1-yWK83jESYB-hUQVe_J65jR65UOy0MImDh5GkX3uHlk/edit) before documenting how to use these ports, in case the best next step is to roll back the availability.

mogul (approved):

PR #1061, opened 2017-09-15
Be explicit about responsibility for fed requirements and bad DNSSEC config causing problems

mogul:


brittag:
Thanks! This helps address the issue that [this customer ran into](https://gsa-tts.slack.com/archives/C039MHHF8/p1503588400000464?thread_ts=1503540084.000049&cid=C039MHHF8).

The underlying bug causing creation to get stuck in this case is filed at https://github.com/18F/cf-cdn-service-broker/issues/91 - so when we fix that, we should also update these docs.

brittag (approved):
This addresses customer questions!

PR #1064, opened 2017-09-18
Document lookup-route for troubleshooting

pburkholder:


brittag:
We should very clearly note this limitation (from the readme): "Note this will only show applications in orgs/spaces that the logged-in user has permissions to view."

brittag:
@pburkholder I tried something a little different - moving it over to the [custom domains page](https://cloud.gov/docs/apps/custom-domains/), since we already talk about domains and routes there. What do you think?

rogeruiz (approved):
:shipit:

PR #1077, opened 2017-09-22
Swim with the Mermaids

rogeruiz:
:fish:

Docs: https://github.com/18F/cg-site/blob/27302cc2abc05ae377ab7dd2a3421cb15b2542aa/README.md#add-diagrams-to-documentation-pages

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,9 +26,11 @@
   "homepage": "https://github.com/18F/cg-site#readme",
   "devDependencies": {
     "browserify": "^13.1.0",
+    "mermaid": "^7.1.0",
     "node-sass": "^3.10.0",
     "onchange": "git+https://git@github.com/msecret/onchange.git",
-    "postcss-import": "^8.1.2"
+    "postcss-import": "^8.1.2",
+    "scuttle": "^1.0.0"
   },
   "dependencies": {
     "anchor-js": "^3.2.1",
--- a/package.json
+++ b/package.json
@@ -29,8 +29,7 @@
     "mermaid": "^7.1.0",
     "node-sass": "^3.10.0",
     "onchange": "git+https://git@github.com/msecret/onchange.git",
-    "postcss-import": "^8.1.2",
-    "scuttle": "^1.0.0"
+    "postcss-import": "^8.1.2"
   },
   "dependencies": {
     "anchor-js": "^3.2.1",
--- a/static_src/js/.eslintrc
+++ b/static_src/js/.eslintrc
@@ -1,0 +1,18 @@
+{
+  "rules": {
+    "indent": [ 0, 2 ],
+    "quotes": [ 0, "single" ],
+    "linebreak-style": [ 2, "unix" ],
+    "no-multi-spaces": [ 0 ],
+    "no-mixed-requires": [ 2, true ]
+  },
+  "env": {
+    "es6": true,
+    "node": true,
+    "browser": true
+  },
+  "extends": "eslint:recommended",
+  "parserOptions": {
+    "sourceType": "module"
+  }
+}
--- a/static_src/js/main.js
+++ b/static_src/js/main.js
@@ -1,5 +1,5 @@
 var $ = require( 'jquery' );
-window.jQuery = $;
+var jQuery = window.jQuery = $;
 window.$ = $;
 require('cloudgov-style');
 
@@ -267,7 +267,25 @@ function initializeJS() {
 	})();
 })(jQuery);
 
+var swimWithTheMermaids = function() {
+
+  // Are there any diagrams here? if not skip.
+  if ( 1 > $( '#js-stage' ).length ) {
+    return undefined;
+  }
+
+  mermaid.mermaidAPI.render(
+    'js-stage',
+    ss.textContent.replace( /\n/g, '\\n' ),
+    function callbackMermaidRender( c ) {
+      ss.innerHTML = c;
+    }
+  );
+
+}
+
 jQuery(document).ready(function(){
     initializeJS();
     $('[rel=show-github]').showGithub();
+    swimWithTheMermaids();
 });
--- a/static_src/js/main.js
+++ b/static_src/js/main.js
@@ -1,7 +1,12 @@
 var $ = require( 'jquery' );
+var mermaid = require( 'mermaid' );
 var jQuery = window.jQuery = $;
 window.$ = $;
 require('cloudgov-style');
+
+mermaid.initialize( {
+  startOnLoad: false,
+} );
 
 var anchorJS = require('anchor-js');
 require('jquery.scrollto');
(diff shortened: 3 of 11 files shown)


static_src/js/main.js
```diff
@@ -267,7 +267,25 @@ function initializeJS() {
...
 })(jQuery);
 
+var swimWithTheMermaids = function() {
+
+  // Are there any diagrams here? if not skip.
+  if ( 1 > $( '#js-stage' ).length ) {
+    return undefined;
+  }
+
+  mermaid.mermaidAPI.render(
```
jmcarp: Does this assume there's either zero diagrams or one? Since we're selecting by (hopefully unique) id, would this work with multiple diagrams?

rogeruiz: yeah, I think I'm going to have to pass in nodes rather than a string to the render function, for now I'm just copying over Scuttle, but I'm not sure how I'm gonna get the styles over. Probably going to need to publish it to get it over in `cg-style` as a `npm install` 🤷‍♂️

rogeruiz:
To accept this, change to this branch locally then add a diagram shortcode with the diagram as the body. You can copy paste the example below. The `id-prefix` is also used to title the diagram, so you'll see it render like the image below.

```html
{{< diagrams id-prefix="sample-diagram" >}}
graph LR
    A[Square Rect] -- Link text --> B((Circle))
    A --> C(Round Rect)
    B --> D{Rhombus}
    C --> D
{{< /diagrams >}}
```

screen shot 2017-09-27 at 11 11 54 pm

package.json
```diff
@@ -26,9 +26,11 @@
   "homepage": "https://github.com/18F/cg-site#readme",
   "devDependencies": {
     "browserify": "^13.1.0",
+    "mermaid": "^7.1.0",
     "node-sass": "^3.10.0",
     "onchange": "git+https://git@github.com/msecret/onchange.git",
-    "postcss-import": "^8.1.2"
+    "postcss-import": "^8.1.2",
```
jmcarp: I see that we're importing `mermaid`, but where do we use `postcss-import` or `scuttle`?

static_src/js/main.js
```diff
@@ -267,7 +272,37 @@ function initializeJS() {
...
+    var $el = $( el );
+    var diagramText = $( el ).find( '.js-diagrams__raw' ).text()
+    mermaid.render(
+      $el.find( '.js-diagrams__stage' ).attr( 'id' ),
+      diagramText,
+      function callbackMermaidRender( c ) {
+        $el.find( '.js-diagrams__stage' ).html( c );
+      }
+    );
+    $el.find( '.js-diagrams__stage' ).find( 'style' ).remove();
```
jmcarp, line 295: Could we do this with css instead of js?

static_src/js/main.js
```diff
@@ -267,7 +272,37 @@ function initializeJS() {
...
+    $el.find( '.js-diagrams__stage svg' ).height( updatedHeight );
+    $el.find( '.js-diagrams__stage' ).width( updatedWidth );
+  } );
+
+}
+
 jQuery(document).ready(function(){
     initializeJS();
     $('[rel=show-github]').showGithub();
+    swimWithTheMermaids();
```
jmcarp: Sorry to be un-fun, but wdyt about calling this something more descriptive like `renderDiagrams`?

static_src/js/.eslintrc
```diff
@@ -0,0 +1,18 @@
+{
```
jmcarp, line 1: I'm all in favor of checking in some linting rules, but if we do, we should probably run `eslint` in ci.

static_src/js/main.js
rogeruiz: it's coming [from the mermaid library here](https://github.com/knsv/mermaid/blob/d6e22d5f5e969d78d6b2e9ab85aaa62119a18b84/src/mermaidAPI.js#L437-L448). so we've got to remove it after the fact for now.

jmcarp:
This looks great! My only non-style comment is that I'd like to see this patch include at least one diagram. I don't want to say we've shipped this feature until we're actually using it 😄 

cc @mogul

package.json
rogeruiz: so I believe, that `postcss-import` was already there. I'm hoping to publish Scuttle `1.1.0` which will include the compiled CSS in the package so we can copy it out the same way that `cg-style` is copied into the project from `node_modules`. We can remove it for now though, it's not being used at the moment.

rogeruiz:
Yeah, I wasn't sure what we should include as the first diagram. But I'm sure someone else will have better ideas there 🏆

package.json
rogeruiz: in other words, I'm not really sure where we use `postcss-import` 😁

It might be good to clean things up, but I don't want to turn this into a huge amount of work to refactor my way out of this frontend paper bag.

static_src/js/.eslintrc
rogeruiz: ah dang, fat-fingered this commit. do you think I should remove it or add the CI tests? I'm up for either. but if I add lint file, I'll want to update this one so it's... more opinionated. 🤷‍♂️

mogul (commented):
I think this is great, but I'm unqualified to opine as to the technical implementation.

mogul:
Any chance we can get this merged in time to have a slide on it in the quarterly update slides on Thursday afternoon?

mogul:
Also can you add [this information](https://github.com/18F/cg-site/pull/1077#issuecomment-332722815) to the README.md so people know that this is possible and how to do it?

rogeruiz:
@mogul sure, I'm asking for your review of what it looks like in that `demo` URL above, and if you could help with this part possibly.

> "My only non-style comment is that I'd like to see this patch include at least one diagram. I don't want to say we've shipped this feature until we're actually using it" ~ @jmcarp 

Is there a flowchart we want to include somewhere on the site? of course it can also come as another PR from someone else too. 🤷‍♂️

brittag:
@rogeruiz I'd love for the Not Great diagram at https://cloud.gov/overview/technology/responsibilities/ to be a better diagram! Not sure if it's suitable for conversion.

rogeruiz:
I updated the Customer Responsibility page with new diagrams for each application architecture type. The demo URL has been updated as well.

package.json
jmcarp: OK, seems like we can leave `scuttle` out until installing it actually does something.

jmcarp (approved):

brittag:
@rogeruiz @jmcarp Ah sorry, I put this in Favro yesterday but not here - I think the responsibilities page diagrams need more work before we update them to use this. I think the horizontal flowchart makes it harder to compare the three categories. Can you flip that part back?

PR #1078, opened 2017-09-23
Add basic info on RDS page about getting database version info

brittag:
Prompted by [user question here](https://gsa-tts.slack.com/archives/C09CR1Q9Z/p1506111900000018). This PR should probably be fleshed out with MySQL and Oracle info too - I'll leave that to somebody else (since I'm not sure how to get it).

Also stuck in `sh` to make some syntax highlighting a little less weird.

### DIFF
--- a/content/docs/services/relational-database.md
+++ b/content/docs/services/relational-database.md
@@ -12,19 +12,19 @@ If your application uses relational databases for storage, you can use the AWS R
 
 ## Plans
 
-Plan Name | Description | Price
---------- | ----------- | -----
-`shared-psql`            | Shared PostgreSQL database for prototyping (no sensitive or production data) | Free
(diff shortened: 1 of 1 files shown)


brittag:
Putting this in the service plan table is great, thank you @jmcarp! I tried updating the "Version" column to say "Software Version" (to distinguish from people potentially interpreting it as service version) - not sure if that succeeds in the goal, but might help.

mogul (approved):

brittag:
Th `cf ssh` info was sort of incomplete - to successfully inspect the database version info, you have to do a somewhat complex set of steps beyond just SSHing - so I took that out for now. We can write that back in if the version table doesn't turn out to be good enough (if people really want to inspect it themselves).

PR #1079, opened 2017-09-28
update code block in ops docs

wjwoodson:


content/docs/ops/managing-users.md
```diff
@@ -36,44 +36,44 @@ cf services
...
+uaadb=> begin;
+   BEGIN
+uaadb=> delete from totp_seed where username='pat.jones@agency.gov';
+   DELETE 1
+uaadb=> commit;
+   COMMIT
+uaadb=>
+```
 
 4. Let the user know the reset process is complete, so they can set up a new authentication application and request access from Space Managers and Org Managers again.
```
rogeruiz, line 78: could we change these from `1, 2, 3, 4` sequential numbering to `1, 1, 1, 1` instead. Markdown will take care of numbering them appropriately for us and it'll help in the maintenance in case we ever need to rearrange these.

rogeruiz (commented):
looks good @wjwoodson

PR #1080, opened 2017-09-28
Document ephemeral sandbox policy.

jmcarp:
Document notification and purging behavior implemented in https://github.com/18F/cg-sandbox/pull/5.

Note: I'm guessing we'll want to make this notice more prominent so that users don't miss it.


mogul:
@khandelwal is saying he thinks this should be 90 days rather than 30, btw.

mogul (dismissed):
Looks good to me. We should probably merge this now even though the actual clock hasn't started so that people creating sandboxes will know about the policy up front. What do you think @brittag ?

brittag:
@mogul I don't think we should merge this exact change until we actually implement the policy, since I don't know how long it'll be before we can ship this feature (since it has an external dependency on the TRs). We could ship it with a "Note: this is coming soon but not happening yet" thing.

gboone (dismissed):

mogul (changes requested):
I'll approve and merge this as soon as it's brought up to date with master.

jmcarp:
Rebased.

mogul (dismissed):

mogul:
Looks like maybe it needs to be rebased again...?

mogul (approved):

PR #1082, opened 2017-10-01
Iaas sidecar docs mvf

mogul:
This is just enough to give us something to point people to when we talk about the IaaS Sidecar functionality. 

The provisioning of accounts is still largely hand-cranked behind the scenes for now... We'll automate it via a broker, and sequence that work to happen before this becomes an overwhelming headache in support.

We won't process any requests for this service before the SCR is approved.


jmcarp:
What do you think about listing some sample use cases? It might not be clear from this why somebody would want to use a separate govcloud account.

mogul:
@jmcarp Sure... Do you have in mind a way to convey that?

PR #1083, opened 2017-10-03
Swapped out workshop info with latest update

femmebot:
screencapture-localhost-1313-1506991591146


brittag:
@femmebot Here's one way to do this in an automatic way with Hugo templates instead of hardcoding it: https://github.com/18F/cg-site/compare/latest-post?expand=1 - feel free to adapt if helpful. The result looks like this:

screen shot 2017-10-02 at 6 07 05 pm

brittag (approved):
works for me!

PR #1084, opened 2017-10-03
Add another example of a SSH error message

brittag:
See also: https://github.com/18F/cf-service-connect/pull/40


el-mapache (approved):
Looks great!

PR #1087, opened 2017-10-05
Hints on CDN route

pburkholder:


content/docs/ops/services.md
```diff
@@ -0,0 +1,13 @@
...
+menu:
+  docs:
+    parent: tenants
+layout: ops
+title: Managing services
+---
+
+## Reset pending `cdn-route` service creation
+
+The `create-service` command for the cdn-route broker will be in a pending state until either a) the tenant completes the verification of domain ownership or b) one week elapses. If someone has mistyped a domain, they won't want to wait a week to try again.
```
el-mapache: Do we need to explain what `tenant` means in this case, or do you expect that to be clear to someone who is reading the docs?

brittag (approved):

PR #1089, opened 2017-10-06
Revert "Update diagram to use Scuttle"

jmcarp:
This reverts commit a702d9caf83c4ad221d5f1dd7cecb1005a3c16cb.

Sorry @brittag, I thought the mermaid diagrams had been reverted.

cc @rogeruiz


mogul:
Doesn't the partial need to stay in there for diagrams to work on other pages?

rogeruiz (approved):
:shipit:

My original work was to only bring diagrams to the docs pages. We’ll need to update the partial to bring this to the overview pages. But we can do that in the future.

PR #1090, opened 2017-10-06
P-ATO Edits

gboone:
Includes edits to the FedRAMP Authorized page to clarify what the P-ATO is and how agencies can use it. Also includes the chart showing how many controls are handled by cloud.gov for a Moderate level application.


content/overview/security/fedramp-tracker.md
```diff
@@ -12,30 +12,24 @@ aliases:
...
 
-We plan to publish much of the documentation from our P-ATO package, as part of our open source system documentation, after we add context for public release.
+If you want to use cloud.gov, [**request the P-ATO documentation package from FedRAMP**](https://s3.amazonaws.com/sitesusa/wp-content/uploads/sites/482/2015/03/FedRAMP-Package-Request-Form_V4_06192014.pdf) (the Package ID for that form is F1607067912). You can also view the [FedRAMP Marketplace page for cloud.gov](https://marketplace.fedramp.gov/#/product/18f-cloudgov?sort=productName).
 
-As a government team with a mission to support agency efforts to improve the way they deliver services to the public, we want our compliance documentation to be available as a model that helps additional Platform as a Service providers (including commercial providers) write the documentation they need to achieve FedRAMP JAB P-ATO as well.
+The [**Control Implementation Summary + Customer Responsibility Matrix (.xlsx)**](/resources/cloud.gov-CIS-Worksheet.xlsx) is a quick summary of each Low and Moderate security control and whether it is handled by cloud.gov, shared responsibility, or customer responsibility.
```
el-mapache, line 38: This link returns a 404.

el-mapache (changes requested):
Reads fine!

content/overview/security/fedramp-tracker.md
```diff
@@ -12,30 +12,24 @@ aliases:
...
-
-The majority of federal systems are at the [Low and Moderate impact levels](http://csrc.nist.gov/publications/fips/fips199/FIPS-PUB-199-final.pdf#page=6), which can be hosted on cloud.gov. [cloud.gov can’t yet host High impact systems.]({{< relref "overview/security/fisma-high-applications.md" >}})
+FedRAMP is like an outfitter for cloud services. Your agency still needs to grant the system you want to build an Authority to Operate but FedRAMP has done the labor intensive work of reviewing cloud.gov's security posture and endorsed it. Your agency's authorizing official can request the P-ATO documentation package from FedRAMP and accept that endorsement for your own system.
 
-FedRAMP requires cloud.gov to maintain continuous monitoring and undergo annual re-assessment to retain the P-ATO, which gives your agency ongoing assurance that cloud.gov is compliant.
+![Graph showing the breakdown of how many controls are fully covered by cloud.gov.](/img/fedramp-moderate-controls.png)
```
el-mapache: Leads to a 404

content/overview/security/fedramp-tracker.md
gboone: Hmmm, I wonder if it won't return a 404 on cloud.gov? I copied the same relative URL from the current version…

gboone (commented):
I'm thinking the image URL won't return a 404 on cloud.gov. I followed the template in the README which says to use `alt text`

content/overview/security/fedramp-tracker.md
brittag: This image and the spreadsheet download work fine when I run the site locally:

screen shot 2017-10-06 at 7 07 27 pm

content/overview/security/fedramp-tracker.md
```diff
@@ -12,30 +12,24 @@ aliases:
 ---
 
 
-cloud.gov has a [Provisional Authority to Operate (P-ATO) at the Moderate impact level from the FedRAMP Joint Authorization Board (JAB)](https://marketplace.fedramp.gov/#/product/18f-cloudgov).
+cloud.gov has a [Provisional Authority to Operate (P-ATO) at the Moderate impact level from the FedRAMP Joint Authorization Board (JAB)](https://marketplace.fedramp.gov/#/product/18f-cloudgov). This means cloud.gov has undergone a significant, thorough security and compliance review so that your agency doesn't have to.
```
brittag: "Doesn't have to" seems to end a bit abruptly. Maybe "This means cloud.gov has undergone a significant, thorough security and compliance review, so that your agency can use that review instead of conducting its own." ? Not quite right either, but I think it needs something to specify the benefit.

content/overview/security/fedramp-tracker.md
```diff
@@ -12,30 +12,24 @@ aliases:
...
+The **[Federal Risk and Authorization Management Program (FedRAMP)](https://www.fedramp.gov/)** evaluates cloud services and issues a **Provisional Authority to Operate (P-ATO)** to those that pass review. Those come in two flavors: Agency and JAB. Both authorizations [look at a standardized set of requirements in accordance with FISMA](https://www.fedramp.gov/jab-or-agency-how-do-i-get-a-fedramp-ato/) but JAB P-ATOs approve cloud services for governmentwide operation. The **Joint Authorization Board (JAB)** is the panel of the CIOs of the General Services Administration, Department of Defense, and Department of Homeland Security that issues [these kinds of authorizations](https://marketplace.fedramp.gov/#/products?status=Compliant&sort=productName&authorizationType=JAB). **Authorization at the moderate level** means cloud.gov is a vetted and trustable service for any government system that contains data where [the impact of loss is limited or serious — but not catastrophic](http://csrc.nist.gov/publications/fips/fips199/FIPS-PUB-199-final.pdf#page=6).
```
brittag: Agency P-ATOs also enable use of cloud services in other agencies, not just JAB P-ATOs. We should make sure our descriptions of FedRAMP are in line with their messaging. A helpful thing about the previous language is that FedRAMP already checked it as part of the post at https://18f.gsa.gov/2017/02/02/cloud-gov-is-now-fedramp-authorized/ - we can ask FedRAMP to take a look again though.

content/overview/security/fedramp-tracker.md
```diff
@@ -12,30 +12,24 @@ aliases:
...
+FedRAMP is like an outfitter for cloud services. Your agency still needs to grant the system you want to build an Authority to Operate but FedRAMP has done the labor intensive work of reviewing cloud.gov's security posture and endorsed it. Your agency's authorizing official can request the P-ATO documentation package from FedRAMP and accept that endorsement for your own system.
 
-FedRAMP requires cloud.gov to maintain continuous monitoring and undergo annual re-assessment to retain the P-ATO, which gives your agency ongoing assurance that cloud.gov is compliant.
+![Graph showing the breakdown of how many controls are fully covered by cloud.gov.](/img/fedramp-moderate-controls.png)
 
-## Sharing our work for reuse
+Here's how it works: Every "moderate" impact federal system needs to account for 325 required controls before it can be granted an ATO. Once your agency reviews and accepts cloud.gov's P-ATO, 269 of these requirements are already implemented and documented. 41 of them are shared between cloud.gov and your application, and only 15 are fully your responsibility.
```
brittag: The FedRAMP Moderate baseline has 325 controls but agencies can choose their own Moderate baselines (so they can have varying numbers)

content/overview/security/fedramp-tracker.md
gboone: > so that your agency can focus on reviewing the parts of the system that serve your mission.

content/overview/security/fedramp-tracker.md
gboone: I was trying to adapt the language on [FedRAMP's page](https://www.fedramp.gov/jab-or-agency-how-do-i-get-a-fedramp-ato/) for this. Seems like the difference is that the JAB review is done with a governmentwide scope whereas the agency review is done only for that specific agency. My understanding is that an agency could accept another agency's Agency P-ATO but they'd have more confidence with a JAB P-ATO.

mogul (approved):

PR #1097, opened 2017-10-12
Rearrange TIC docs

mogul:
IP restriction doc -> experimental section
TIC doc -> new "compliance" top-level section 🤷‍♂️


mogul:
Lemme know if you have a better plan for where Compliance stuff goes. TIC isn't technically part of getting an ATO, but agencies are very concerned about it as another OMB mandate they have to meet.

brittag:
Makes sense to me. Rearranged a little bit more:

screen shot 2017-10-12 at 10 57 21 am

Can we put https://cloud.gov/overview/security/for-assessors/ into this new section too, instead of in the "overview" section? It's more docs than marketing.

brittag:
cc @gboone and @femmebot

gboone (approved):
This looks good to me. It should be relatively easy to regroup this if/when we determine we need to. I've added it to the content audit so we don't forget about it.

PR #1098, opened 2017-10-13
Correct misstatement of SSP and actual policy

mogul:
We rotate yearly, not monthly.


brittag (approved):
👍  This makes this procedures doc match our SSP (specifically SC-12).

PR #1099, opened 2017-10-13
Add customer testimonial on homepage

femmebot:
See: https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-9499

screencapture-localhost-1313-1507904617760

cc @mogul @gboone


brittag:
Just took a quick look - in the PDF, Communicart has the Federalist URL. Other than that, looks good to me.

For assets like PDFs, I've found it helpful to add a hidden link in the source HTML to where we're keeping the source material in Google Drive (or elsewhere), so that it's easier for other people to later go edit the source material and make new versions - for example as seen in https://raw.githubusercontent.com/18F/cg-site/master/content/overview/technology/responsibilities.md

brittag (approved):

PR #1101, opened 2017-10-21
Add diagrams for domains and continuous deployment

brittag:
I had too much fun with the [Mermaid live editor](https://mermaidjs.github.io/mermaid-live-editor/) while thinking about how to [answer this question](https://gsa-tts.slack.com/archives/C0DF6LAMQ/p1508551252000078), so here are a couple new diagrams for review, along with a page that serves as a directory of diagrams.

I think [leveraging authentication](https://cloud.gov/docs/apps/leveraging-authentication/) could be illustrated next, giving an abstracted version of [this diagram](https://docs.google.com/drawings/d/1k1wykk5PbLKSNJj8FyZbIlpX0D8r1q3-w-uRK_WWt9g/edit). We could also provide an abstracted version of [this other diagram](https://docs.google.com/drawings/d/1nwclBJQfbuzsnGOqe88VukQl3uiH1Jfa4c0FT1Cq43I/edit).

For CI/CD page:
screen shot 2017-10-20 at 10 56 26 pm

For custom domains page:
screen shot 2017-10-20 at 10 56 47 pm

Under compliance section:
screen shot 2017-10-20 at 10 57 47 pm

### DIFF
--- a/content/docs/apps/continuous-deployment.md
+++ b/content/docs/apps/continuous-deployment.md
@@ -28,6 +28,34 @@ cloud.gov does not yet provide a CI/CD (continuous integration/continuous deploy
 
 You can configure your code repositories, [spaces]({{< relref "docs/getting-started/concepts.md#spaces" >}}), and CI/CD service together to enable automated or semi-automated deployments to your environments (such as development, staging, and production environments). For deployments in each environment, you can configure access control and testing requirements according to your project's needs.
 
+An overview of a common workflow for CI/CD with cloud.gov:
+
+{{< diagrams id-prefix="Figure-1.-Continuous-deployment-workflow" >}}
+graph TD
+
+subgraph cloud.gov
+  subgraph Org
+  subgraph Dev space
+      Dev[App]
+    end
+    subgraph Staging space
+      Staging[App]
+    end
+    subgraph Production space
+      Prod[App]
+    end
+  end
+end
+
+Developer((Developer)) -->|Commit code| Repo(Code repository)
+Repo -->|Automatically notify that a commit happened|CD(Continuous Deployment service)
+CD -->|If dev branch, deploy| Dev
+CD -->|If staging branch, deploy| Staging
+CD -->|If prod branch, deploy| Prod
+{{< /diagrams >}}
+
+## Service examples
+
 Here are examples of how to set up two cloud-based services that have free tiers for open source projects, [Travis](https://docs.travis-ci.com/) and [CircleCI](https://circleci.com/docs/1.0/). (You can also find and adapt instructions for using other CI/CD services with other Cloud Foundry deployments, such as [this explanation of how to use Jenkins](https://docs.cloud.service.gov.uk/#setting-up-the-cloud-foundry-jenkins-plugin).)
 
 ### Travis
--- a/content/docs/apps/continuous-deployment.md
+++ b/content/docs/apps/continuous-deployment.md
@@ -28,7 +28,7 @@ cloud.gov does not yet provide a CI/CD (continuous integration/continuous deploy
 
 You can configure your code repositories, [spaces]({{< relref "docs/getting-started/concepts.md#spaces" >}}), and CI/CD service together to enable automated or semi-automated deployments to your environments (such as development, staging, and production environments). For deployments in each environment, you can configure access control and testing requirements according to your project's needs.
 
-An overview of a common workflow for CI/CD with cloud.gov:
+To illustrate how a CI/CD workflow could be incorporated with cloud.gov:
 
 {{< diagrams id-prefix="Figure-1.-Continuous-deployment-workflow" >}}
 graph TD
--- a/content/docs/apps/custom-domains.md
+++ b/content/docs/apps/custom-domains.md
@@ -30,10 +30,10 @@ subgraph Amazon Web Services
     CG-DNS
     Router[App router]
     subgraph Org: agency-org
-    subgraph App A prod space
+    subgraph App A space
         AppA[App A]
       end
-      subgraph App B prod space
+      subgraph App B space
         AppB[App B]
       end
     end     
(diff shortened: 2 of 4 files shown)


content/docs/apps/continuous-deployment.md
```diff
@@ -28,6 +28,34 @@ cloud.gov does not yet provide a CI/CD (continuous integration/continuous deploy
 
 You can configure your code repositories, [spaces]({{< relref "docs/getting-started/concepts.md#spaces" >}}), and CI/CD service together to enable automated or semi-automated deployments to your environments (such as development, staging, and production environments). For deployments in each environment, you can configure access control and testing requirements according to your project's needs.
 
+An overview of a common workflow for CI/CD with cloud.gov:
```
pburkholder: This may be a common workflow, but it's not a recommended one because you have long-lived distinct branches which introduces a lot of integration problems. 

"A key correlation with high performing orgs is trunk-based delivery"

content/docs/apps/custom-domains.md
```diff
@@ -17,6 +17,38 @@ To make your app accessible via your custom domain name, use the [CDN service]({
 
 *Additional details are available in the [cloud.gov FedRAMP P-ATO documentation package]({{< relref "overview/security/fedramp-tracker.md#how-you-can-use-this-p-ato" >}}), including in System Security Plan controls SC-20, SC-21, SC-22, and SC-23.*
 
+### Comparison of default domains and custom domains
+
+Here's an example of the difference between a default *.app.cloud.gov domain and a custom domain. In this example, a staging application is using a default domain, and the production version of the application is using a custom domain.
```
pburkholder: I find this diagram pretty confusing because you're conflating a number of different things: 

* Custom DNS/app.cloud.gov DNS 
* CDN / origin
* staging spaces / production spaces

pburkholder (changes requested):
CI/CD should illustrate trunk-based delivery

We should probably draw the custom domains on a paper to make sure we the concepts and relationships right, then get into mermaid (or determine if it's the right tool for the job)

content/docs/apps/continuous-deployment.md
jmcarp: I'm not sure we want to get into specific recommendations about branching workflows. Maybe we can simplify this to something like commit -> test -> deploy, without reference to branches. Although that might be simple enough that we don't need a diagram at all.

content/docs/apps/custom-domains.md
jmcarp: I think this is confusing because it's accurate 😆. Maybe we can simplify by dropping *.app.cloud.gov DNS and staging--I think the main thing we want users to understand here is how production apps work.

content/docs/apps/continuous-deployment.md
brittag: I'd be happy not getting into specific recommendations for branching workflows, if we can avoid it. The question I'm attempting to answer is people asking "so, how would I set up staging and production environments in cloud.gov, using CI/CD?" - is there a better way to answer that without getting into the details? Diagram is optional, but diagrams can help a lot.

content/docs/apps/custom-domains.md
brittag: I want people to understand how default domain names work as well as production custom domain names, so that they understand the difference clearly, especially because sometimes people use `*.app.cloud.gov` for their "beta" sites.

We could remove "staging" and "production" space names though, and call them something like "space A" and "space B".

pburkholder:
So I tried to get a more accurate flowchart for testing but the rendering went all to hell because Mermaid doesn't have this:

https://github.com/knsv/mermaid/issues/270

e.g.: 

```
Developer((Developer)) -->|Commit code| Repo(Code repository)
Repo -->|Automatically notify that a commit happened|CD(Continuous Deployment service)
CD -->|If dev branch, deploy| Dev
CD -->|If prod branch, deploy, run tests | Preprod
Preprod --> | Test Results | CD
CD -->|If preprod tests OK, deploy prod branch| Prod
```

Resulted in:

image

So maybe we can go with:

```
Developer((Developer)) -->|Commit code| Repo(Code repository)
Repo -->|Automatically notify that a commit happened|CD(Continuous Deployment service)
CD -->|If dev branch, deploy| Dev
CD -->|If prod branch, deploy, run tests | Preprod
CD -->|If preprod tests OK, deploy prod branch| Prod
```

Which looks something like:

image


but I'd rather have a diagram that makes sense of the general idea more than I would one that absolutely-represents-best-practice.

content/docs/apps/continuous-deployment.md
pburkholder: In the end, lets leave the diagram exactly as is, but note that: "This workflow only demonstrates how the components of CICD and cloud.gov could work together, and isn't a recommendation. See (More resources) below"

Then I'm on the hook for providing some links(!).

pburkholder:
How about this to make the staging/production flow more distinct?

```
graph TD

subgraph Amazon Web Services
  subgraph cloud.gov
    CDN
    CG-DNS
    Router[App router]
    subgraph Org: agency-org
    subgraph Staging space
        Staging[App]
      end
      subgraph Production space
        Production[App]
      end
    end     
  end
end

Public((Public user)) -->|HTTPS| A-DNS(Agency DNS: agency.gov)
Tester((Testing user)) -.->|HTTPS| CG-DNS(DNS: agency_staging.app.cloud.gov)
A-DNS -->|HTTPS| CDN(CDN)
CG-DNS -.->Router
CDN -->Router
Router -.->Staging
Router-->Production
```

Ok, it's true it's hard to stop playing w/ Mermaid once you start!

image

pburkholder:
image


image

New diagrams!

pburkholder (dismissed):

brittag:
Thanks @pburkholder! I simplified one more piece (so that we don't have to talk about the subtleties of having multiple different production apps in an org), and this looks good to me now. You can approve and merge if it looks good to you too.

pburkholder (approved):

PR #1104, opened 2017-10-26
Add to maintenance list to check Kibana timezone

brittag:
Recommending a routine check once a week or so. See [thread here](https://gsa-tts.slack.com/archives/C09CR1Q9Z/p1508769927000054?thread_ts=1508768305000455&cid=C09CR1Q9Z) for context.


rogeruiz (approved):
:shipit:

PR #1105, opened 2017-10-26
[wip] Secrets dependency graph

rogeruiz:
Using Mermaid diagrams to map out our cloud.gov secrets dependencies in our Operator runbooks.

### DIFF
--- a/content/docs/ops/secrets-dependencies.md
+++ b/content/docs/ops/secrets-dependencies.md
@@ -1,0 +1,59 @@
+---
+menu:
+  docs:
+    parent: operations
+layout: ops
+title: Secrets Dependencies
+---
+
+This page exists to contain the secrets documentation for the dependencies
+across all of our deployments. As more of our secrets are automated, they will
+be address / removed from this age as resolved.
+
+### Certificate dependencies
+
+{{< diagrams id-prefix="certificates-depending-on-single-root-certificate" >}}
+graph TB;
+    %% ID Definition
+
+    s3(s3://cloud-gov-varz)
+    pipelines(Concourse Pipelines)
+
+    subgraph Root CA Certificates
+      ca_cert[Root CA Certificate]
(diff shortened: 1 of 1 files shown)


content/docs/ops/secrets-dependencies.md
```diff
@@ -0,0 +1,59 @@
...
+    pipelines --> pipelines_resource_bosh-deployment
+    pipelines --> pipelines_resource_cg-common
+
+{{< /diagrams >}}
+
+This chart represents the dependencies that rely on the single Root CA
+certificate. These dependencies exist in BOSH deployment manifests and in
+Concourse pipelines.
+
+Any Concourse pipelines with a type of `bosh-deployment` all use the Root CA
```
jmcarp: s/pipelines/resources

content/docs/ops/secrets-dependencies.md
```diff
@@ -0,0 +1,59 @@
...
+across all of our deployments. As more of our secrets are automated, they will
+be address / removed from this age as resolved.
+
+### Certificate dependencies
+
+{{< diagrams id-prefix="certificates-depending-on-single-root-certificate" >}}
+graph TB;
+    %% ID Definition
+
+    s3(s3://cloud-gov-varz)
```
pburkholder, line 19: How about a little less explicit, like 
```
s3(s3://secrets-bucket)
```

rogeruiz: If these docs were long-lived, I'd say that it may make sense to be generic, but we will have to be able to differentiate between the various buckets we have secrets in. Also this information is public in our `cg-provision` repo and are very likely never going to change unless we want to change _all the things_. 

https://github.com/18F/cg-provision/search?utf8=✓&q=cloud-gov-varz

pburkholder (dismissed):

PR #1113, opened 2017-11-08
Simplify getting started workflow 

brittag:
Here are some experiments to make getting started less confusing. The best way to review this would be to run it locally and try out the flow!

* On the "get access" page, remove a few unnecessary details.
* On the "how to log in"/"set up" page:
  * Simplify the explanation of the dashboard - I believe people don't need so much information just to get started, and not everyone gets a sandbox (contractors without federal email addresses).
  * Remove the sandbox targeting info, because this page should _just_ do one thing: help people get the tools set up.
* On the "try a deploy" page:
  * Add the sandbox targeting info.
  * Add a hello-worlds zip download link so that people don't have to mess with git and GitHub.
  * At the end, get people to the next step faster.


gboone (approved):

content/docs/getting-started/setup.md
```diff
@@ -6,64 +6,23 @@ title: Set up cloud.gov and log in
 weight: -50
 ---
 
-If you have set up your [access to cloud.gov]({{< relref "accounts.md" >}}), you can log into cloud.gov in the following ways -- dashboard (web interface) and command line.
+If you have [access to cloud.gov]({{< relref "accounts.md" >}}), here's how to log in.
 
 ## Log into the dashboard (web interface)
 
-cloud.gov has a dashboard that gives you easy web-based access to common tasks.
+The dashboard gives you web-based access to common tasks: [**`https://dashboard.fr.cloud.gov/`**](https://dashboard.fr.cloud.gov/)
```
femmebot, line 13: We should probably eliminate the `code` markup since it's just a hyperlink, not actual code.

brittag: Good point! Filed at https://github.com/18F/cg-site/pull/1114

PR #1116, opened 2017-11-15
Quickstart landing page

femmebot:
New landing page for workshop videos. (see https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-10036)

@gboone you may want to take a look at the content

screencapture-localhost-1313-quickstart-1510763469501


layouts/quickstart/index.html
```diff
@@ -0,0 +1,217 @@
...
+              <h4>Sample applications</h4>
+              <ul>
+                <li>
+                  <a href="https://github.com/18F/cf-hello-worlds">“Hello World” in different frameworks (Python, Ruby on Rails, PHP, etc)</a>
+                </li>
+                <li>
+                  <a href="https://github.com/18F/cf-ex-drupal">Drupal example</a>
+                </li>
+                <li>
+                  <a href="https://github.com/18F/cf-ex-wordpress">Wordpress example</a>
```
gboone: Should be WordPress

layouts/quickstart/index.html
```diff
@@ -0,0 +1,217 @@
...
+                <li>
+                  <a href="https://github.com/cloudfoundry-samples/spring-music">SpringMusic: Java + any of MySQL, Oracle, Postgres or Redis</a>
+                </li>
+              </ul>
+            </article>
+            <article class="usa-width-one-third">
+              <h4>Join the communities</h4>
+              <ul>
+                <li>
+                  <a href="https://chat.18f.gov/">The public Slack #devops channel at 18F (sign up at chat.18f.gov)</a>
```
gboone: "The public DevOps channel on 18F's Slack (sign up at chat.18f.gov)."

layouts/quickstart/index.html
```diff
@@ -0,0 +1,217 @@
...
+                    </p>
+                    <ul>
+                      <li>
+                        <a href="../assets/files/01-slides-why-cloudgov.pdf">Download slides</a>
+                      </li>
+                    </ul>
+
+                  </article>
+                  <article class="usa-width-one-third">
+                    <iframe width="280" height="157" src="https://www.youtube.com/embed/ip8G_Rafb_c?rel=0&amp;showinfo=0" frameborder="0" allowfullscreen></iframe>
```
pburkholder: Link to 

https://www.youtube.com/embed/ip8G_Rafb_c?rel=0&t=3m1s

Since the first 3 minutes are not relevant

layouts/quickstart/index.html
```diff
@@ -0,0 +1,217 @@
...
+          </section>
+        </div>
+      </div>
+
+      <div class="usa-content bgcolor-ltputty">
+        <div class="usa-grid">
+          <section class="step second">
+            <h2>Go build in your sandbox</h2>
+            <p>
+              Follow any of the tutorials below to launch your own web application from your sandbox. You’ll need your cloud.gov account and some familiarity with Windows Powershell or Mac/Linux terminal.
```
pburkholder: The additional sample applications are way down at the bottom of the page.

Can we move the other samples into this section? Or change the wording to either just reference a singular tutorial?

pburkholder:
If we're going to be using this style more, should it be named something other than `static/assets/css/quickstart.css`?

layouts/quickstart/index.html
femmebot: Should we just re-edit the videos? It seems weird not to have it start from the beginning. I added some copy that (hopefully) explains the jump at the 3:01 mark.

layouts/quickstart/index.html
femmebot: Fixed! see: https://github.com/18F/cg-site/pull/1116/commits/7dc2b8feb1b9c7f187d49254a6d29b8e84359c54

pburkholder (commented):
Some comments inline...

layouts/quickstart/index.html
femmebot: the original intent was to also include other tutorials in an accordion (see wireframe) but i'm just including the video for now since we needed to expedite this. (I think where things got muddy was that the content didn't match the  intent of the wireframe.) Anyhow, to solve this specific issue, I changed it to singular instead (see https://github.com/18F/cg-site/pull/1116/commits/fcedb3fb5f94cb8edabf66506184b0cca861f4fd)

mogul:
I'd move the `Sample Applications` from `Step 3` up into the bottom of `Step 2`... Eg " Ready for more? Check out some of these other sample applications".

layouts/quickstart/index.html
```diff
@@ -0,0 +1,217 @@
...
+            </p>
+            <!-- <form>
+              <label for="email">Federal email address</label>
+              <div class="form-wrapper">
+                <input type="text" name="email" value="">
+                <input type="submit" value="Submit">
+              </div>
+            </form> -->
+            <p>
+              <a href="https://cloud.gov/docs/getting-started/accounts/" class="usa-button">Get access</a>
```
pburkholder: Let's link directly to https://account.fr.cloud.gov/signup (see https://gsa-tts.slack.com/archives/C4XLJ1JBV/p1510777124000397)

femmebot: This is actually just a stopgap so we can have a landing page for the workshop emails. The intent is actually to have the input field to capture fed email addresses. Do we need to pause this so we can sync up tomorrow? I feel like we're not on the same page.

mogul:
That's fine and I like the direction, it just feels weird to have the sample applications after the "contact us" stuff. Also the introductory paragraph in Step 2 and the text under `Tutorials` is repetitive without that pattern established
image

So I suggest you make the Tutorials text the first thing under Step 2 and just remove the initial text (until the video tutorials are the first of more than one option following it).

pburkholder (dismissed):
I've decided I'm 👌 with this to go live so that /quickstart is a landing page we can use in the email to the workshop attendees. That will let us move one thing off the board to done, and move forward on trying to sell something to the attendees. 

I think that tuning the agreement between "Follow the tutorial" and "Watch our tutorials" and the placement of the other Getting Started links can come with the next iteration to close out the Favro card for hosting workshop content.

brittag:
Wonderful progress! Looking forward to seeing this on the site.

I would like us to clarify "public websites in production":
- Production applications don't need to be public, and I would like prospective users to feel comfortable about reaching out if they're interested in running an application for internal/non-public use.
- Not sure if people might interpret "websites" as meaning "static websites" rather than fully dynamic web applications.

For example, this could be clarified with a tiny edit: "websites in production" or "websites and other web applications in production".

Optional suggestions, up to your judgment:
- Not sure about calling this a "quick" start guide, because it may take a long time for some readers to complete (according to their own perception). This is based on my understanding of recommendations for general educational writing, such as [avoiding "simply", "just", and "easy"](https://css-tricks.com/words-avoid-educational-writing/).
- Not sure about using all-caps for headings, based on my understanding that [all-caps can be harder to read than sentence case](http://uxmovement.com/content/all-caps-hard-for-users-to-read/) and [can be interpreted as shouting](https://en.wikipedia.org/wiki/All_caps#Association_with_shouting.5B10.5D).
- Suggest asking people to email us if they have a suggestion, in addition to suggesting filing directly on GitHub. Many government people don't have access to GitHub or aren't allowed to use it.

layouts/quickstart/index.html
femmebot: done! (see https://github.com/18F/cg-site/pull/1116/commits/3a77ecfdc61d8f6e95221ab13d657e1f0f92fd73)

pburkholder:
I ran this locally and I like all the changes.  I made one commit to change the link for "Submit a Suggestion" from https://github.com/cg-site to https://github.com/cg-site/issues. That Ok?

> > Production applications don't need to be public, and I would like prospective users to feel comfortable about reaching out if they're interested in running an application for internal/non-public use.
> I can revise "…public websites in production" to "production websites" and eliminate "public" altogether.

I like calling it a "web application"

> @pburkholder @gboone thoughts on this? I wonder about this too, especially since a lot of government websites do tend to be brochureware.

> > Not sure about calling this a "quick" start guide

> Open to suggestions. The intent was to de-conflict this page with other Get started content on the site which, arguably, occurred later in the evaluation phase in the user's journey. The goal is simply to aggregate information for a new user who wants to get a feel for the product on their own without a lot of bother. While it may not be quick, psychologically, it's more about the reward ratio. We want to make them feel like the reward was at least commensurate to, if not more valuable, than the effort expended.

It's a lot quicker than the getting started page since we've not tried to address the permutations that the other page does. I'm accustomed to quick start meaning: You'll get something running without really knowing what you're doing.   Althought the URL is /quickstart, it's not as if we're putting quick in bold or allcaps. In short, I'm okay with it, but I would be open to "Express guide" or "Walk through" or something else. 

> > Not sure about using all-caps for headings

> Yup, I'm familiar with this! As with a lot of prescriptive guidance, it depends. I've also observed in eye-tracking labs where all-caps helped readers quickly skim for clues. I think the takeaway is that if you want to write for deep comprehension, then use sentence case. Also, the titles were breaking oddly or appeared justified and didn't read well as sentences. I just made other adjustments to the titles themselves so they rag.

I approve of the decapitation.

> > Suggest asking people to email us if they have a suggestion

> Do we have an email address for this? I think we should also figure out how to track/segment different inbound emails — esp. when they point to the same address.

What you have works for me.

brittag:
Thanks! We just have two email addresses, inquiries and support, and either is fine for general suggestions. We can sort-of segment with the pre-filled subject line or email body.

Integrating these comments, a couple alternate options to consider re "quick start":
- Leave the URL as-is and trim "Follow this quick start guide" to "Follow this guide" in the intro.
- Call this page something like a "get started" guide and rename the "getting started" category in the docs to something more specific. :)

(I don't consider these to be blockers for integrating this page into the site; they can be considered later.)

brittag:
That looks wonderful - thank you @femmebot! I'll leave this to @pburkholder and @gboone for approval/merge if they agree.

gboone (approved):

mogul:
> We just have two email addresses, inquiries and support, and either is fine for general suggestions. We can sort-of segment with the pre-filled subject line or email body

For future reference, GMail supports [plus-addressing](https://www.wikihow.com/Use-Plus-Addressing-in-Gmail) which is an easy way to pretend you have a different mailbox for each place where you list our address, while allowing you to sort/filter/process them based on origin later.

PR #1120, opened 2017-11-16
Rotating service keys for deployer accounts

pburkholder:


content/docs/services/cloud-gov-service-account.md
```diff
@@ -44,12 +44,53 @@ cf create-service-key my-service-account my-service-key
...
+## Updating/Rotating credentials
+
+Service accounts are subject to the same password expiration policies as other cloud.gov accounts. If you see an error like:
+
+```
+Error Code: 403
+Raw Response: {"error":"access_denied","error_description":"Your current password has expired. Please reset your password."}
+```
+
+Then you'll need to delete the service account, recreate it, and update the username/password in your deployment scripts. For example:
```
jmcarp: No need to delete the service instance--users can just delete and recreate the service key. Each key/binding gets a unique account.

jmcarp: Actually, this page already has a section on rotating credentials at the bottom. Let's revert the new section and add any missing information to the existing section.

content/docs/services/cloud-gov-service-account.md
```diff
@@ -60,9 +88,6 @@ If you're trying to retrieve credentials for a service instance created before J
 
 To use this service, see [continuous deployment]({{< relref "docs/apps/continuous-deployment.md" >}}).
 
-### Rotating credentials
```
jmcarp: We include this section for most service pages, so it might more sense to keep the usual section heading and move extra information here. WDYT @brittag ?

pburkholder: We use "Rotating" in our operations docs, but I think "updating" is more applicable to our audience. I don't see any example of the term "rotating" in customer-facing docs; LMK if I'm mistaken.

pburkholder: Sure, let's use rotate for consistency.

content/docs/services/cloud-gov-service-account.md
```diff
@@ -44,11 +44,39 @@ cf create-service-key my-service-account my-service-key
...
 
 After you create one of these service keys, you will see a new "user" in your org and space with a name made of 36 letters, numbers, and dashes as its unique identifier, similar to `f6ab4cfb-6e6c-4b10-8585-3f39e740905c`. In your event logs, its actions will display as actions by `service-account@cloud.gov`.
 
-These credentials can be used with the `cf login` command in automated deployment scripts.
+These credentials can be used with the `cf auth` command in automated deployment scripts.
+
+### Rotating credentials
+
+Service accounts are subject to the same password expiration policies as other cloud.gov accounts. If you see an error like:
```
brittag: "same password expiration policies as other cloud.gov accounts" - this might be kind of confusing since a lot of us have cloud.gov accounts via agency SSO. Can we list out something specific instead, like "they expire every 90 days" (if that's accurate)?

content/docs/services/cloud-gov-service-account.md
brittag: Consistency in headers across the various service pages is great. Thanks!

brittag:
Just checking - is this just documenting existing behavior or are there any breaking changes here? I _think_ this is just documenting existing behavior - if so, no concerns. Yay for docs!

content/docs/services/cloud-gov-service-account.md
```diff
@@ -44,11 +44,39 @@ cf create-service-key my-service-account my-service-key
...
+The last command will return a username/password pair, that you can use, like this:
+
+```
+{
+ "password": "oYasdfliaweinasfdliecV",
+ "username": "deadbeed-aabb-1234-feha0987654321000"
+}
+```
+
 This will create a cloud.gov service account and make the credentials available to you via a service key. Keep these credentials secure. If they’re compromised, the way to invalidate the credentials is to delete the service key (you can create another, and it will have a fresh set of credentials). Each service key that you bind to your instance creates a separate service account with different credentials; you can create as many service keys per instance as you like. <!-- this advice should match on /docs/services/cloud-gov-service-account/ + /docs/services/cloud-gov-identity-provider/ -->
```
brittag: Suggest linking https://docs.cloudfoundry.org/devguide/services/service-keys.html#delete whenever we mention "delete the service key", since people may read this in a panic after a compromise and want a quick link to the instructions (without reading further on the page).

brittag:
I checked this more carefully and verified that it is only documenting current behavior better (no breaking changes). I added a commit with the following changes:

* Add link to `https://docs.cloudfoundry.org/devguide/services/service-keys.html#delete` to the same spot in both the service account service page and the identity provider service page.
* Link to the docs for cf auth.
* To keep the rotating credentials section relatively consistent with similar service pages, I rearranged the order of the paragraphs and added back a slightly-redundant-but-probably-still-useful paragraph.
* Updated "Service accounts are subject to the same password expiration policies as other cloud.gov accounts." to "Service account passwords expire every 90 days." to help readers understand what to expect, because not everyone has a cloud.gov account (some people have SSO accounts).

mogul (approved):

PR #1122, opened 2017-11-16
Avoid confusing sandbox and space terminology

mogul:


brittag (approved):
Thanks, I believe this accurate phrasing is helpful to prevent confusion as people get to know cloud.gov better. (For example, understanding that they can use the same account for access to both their sandbox space and paid orgs.)

Issue #592, opened 2016-11-23
Make checklist of requirements for making a new cloud.gov repository


brittag:
In order for the cloud.gov team to follow our own best practices along with our security and compliance requirements, we should have a checklist for making a new cloud.gov repository.

This could be part of our [configuration management plan](https://cloud.gov/docs/ops/configuration-management/), which already includes our GitHub contribution guidelines.

**Examples of what this might include**

* Give it a name starting with `cg` or `cf`
* Add LICENSE, CONTRIBUTING, and README files in compliance with the 18F Open Source Policy
* For repositories with 18F-developed code, set up Code Climate (SI-3)
* Configure a protected master branch with required reviews (CM-9)
* If it's a platform configuration repo, restrict permissions to Cloud Ops
* A continuous integration (CI) server handles automated tests and continuous deployment (CD) of the merged changes (CM plan)

**Acceptance criteria**

- [ ] A cloud.gov docs page provides a helpful checklist for repository requirements.
- [ ] We've notified our team about this (through a `@cloud-gov-team` notification and a team-wide email).

brittag:
Related: https://github.com/18F/development-guide/tree/master/project_setup#initialization-checklist

msecret:
Speaking of `Give it a name starting with cg or cf`, there discussion going on in `cg-dashboard` about how it should technically be renamed to `cf-dashboard` because we're open sourcing it and now other teams using Cloud Foundry are using it. Maybe the guidance should be something like "if the repository is specific to cloud.gov or 18F, it should be named `cg`, otherwise `cf"?

PR #1126, opened 2017-11-17
Add checklist for new repositories to configuration management plan

brittag:
This follows up on the idea at https://github.com/18F/cg-site/issues/592 by adding a small, slightly aspirational checklist for new cloud.gov repositories. Needs review to make sure this says reasonable things, including the right set of repo permissions.

This change also removes the pipeline screenshot and the link to an example, since those parts seemed unnecessary.

### DIFF
--- a/content/docs/ops/configuration-management.md
+++ b/content/docs/ops/configuration-management.md
@@ -48,12 +48,14 @@ is encouraged and qualifies as a review. Review should include assessment of arc
 
 ## Checklist for new repositories
 
-* Give it a name starting with `cg-` or `cf-`.
-* [Add `LICENSE`, `CONTRIBUTING`, and `README` files](https://github.com/18F/open-source-policy/blob/master/practice.md#how-to-license-18f-repos).
+Before we put a new repository into production:
+
+* Give it a name starting with `cg-` or `cf-` (to help us identify our work, unless we have a good reason to name it something else).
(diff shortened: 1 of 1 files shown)


pburkholder:
How about a pattern of naming IP projects or research spike projects, `cfx-...` or `cgx-...` and the completing the full checklist before bring projects into production? 

I don't want to include things in our list of repos until we're actually dependent upon them, as that implies we're actually maintaining them.

brittag:
@pburkholder I put in a commit to reframe this checklist as only applying to production repositories, since it's fine to be much more flexible about repositories for research and experiments. Does this revised version make sense to you?

pburkholder:
Why are the repos "Read" for 18F since they should be open to all?

Do we ever use private repos? I can think of instances where we should use private repos, like when we have the public certs for OAuth, instead of treating them like secrets, but that's a whole other kettle of fish, just asking if we should say:

> At this point we don't use private repositories.

Thanks, Peter

brittag:
Good point, "Read" is a bit redundant. I think it's mildly helpful though, since we may occasionally have non-public repositories for internal coordination (like https://github.com/18F/cg-migration), or we may temporarily make a repository non-public during incident response.

pburkholder:
TIL that "Read" enables the ability to submit PR reviews. https://help.github.com/articles/about-pull-request-reviews/

pburkholder (approved):

PR #1132, opened 2017-11-21
Link to CF docs on env vars, kudos to DTA

pburkholder:


jmcarp:
I kind of want to add "...but you shouldn't use this feature, since environment variables will be lost if your app is recreated--for example, by the `autopilot` plugin. We recommend storing non-sensitive configuration variables in your application manifest and sensitive variables in user-provided services."

pburkholder:
@jmcarp Good point. Edited.

gboone (approved):

PR #1137, opened 2017-11-28
Remove the public user path

mogul:
DHS found it confusing since it's unrelated to the text and not specifically relevant for explaining TIC concepts.


pburkholder:
Before:

image

After:

image

pburkholder (commented):
The meaning of the dashed line is unclear. Is it blocked/filtered at the TIC IP filter?

If mermaid flowcharts supported `--x` arrowheads that might be clearer. 

Maybe make the caption longer to describe the flow.

mogul:
Yes, that's the meaning of the dashed line; I believe it's explained in the text. 

Mermaid sadly gives us no such option. If I could, I'd have the arrow end at the subgraph boundary, but it doesn't support arrows to subgraphs either. :(

I don't want to make the caption longer because that caption by itself element represents something specific (the "encrypted tunnel") in the DHS reference TIC architecture diagrams we're referring to. Extending it with an unrelated meaning will make it harder to understand. Do you have other ideas for how represent this?

brittag (approved):
I believe this is an improvement over the current version, simplifying the first diagram in a useful way! We should definitely keep thinking about ways to clarify the diagrams as we get more feedback on them though.

PR #1140, opened 2017-12-05
More detail for reviewing alerts in ops maintenance list

wjwoodson:
+ Add pagerduty to alert sources
+ Add more detail for reviewing alerts and specify that possible security issues should enter security incident response process


brittag (approved):
Thanks! This is very helpful for making the previously-implicit expectations clear. Just approving for now and not merging yet in case other Platform people have suggestions, but can merge tomorrow if nobody has further feedback.

LinuxBozo (approved):
Agree, these are great additions.

PR #1144, opened 2017-12-06
Update version info for RDS plans

brittag:
Update current Oracle and Postgres minor versions (didn't check MySQL).

Add details about how versions work.


pburkholder:
I think we'll need to say something about Oracle upgrades. From my read of https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_UpgradeDBInstance.Oracle.html#USER_UpgradeDBInstance.Oracle.Minor, how about:

```
For Oracle, minor upgrades are not automatic. You'll need to contact [cloud.gov support](mailto:cloud-gov-support@gsa.gov), and schedule a maintenance window for the upgrade to take place. The time for the outage varies based on your engine version and the size of your DB instance. 
```

brittag:
@pburkholder Good point, added a slightly modified version!

mogul (approved):

PR #1148, opened 2017-12-07
Remove mention of cg-migrate-db

brittag:
https://github.com/18F/cg-migrate-db is labeled DEPRECATED, so we shouldn't refer to it in our documentation.

### DIFF
--- a/content/docs/services/relational-database.md
+++ b/content/docs/services/relational-database.md
@@ -61,22 +61,6 @@ The contents of the `DATABASE_URL` environment variable contain the credentials 
 
 ## Access the data in the database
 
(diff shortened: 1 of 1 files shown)


adborden:
@brittag do you know why cg-migrate-db is deprecated? It doesn't list a reason why or if there is a preferred method for migrating databases that supersede cg-migrate-db (I am currently looking for something to use on my project fwiw).

brittag:
@adborden Hmm, that's a good question. I'll bring it up in #cg-services.

brittag:
We discussed this more, and we'll maintain it as a team.

PR #1152, opened 2017-12-13
Update secrets management docs

wjwoodson:
Update secrets management docs to cover processes for shared secrets and concourse creds:

1. [Shared](#updating-shared-secrets) - centrally managed secrets used across multiple deployments
1. [Deployment](#updating-deployment-secrets) - secrets managed and stored per bosh deployment and environment
 1. [Concourse](#updating-concourse-secrets) - secrets consumed as credentials in Concourse CI pipelines


content/docs/ops/secrets.md
```diff
@@ -53,27 +53,31 @@ aws-vault exec cloud-gov-govcloud bash
...
+Centrally managed secrets are generated by a Concourse pipeline.
 
+1. Trigger the relevent build in [secrets-rotation](https://ci.fr.cloud.gov/teams/main/pipelines/secret-rotation). This creates new secrets and stores them in a common file.
+2. Trigger the relevent deployment(s).
 
-#### UAA credentials
+#### Updating Deployment Secrets
 
-All UAA clients and users and associated credentials should be created via the Cloud Foundry secrets or the [service account]({{< relref "docs/services/cloud-gov-service-account.md" >}}) or [identity provider]({{< relref "docs/services/cloud-gov-identity-provider.md" >}}) services. UAA accounts should not be created manually; we reserve the right to drop permissions for or deprovision hand-propped accounts.
+Secrets which are not centrally managed must be updated per deployment and environment. Substitute `cf.main.yml` in the following steps for the relevant file.
```
LinuxBozo, line 80: Does `cf.main.yml` still make sense to reference here, even to be replaced?

content/docs/ops/secrets.md
```diff
@@ -157,3 +161,41 @@ All UAA clients and users and associated credentials should be created via the C
...
+
+1. Update the Concourse pipeline with `fly`.
+
+    ```sh
+    fly -t cloud-gov-govcloud set-pipeline -p deploy-bosh -c pipeline.yml -l concourse-environment.yml
+    ```
+
+1. Don't leave the secrets lying around (for security reasons, and because they get stale).
+
+    ```sh
```
jmcarp, line 199: Would it be overkill to recommend `mktemp` or `mktemp -d`? Or maybe begin each example with `cd $(mktemp -d)`? I would never do this, but it's possible that we might forget to clean up files after doing secrets work.

wjwoodson: 👍

content/docs/ops/secrets.md
wjwoodson: For the moment I think we can still reference `cf.main.yml`. This portion of secrets rotation may/will change with [cf-deployment](https://github.com/cloudfoundry/cf-deployment) and eventually opsfiles for all deployments.

LinuxBozo (approved):

jmcarp (approved):

PR #1153, opened 2017-12-14
Homepage sans industry reports section

femmebot:
I created a version of the homepage without the links to the reports so we can hopefully push this while we work on getting clearance for including links to 3rd-party reports. (This replaces https://github.com/18F/cg-site/pull/1150)

screencapture-localhost-1313-1513280914402


layouts/index.html
```diff
@@ -9,6 +9,7 @@
 
 
 
+<<<<<<< HEAD
```
brittag: @femmebot Looks like some merge syntax got in here, and I'm not quite sure how to fix it...

brittag: (I tried just removing the merge syntax locally, but the homepage displayed weird after I did that.)

brittag: Ah nevermind, I compared it to https://landing-home.app.cloud.gov/ and figured out how to fix it! Time to merge!

brittag (approved):
This revision is great! I have a couple suggestions for minor adjustments that I can file as a followup PR instead of blocking this one.

PR #1154, opened 2017-12-14
Eliminate redirects to landing.app.cloud.gov when there's no trailing slash

mogul:
Trying to increase deployment independence and avoid landing on unexpected URLs depending on the presence of a trailing slash.

I _think_this means that URLs not reached through the baseurl will redirect to the baseurl, and any generated URLs will be relative. independence.
https://discourse.gohugo.io/t/canonifyurls-relativeurls-interaction/5448

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -1,4 +1,6 @@
 baseurl = "https://cloud.gov"
+canonifyurls = "true"
+relativeurls = "true"
 languageCode = "en-us"
 title = "cloud.gov"
 pluralizelisttitles = "false"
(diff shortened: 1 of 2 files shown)


mogul:
This is an attempt to fix the problem noted in #1009:

> when we have links without a trailing slash, those links redirect to URLs like this: `https://landing.app.cloud.gov/docs/services/cloud-gov-identity-provider/`

I'm not entirely sure how to test this...

mogul:
I'm unable to reproduce Concourse's Hugo build error locally. Can someone help?

wjwoodson (approved):

mogul:
Any idea why I want able to reproduce this outside Concourse??

PR #1156, opened 2017-12-15
Document default security headers.

jmcarp:
cc @femmebot


brittag:
Yes this is great! I moved this to the "advanced"/"managing apps" section (since it's not usually something you need at first), and I added some background.

LinuxBozo (approved):
🚀

PR #1159, opened 2017-12-18
Add login info to first deploy page

brittag:
This duplicates the info at https://cloud.gov/docs/getting-started/setup/ so that the first deploy page is more self-contained.


pburkholder:
Worth remembering:
```
$ cf login -a api.fr.cloud.gov --sso
API endpoint: api.fr.cloud.gov

One Time Code ( Get one at https://login.fr.cloud.gov/passcode )>
```

You get two (2) URLs: api.fr.cloud.gov AND https://login.fr.cloud.gov/passcode. So if we say, "visit that link" it can be ambiguous which link we mean.

I really wish we could change that to echo our login link in bold and color.

brittag:
@pburkholder That sounds like a good PR or issue to file against https://github.com/cloudfoundry/uaa/blob/master/model/src/main/java/org/cloudfoundry/identity/uaa/zone/IdentityZoneConfiguration.java - it makes sense to highlight the link with bold escape codes (color is trickier since people have different terminal color configurations).

pburkholder:
https://github.com/cloudfoundry/uaa/issues/751

pburkholder (approved):

PR #1165, opened 2017-12-26
removed hover underlines from USA banner and top nav

femmebot:
Removed hover underlines from appearing on nav and banner items. (Apply hover underlines to content areas only.)
Also removed unnecessary youtube video that was commented out in the quickstart section.


pburkholder:
I can't see what's different, visually, between this branch and master.

pburkholder (approved):

PR #1172, opened 2017-12-28
Mention multi-buildpack support

brittag:
Followup on https://github.com/18F/cg-site/pull/1170 - adding a brief mention of built-in multi-buildpack support on the custom buildpacks page. It's not _exactly_ a custom buildpack in itself, but this is likely where people will look for experimental buildpack-related features.


mogul (changes requested):
This probably needs at least a nod to the actual use-cases for this, eg if your application needs to make use of components implemented in a different language.

brittag:
Added this info; up for another review!

mogul (approved):

PR #1176, opened 2018-01-03
Clean up stale assets, styles

femmebot:
Please see corresponding cg-style PR: https://github.com/18F/cg-style/pull/334

- added alert notification in header (currently commented out)
- updated 404 page style so we can eliminate dependency on outdated styles and assets from old homepage versions
  
[see Favro card](https://favro.com/card/1e11108a2da81e3bd7153a7a/18F-10886)


brittag:
Yeah, only Platform squad can see Concourse - @pburkholder should be able to access it. (I can't.)

jmcarp:
@femmebot @pburkholder: what's the status on this patch? Should we resolve conflicts and merge?

brittag:
@jmcarp I believe the status is that this is blocked on understanding why the build is failing - we should resolve conflicts and fix the build, then merge.

pburkholder:
I don't think we should accept this PR. It adds the 'Alert' feature, with which we could enable an alert and link to a statuspage in the event of an extended outage. But that assumes we'd want to, and be able to, rebuild our site in the event of an outage.

The other clean up changes are pretty minor and are mixed in with some style tweaks that would probably take more time to sort out than is merited.

OK to close without merging? --Peter

PR #1185, opened 2018-01-10
WIP Switch new subscriptions to different list

gboone:
This updates the newsletter form to send subscriptions to the master list. Right now they're subscribing to one list and we're batch migrating them every month. This results in duplicative users which is messy and, eventually, expensive.


gboone:
Alright, I've tested that this works. I subscribed myself with a test email address on localhost and showed up on the list in MC.

We can merge this now, then I need to combine the remaining lists into one master.

brittag (approved):

PR #1187, opened 2018-01-11
Deletion policy not clear

dylanking42:
One bullet said "in the future we will delete", the final bullet point said "we delete every 90 days."

I presume the second one is the actual policy in place, but it should be clarified.


pburkholder (approved):
Thanks!

pburkholder:
Appreciate your help, Dylan!

PR #1189, opened 2018-01-16
update 'record' to 'records'

tmaybe:
record is not valid syntax for terraform


pburkholder (approved):
I see what you mean: https://www.terraform.io/docs/providers/aws/r/route53_record.html

PR #1190, opened 2018-01-19
Added Google Tag Manager scripts on homepage

femmebot:
Included the GTM scripts, one in the head, the other in the body, so we can begin tracking click events on the homepage.

### DIFF
--- a/layouts/index.html
+++ b/layouts/index.html
@@ -6,7 +6,7 @@
 
   <body id="home">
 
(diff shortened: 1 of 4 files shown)


pburkholder:
Ok, I looked at https://developers.google.com/tag-manager/quickstart and I understand all this except: What's the purpose of partials/single.html?

pburkholder (approved):

PR #1191, opened 2018-01-20
Added announcement page in the event of a shutdown

femmebot:
Used content from this repo: https://github.com/18F/cf-ex-staticfile

screencapture-localhost-1313-announcement-1516633259231

### DIFF
--- a/content/announcement/index.md
+++ b/content/announcement/index.md
@@ -1,0 +1,5 @@
+---
+type: announcement
+layout: index
+title: Announcement
+---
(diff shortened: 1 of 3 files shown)


mogul:
We should be advising that they deploy the static app first, remap the route, and then stop their existing app.

pburkholder:
@mogul In this case I disagree. People trying to shutdown this a.m. probably aren't worried about a few minutes of downtime. I'd rather have them get it working with fewer steps, then run into snags or have to open support requests.

pburkholder:
How about we move this to docs/apps/staticsite and just publish as a regular site update?

brittag:
Please do not merge this until approved by Shashank or me, so we can ensure it is aligned with org comms.

mogul:
@pburkholder it's not more steps to do it with zero downtime. They just push the static app with no-route, map-route to it, then unmap-route from the original app.

pburkholder:
Let's close this P.R. and come back to it if there's another, longer shut down.

PR #1197, opened 2018-01-30
move GTM head script inside head

femmebot:
Small update to move Google Tag Manager head script inside head as prescribed by Google.


pburkholder (commented):
Can we just bite the bullet and add `{{ partial "gtm-head.html" . }}` to head.html?

Also, `git rm gtm-head.html` if we don't need it anymore

pburkholder (approved):

PR #1207, opened 2018-02-15
Drop statically imported and outdated jquery.

jmcarp:
Because we already install jquery with npm and bundle it with browserify.


brittag (approved):
I tried this locally, and tabs still worked on https://cloud.gov/docs/apps/managing-teammates/ and everything else looked fine too.

PR #1212, opened 2018-02-27
Update javascript.html

pburkholder:
Not sure how to test this, but according to https://mailchi.mp/support/action-needed-re-type-ahead-script-february-search-tips this should read like:


mogul:
PR is stale... Who can help us determine that this change is right? Do we just need to test that it works?

jmcarp (approved):

PR #1218, opened 2018-03-08
Pricing landing page

femmebot:
WIP 

cc @jeanninehunter @mogul @pburkholder


layouts/pricing/index.html
```diff
@@ -0,0 +1,499 @@
...
+            </li>
+            <li>
+              Continuous platform-level scanning and security updates
+            </li>
+          </ul>
+        </div>
+
+        <div class="usa-width-one-half">
+          <p>
+            Access to free services
```
mogul: We should move `Access to free services` to the bottom of the column to the left, and just make this text `Free services`.

layouts/pricing/index.html
```diff
@@ -0,0 +1,499 @@
...
+              </li>
+              <li>
+                Access to premium add-on services
+              </li>
+          </ul>
+        </div>
+
+        <div class="usa-width-one-half">
+          <p>
+            Access to premium services
```
mogul: Similarly, we should move `Access to premium services` to the bottom of the column to the left, and just make this text `Premium services`.

mogul (changes requested):
I feel pretty confident about these two changes, but I'm not all that confident that we've retained clarity that the Premium services will be charged individually in future.

layouts/pricing/index.html
mogul: This probably needs a text line after the title, something to approximate the `We aren’t yet charging for paid services. See managed services for details.` item in the existing page.

femmebot: One of the earlier versions had "Currently free for a limited time". Does that work?

mogul: How about `Available free for a limited time`? That's pretty explicit: Paying gets you access, but you're not charged extra... yet!

jeanninehunter: @mogul That sounds like a plan: `Available free for a limited time`

layouts/pricing/index.html
```diff
@@ -0,0 +1,500 @@
...
+              <li>
+                Single-sign-on with your SAML identity provider
+              </li>
+              <li>
+                Access to premium add-on services
+              </li>
+          </ul>
+
+          <p>
+            Premium services &mdash; free for a limited time
```
pburkholder: I think "free for a limited time" doesn't provide any clarity over just saying "Premium services include"

It's not as if Netflix is priced at "$8.99 for a limited time" - it's implicit in the price of anything that future pricing will be different. And it's not clear if "free for a limited time" means you get the first six months, say, of your services for free, then start paying.

This is a minor point, however, and I rather ship this and bandy about the wording later if we need to.

pburkholder (dismissed):
This is terrific.  There are a couple things I'm not 100% on, but I'd love to get this out there and see how it feels having customers see it.  :shipit:

layouts/pricing/index.html
```diff
@@ -0,0 +1,500 @@
...
+                MySQL database
+              </li>
+              <li>
+                Elasticsearch RESTful search and analytics
+              </li>
+              <li>
+                Redis in-memory data store
+              </li>
+              <li>
+                S3 object store
```
pburkholder, line 185: Whoops: S3 is _NOT_ available in sandboxes.

layouts/pricing/index.html
```diff
@@ -0,0 +1,500 @@
...
+              </li>
+            </ul>
+          </div> -->
```
pburkholder: Remove comment to clean up....

layouts/pricing/index.html
femmebot: Done

layouts/pricing/index.html
femmebot: Done

mogul:
So sorry... I meant keep the two columns, just put the "access" text under the list of "things you get" on the left and rename the top of the right-hand column accordingly!

layouts/pricing/index.html
mogul: If we don't telegraph that this is coming, It's not just that pricing will go up, it's that the model will change. We want them to "get" the model and not have to readjust that understanding if possible so that it will be non-controversial when it starts appearing in their renewals and bills.

layouts/pricing/index.html
mogul: We're making it included.

mogul:
(Because I want people to be able scan down between the columns and understand the differences easily... I think one column hinders that!)

mogul (approved):

PR #1219, opened 2018-03-10
Add minimal sso integration docs.

jmcarp:
Our customer-facing docs mention sso obliquely, but don't clearly explain that sso is possible or how to set it up. This is my attempt at a minimal description of the feature and how to get started. It's not meant to explain how saml works, just who needs to be on the call to get the integration set up.

cc @cmc333333


content/docs/apps/sso.md
```diff
@@ -0,0 +1,8 @@
+---
+menu:
+  docs:
+    parent: advanced
+title: Single sign-on integration
+---
+
+If your agency has an identity provider that supports SAML single sign-on, such as Active Directory or SecureAuth, you can integrate your identity provider with cloud.gov so that users from your agency can authenticate to cloud.gov using your agency provider. To integrate your identity provider with cloud.gov, you'll need to exchange SAML metadata with us. This is a straightforward process, but it isn't self-service; you can set up a meeting between a cloud.gov engineer and your identity provider administrator by emailing cloud-gov-support@gsa.gov.
```
mogul: Active Directory doesn't support SAML by itself, just provides an embraced and extended LDAP. We should refer to `Active Directory Federation Services (ADFS)` instead, which is the SaaS IdP which supports  SAML clients.

mogul (changes requested):

mogul (approved):

PR #1224, opened 2018-03-15
Remove old pricing page, and ensure redirects land on the new one.

mogul:
I can't seem to get rid of a bunch of errors around relref expansions, even after upgrading , but that's on master too. Oh well!


mogul:
How about instead we...
* Replace the content of this page with a short blurb that basically says, "For more information about our rates, check out our pricing page" with a link. 
* Put a link at the bottom of the pricing page that basically says, "Want to know more about our pricing model and policies? We've got that info in our docs!" with a link

mogul:
(Will hold off on fixing the bad refs that caused the status failure above until we decide what to do.)

PR #1225, opened 2018-03-16
Note about disabling SSH for production

mogul:
First commit to help clarify that SSH should be disabled for production spaces/applications.


mogul:
I'm doing this as part of following through on a request from the GSA OCISO team way back when before I re-engage them about the status of the stuff that they considered blockers to adoption of cloud.gov.

brittag (approved):

PR #1230, opened 2018-03-22
Add initial sysdig falco documentation.

jmcarp:
I'm sure this will need polishing, but here's a start.


brittag:
This is a good start - I believe users would want to know (described very briefly):

* What is the definition of "unexpected behavior" / who decides what's unexpected? Is there a list of types of unexpected behavior?
* What should they do about those log items? Do they need to do anything?
* How does this help them? Does it help them comply with any NIST requirements?

brittag:
This is looking pretty good to me! It could always be better, but I believe it's ready for merge as a preview of the feature. Changes in my commits:

* Put intro paragraph in future tense, to help the reader get in the right mindset for reading the page.
* Declarative titles instead of questions (for easier skimming).
* Only use "we" when speaking about human/manual actions; use "cloud.gov" and "automatically" when speaking about platform features.
* Consistently speak to the reader as "you".
* Rearrange a couple of sentences to clarify cause-and-effect.

mogul (approved):

PR #1236, opened 2018-03-30
Move maintenance page out of "getting started"

mogul:
Also includes that we now notify when a stale buildpack is in use.


jeanninehunter (dismissed):
LGTM!

pburkholder:
Shoot, still getting:

```
+ hugo
Started building sites ...
ERROR: 2018/04/03 00:57:53 template_funcs.go:1398: No page found with path or logical name "docs/getting-started/app-maintenance.md".
```

brittag:
I tried fixing some things that looked wrong, but the build is still breaking locally for me. I pushed up my changes to see if the real environment would work better, but nope. Not sure what's wrong here.

mogul:
The issues with building here will probably fixed once #1256 is merged to master and then this branch.

brittag:
@mogul Would you be up for taking a look at why this build is failing? I can't figure it out.

mogul:
Doh, that should fix it.

brittag (approved):

PR #1237, opened 2018-04-03
clarified 90-day expiration verbiage for sandbox on Pricing page

femmebot:
re: https://gsa-tts.slack.com/archives/C4XLJ1JBV/p1522772258000320

Changed `Non-production systems, 90-day expiration` to _Non-production systems; 
systems cleared every 90 days_ (Note: I used `cleared` rather than `deleted` because it seemed less destructive.)

screencapture-localhost-1313-pricing-2018-04-03-14_36_40


jeanninehunter (approved):
Agree that opting not to use "deleted" is less alarming, @femmebot. LGTM!

PR #1242, opened 2018-04-06
copyedit on security headers set by cloud.gov

wjwoodson:


brittag (approved):
You're right! Thank you @wjwoodson!

PR #1245, opened 2018-04-11
Initial draft to rotate smtp credentials

sharms:


content/docs/ops/runbook/rotating-smtp.md
```diff
@@ -0,0 +1,54 @@
...
+## Rotate TLS Certificates
+The following keys from the
+[Postfix Deployment Pipeline](https://github.com/18F/cg-deploy-postfix/blob/master/secrets.example.yml)
+can be rotated using the `openssl` command line:
+
+1. `postfix_tls_cert`: `smtp.fr.cloud.gov.crt`
+2. `postfix_tls_key`: `smtp.fr.cloud.gov.key`
+
+```sh
+# Generate CA
```
timothy-spencer, line 27: Not sure if this is better or worse, but this is what I do to generate a self-signed cert:

`openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -days 365 -nodes -sha256`

This might or might not work, because there's no SAN stuff, but we shall see.

timothy-spencer (approved):
Glad you got this going!

PR #1247, opened 2018-04-11
Broken link replaced with current instructions.

jontours:
Verified by deploying cf platform locally using bosh-lite, logging in to cf api, creating org and space, and pushing a sample app.


timothy-spencer (approved):
Oh, I'm glad you did this.  I found it too, but lost focus and didn't put it in!

PR #1251, opened 2018-04-14
Copy cdn broker custom headers docs from broker repo.

jmcarp:
h/t @cmc333333


content/docs/services/cdn-route.md
```diff
@@ -209,9 +209,23 @@ cf create-service cdn-route cdn-route my-cdn-route \
...
 
-If you need a different configuration, contact [cloud.gov support](/help/).
+CloudFront forwards a [limited set of headers](http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/RequestAndResponseBehaviorCustomOrigin.html#request-custom-headers-behavior) by default. If you want extra headers forwarded to your origin, you'll want to add another parameter. Here we forward both the `User-Agent` and `Referer` headers:
+
+```bash
+$ cf create-service cdn-route cdn-route my-cdn-route \
+    -c '{"domain": "my.domain.gov", "headers": ["User-Agent", "Referer"]}'
+```
+
+CloudFront can forward up to 10 custom headers. Because this broker automatically forwards the `Host` header when not using a [custom origin](#custom-origins), you can whitelist up to nin headers by default; if using a custom origin, you can whitelist up to 10 headers. If you want to exceed this limit or forward all headers, you can use a wildcard:
```
mogul: s/nin/nine/

mogul (changes requested):

content/docs/services/cdn-route.md
```diff
@@ -209,9 +209,23 @@ cf create-service cdn-route cdn-route my-cdn-route \
     -c '{"domain": "my.example.gov", "cookies": false}'
 ```
 
-Other headers, such as HTTP auth, are stripped by default.
```
brittag, line 237: Is it still true that HTTP auth headers are stripped by default? If so, that one still seems like a good one to call out specifically, since it can mean that people's sites are unexpectedly missing Basic Auth that they intended to have (if I'm understanding this correctly).

brittag: @jmcarp This PR still seems really useful to me - would you be up for taking another look at clarifying this bit of it? It might not be necessary, I'm just not sure.

mogul (approved):

PR #1255, opened 2018-04-24
[WIP] Add sales docs

femmebot:


content/overview/pricing/service-level-agreement.md
```diff
@@ -0,0 +1,19 @@
...
+  overview:
+    parent: pricing
+title: Service level agreement
+weight: 120
+aliases:
+
+  - /overview/pricing/service-level-agreement
+  - /overview/overview/service-level-agreement
+---
+Lorem ipsum dolor sit amet.
```
mogul, line 12: Maybe use some https://doggoipsum.com/ instead? ;)

content/overview/pricing/purchasing.md
```diff
@@ -0,0 +1,32 @@
...
+  overview:
+    parent: pricing
+title: Purchasing cloud.gov
+weight: 100
+aliases:
+  - /overview/pricing/how-to-purchase
+  - /overview/overview/how-to-purchase
+---
+
+cloud.gov services can only be purchased by federal government agencies through an IAA. We hope to provide services to state and local agencies in the future.
```
mogul, line 12: @brittag should we be explicit here about the fact that vendors can use it if an agency is the one paying for it?

content/overview/pricing/service-level-agreement.md
femmebot: Definitely! Woof!

brittag:
Some of this would still be useful, but I think we would need to split this work into smaller PR chunks and rethink this a bit, so I'll close this up.

PR #1256, opened 2018-04-26
Fix relref problems

mogul:
Starting with Hugo 0.32.0, "index.md" is reserved for [Page Bundles](https://gohugo.io/content-management/page-bundles/) (so you can include referenced files in the same directory as the referring file and easily move them around together as a unit).

When you have a file called "index.md" in a directory it essentially "eats" all references to other files in that directory so they won't resolve. This caused us to start seeing errors about relrefs a while back, but it previously didn't break the render. In a recent Hugo release, these errors became render-breaking problems.

This change stops us from hitting that problem by using `_index.md` instead of `index.md` for those filenames.

I'm not sure why I had to fully change the name in the services directory to `intro.md`, but `_index.md` was not enough of a change there... Maybe because there are other files in that directory...? :shrug: Have to read more about Page Bundles to figure out why it worked. In any case, this fixes the relref errors.


wslack:
👏 (good to note for our hugo use)

brittag (approved):
It works locally, thank you Bret!!!!

PR #1260, opened 2018-04-27
[WIP] Pricing example

femmebot:
pricing-example 2x


mogul:
I like it, but shouldn't we put the `Estimate your monthly cost` section between `Not included` and `Professional services`? It feels like it's helping explain the cost model, but I feel like the ordering isn't conveying just how much you're getting for that money; we're asking them to look at math and start comparing to other estimates they may have from IaaS providers before we make the "total cost of ownership" argument. 

Speaking of which... Anyone have ideas about how to convey "you will have a radically shortened path to ATO" compactly in here, or do we just trust that they're reading other pages to understand that benefit?

mogul (approved):

PR #1262, opened 2018-05-01
[WIP] Document approaches for custom domains.

jmcarp:
This gathers some thoughts on custom domains, how we manage them now, and other options we've used or considered.


timothy-spencer (approved):
Definitely a good start.  I wouldn't mind some implementation details for what we have now, though maybe it's better to keep those in the repos that implement them?

timothy-spencer (approved):
LGTM

PR #1263, opened 2018-05-07
Move volume services page into services list

brittag:
The volume services service is available for people to use for prototyping/sandbox apps if they request it, so I suggest we make the documentation more visible, to encourage people to try it out if it may be helpful to them. (All the other features in the "experimental" section aren't available yet to use.)

Needs review: I don't know where to find the actual metadata that the service provides for itself in the CF CLI marketplace, so the details here should be checked against that info. (And we should include a link to that info in the `volume-services.toml` file.)

### DIFF
--- a/data/services/volume-services.toml
+++ b/data/services/volume-services.toml
@@ -1,0 +1,5 @@
+# Service metadata lives
+# ?
+name = "volume-services"
+description = "Provision persistent filesystem volumes"
+status = "Experimental"
--- a/data/services/volume-services.toml
+++ b/data/services/volume-services.toml
@@ -1,5 +1,6 @@
 # Service metadata lives
-# ?
+# https://github.com/cloudfoundry/nfsbroker/blob/master/nfsbroker/nfsbroker.go#L91-L125 (We only expose the first service and plan.)
+# https://github.com/18F/cg-deploy-volume-services/blob/master/ci/manifest.yml#L6 (name)
 name = "volume-services"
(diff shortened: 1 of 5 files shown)


mogul:
@jmcarp can you check against the broker metadata?

data/services/volume-services.toml
```diff
@@ -0,0 +1,5 @@
+# Service metadata lives
```
jmcarp, line 1: Metadata lives at https://github.com/cloudfoundry/nfsbroker/blob/7b95addb11f6f71fd207f9ed69a935b4de97ba34/nfsbroker/nfsbroker.go#L91-L125. We only expose the first service and plan.

brittag: @jmcarp I'm not sure how to read this - what's the service name for the first one? It seems to just say `b.static.ServiceName`. And to verify, the description we should list in the docs (to match the marketplace) would be `Existing NFSv3 volumes (see: https://code.cloudfoundry.org/nfs-volume-release/)`?

jmcarp: Er, sorry, the service name comes from another place, for some reason: https://github.com/18F/cg-deploy-volume-services/blob/7fbc82688c7d73930ff113db5a8de0886f73ff80/ci/manifest.yml#L6. So the service name is `nfs`, the plan is `Existing`, and you got the description.

mogul (approved):

PR #1267, opened 2018-05-16
Ask people interested in Windows to email us

brittag:
On the page where we explain that we don't support Windows, encourage readers to email us if they're interested in Windows. This may help us gauge interest.

### DIFF
--- a/content/docs/apps/frameworks.md
+++ b/content/docs/apps/frameworks.md
@@ -28,6 +28,8 @@ You can use a custom buildpack to use other languages. See [custom buildpacks]({
 
 cloud.gov cannot run applications that use .NET Framework, or application binaries that require access to Microsoft Windows kernel or system APIs.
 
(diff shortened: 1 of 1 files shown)


mogul (approved):
It could use a default subject line in the mailto link, but is a marked improvement regardless.

brittag:
@mogul I'm inclined to leave the subject line blank so that people have to write their own, so that we don't have a hundred emails titled "cloud.gov inquiry" in the same thread in email clients. I don't have a strong opinion about it, that's just a benefit to leaving it blank.

mogul:
Fine by me, we may just end up with a hundred emails with a blank subject line instead. ;)

khandelwal:
If we have 100 inquiry emails we'll be so excited, we won't care about the blank subject lines. Besides, we can just make @brittag read them all.

PR #1268, opened 2018-05-18
Secrets design documentation

jmcarp:
This is an experiment in planning features based on https://github.com/concourse/rfcs. The goal is to track the design process in git and github instead of or in addition to one-off meetings. By the time we merge this patch, we should have agreed to a plan for implementing Credhub (or possibly something else). Feel free to comment on diffs, push more commits, or whatever.


content/docs/ops/design/secrets.md
```diff
@@ -0,0 +1,48 @@
...
+    * Questions
+        * Topology
+            * Single shared Credhub per AWS account
+            * Colocate Credhub on each BOSH Director / Concourse ATC; shared database
+            * Separate Credhub and database for each BOSH Director / Concourse ATC
+        * Secret paths
+            * BOSH paths: `/$director/$deployment/$secret` or absolute path
+            * Concourse paths: `/$team/$pipeline/$secret`; `$team/$secret`; no absolute path
+                * https://github.com/concourse/atc/pull/273
+            * For secrets shared between BOSH and Concourse
```
timothy-spencer: Isn't there also an "get https://github.com/concourse/atc/pull/273 merged so that everybody can just use full paths for all secrets" option?

rogeruiz:
Discussion with CH devs. https://cloudfoundry.slack.com/archives/C3EN0BFC0/p1532358425000099

rogeruiz:
oh, I totally forgot to include this madness I was using to hash things out with @jontours! I wonder if it'd be more helpful to include some Scuttle diagrams with these architecture strategies?

screen shot 2018-07-23 at 7 41 21 pm

content/docs/ops/design/secrets.md
rogeruiz: yeah, sadly getting Concourse to actually merge ~~something @jmcarp writes into the codebase~~ that feature is probably never going to happen.

timothy-spencer:
This all looks good to me.   I like the idea of being able to start with cutting over concourse.

rogeruiz:
@timothy-spencer I don't be believe we'll be able to do this with Concourse at all. There are issues with related to leveraging Concourse and BOSH.

sharms (approved):

PR #1270, opened 2018-05-23
Document database restore

jontours:
Fairly straightforward instructions so far.  Only for UAA is it somewhat a grey area as to what should be restored.

### DIFF
--- a/content/docs/ops/runbook/restoring-rds.md
+++ b/content/docs/ops/runbook/restoring-rds.md
@@ -51,6 +51,15 @@ Refer to the [RDS documentation](https://docs.aws.amazon.com/AmazonRDS/latest/Us
 
 For the step that specifies choosing the database to restore, this database will be the respective database we found in the steps above.
 
+Take note to carefully set options matching the configuration of the target database.  Specifically pay attention to these options:
+
+1) Database publicly accessible from the internet (yes/no)
+2) Instance size (m4_large, etc.)
+3) Multi-zone (yes/no)
+
+After the database is finished restoring you can then set the security group, ensure it matches the group of the previous configuration.  To do do this click 'Modify' for the database and find the entry in the form to set the security group (there will be a drop down of all groups to choose from).
+
+
 ### CloudFoundry Applications
 The database restore process will create a new instance with the same credentials.  Create a user provided service with the credentials:
 ```sh
@@ -71,10 +80,10 @@ terraform state rm module.stack.module.base.module.rds.aws_db_instance.rds_datab
 terraform import module.stack.module.base.module.rds.aws_db_instance.rds_database my-restored-db-id
 ```
 
-where my-restored-db-id is the database instance defined during the restore procedure.
+where `my-restored-db-id` is the database instance defined during the restore procedure.
 
 Now run the plan job pipeline in concourse for the respective database. If Concourse is not available you will need to run `terraform plan` and `terraform apply` from the command-line (use the `terraform-apply.sh` script in the cg-pipeline-tasks [repository](https://github.com/18F/cg-pipeline-tasks.git) for these commands).
 
-If the diff from Terraform shows no changes for the plan, you are now safe to apply the changes.
+If the diff from Terraform shows no changes for the plan, you are now safe to apply the changes by running the bootstrap pipeline for the proper environment.  The diff will likely show that some options are changed (parameter groups, password, etc.), terraform makes a best effort to approximate the effect of the plan.  Once apply is ran, these options will be set properly, verify this in the output in the `create-update-ENVIRONMENT` job within the bootstrap pipeline.
 
 Now redeploy the respective application and verify proper operation.
--- a/content/docs/ops/runbook/restoring-rds.md
+++ b/content/docs/ops/runbook/restoring-rds.md
@@ -59,7 +59,7 @@ cf create-user-provided-service database-restore -p '{"db_name": "DATABASE_NAME"
 
 The methods described at [Relational Databases](https://cloud.gov/docs/services/relational-database/) can be used to connect to the restoration copy and restore data as needed.
 
-### Bosh + Concourse Database Restore Procedure
+### Platform Applications Database Restore Procedure
 For these deployments we must perform a Terraform import to restore the database.  First follow the steps above to restore the database in question.
 
 Now run `terraform init` and `terraform import` in the directory for the respective Terraform module.  For development Bosh, eg run these commands:
--- a/content/docs/ops/runbook/restoring-rds.md
+++ b/content/docs/ops/runbook/restoring-rds.md
@@ -8,48 +8,53 @@ title: Restoring RDS
 ---
 
 ## Introduction
-
 Many of the major services for the cloud.gov platform are backed by Postgres databases provided by RDS.  In the case of outages or corruption, it is imperative an operator be able to quickly find the affected database and restore it from a snapshot if possible.  These instructions are augment the documenation from the official RDS documentation.
 
-### Find database identifier
+## Find database identifier
 
-If the database is a tenant application or any other Cloud Foundry hosted application, coordinate with the tenant to find out the application's Organization and Space.  Then perform the following:
+### CloudFoundry Applications
+Coordinate with the tenant, informing them that a restore is going to take place:
+1. Shared database plans do not include backups and restores
+2. Dedicated RDS plans include 1 day of backups
+3. The restore will result in data lost after the point of the most recent snapshot
+4. This restore needs to be initiated before the day passes, otherwise new snapshots will be written
 
+Once the tenant consents, identify their Organization and Space.  Then perform the following:
 ```sh
 cf target -s SPACE -o ORGANIZATION
 ```
 
-Then print environment variables for the affected application:
-
+Print environment variables for the affected application:
 ```sh
 cf env APP
 ```
 
-Within the resulting JSON, find the "host" name within the credentials map.  This will be the database identifier needed below for restoration.
+The resulting JSON contains the "host" name within the credentials map.  This will be the database identifier needed below for restoration.
 
-
-If the restore is a Concourse application, we need to find out the database information from the Terraform state file.  This can be found in S3:
-
+### Platform Applications
+Databases created by Terraform store their credentials in a Terraform State Bucket.  This can be found in S3:
 ```sh
-aws s3 cp s3://terraform-state/cf-staging/terraform.tfstate tmp/state.file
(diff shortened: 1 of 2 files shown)


jontours:
This is ready for review.  If there are some steps that I'm missing let me know.  I seem to remember when there was the Tock DB corruption that after we performed the restore, we then informed the Tock team to point their app to a new database?  Or something?  That doesn't seem to be necessary from the RDS documentation but wouldn't be surprised if there are wrinkles there.

brittag:
Thanks @jontours! This looks like a good start; I copyedited it a bit. But is it enough information for a platform operator to be able to confidently restore a backup for Cloud Foundry, BOSH, Concourse, or UAA data? After you restore the snapshot, what do you need to do to ensure that the system can use it? Do you need to do anything with credentials for the database? I remember we had to work this out in the [contingency plan exercise](https://docs.google.com/document/d/15u70uk8r_zNMMR8bYHWumNIzSLskwvBHNsEhYiYsaW8/edit#), but I don't remember what the answer was. Looks like a platform squad member should review this next.

timothy-spencer:
Re: tock db corruption

I just restored a db from a snapshot before the sadness happened, told them what hostname to use, and they were able to log in and do some sort of schema/data extraction that fixed the problem themselves.

timothy-spencer (commented):
I would like to echo Britta's request for a bit more documentation on how to handle recovering the databases for the service (bosh/atc, maybe domains broker db) up into a state where they are being used by the services that they serve.

I'm not entirely sure what that process is, but perhaps it could be doable to sorta one-off a terraform run like what they do here:  https://medium.com/@vankhoa011/how-i-use-terraform-to-restore-the-latest-snapshot-from-productions-db-to-staging-s-db-aws-rds-6ad4f6620df2

By "one-off", I mean that maybe you could figure out a way to do a one-off run of the terraform stuff in cg-provision where the stack had the proper "snapshot_identifier" in it so that it would launch a new db to replace the (presumably missing) one and pull in the snapshot from the old db.  You might need to push the resulting state file up to the tfstate bucket after that.  Not 100% sure.

Either that, or your procedure could be to restore from snapshot/PITR, then dump/load the data into the stuff created by cg-provision.

timothy-spencer:
Not sure if it helps, but it looks like the uaa stuff is in the bosh db, so if you restore the bosh db properly, uaa should be recovered as well.

sharms:
Hi folks - the scope of this story is to document our current process.  Lets open up a new ticket for planned changes or improvements and not block on this one.

sharms (dismissed):

timothy-spencer:
I'm confused.  This document is supposed to contain our current process, agreed.  But our current process includes being able to restore our service from backups, no?  If that's not true, then OK, but it was my understanding that this was the core of what we were trying to document.

mogul:
Per review this morning, we want to include a bit at the end here about what to do with the restored DB: import it into terraform, then adjust the secrets to point to it.

timothy-spencer:
Should I be trying to add this stuff to the process, or should I just wait for updated content to review?

jontours:
I'm currently working on adding documentation on using concourse to perform a terraform refresh for the final step of updating the environment for a restored DB.  Will probably be opening up a PR in `cg-provision` today.

content/docs/ops/runbook/restoring-rds.md
```diff
@@ -0,0 +1,80 @@
...
+
+### CloudFoundry Applications
+The database restore process will create a new instance with the same credentials.  Create a user provided service with the credentials:
+```sh
+cf create-user-provided-service database-restore -p '{"db_name": "DATABASE_NAME", "host": "DATABASE_HOST", "password":"DATABASE_PASSWORD", "port": "5432", "uri": "postgres://DATABASE_USERNAME:DATABASE_PASSWORD@DATABASE_HOST:5432/DATABASE_NAME", "username": "DATABASE_USERNAME"}'
+```
+
+The methods described at [Relational Databases](https://cloud.gov/docs/services/relational-database/) can be used to connect to the restoration copy and restore data as needed.
+
+### Bosh + Concourse Database Restore Procedure
```
jmcarp: This applies to more than bosh and concourse--this is the process for any platform rds instance. Maybe something like "Platform applications"?

content/docs/ops/runbook/restoring-rds.md
```diff
@@ -0,0 +1,80 @@
...
+```sh
+cd terraform/stacks/main
+terraform init -backend=true -backend-config=encrypt=true -backend-config=bucket=terraform-state -backend-config=key=development/terraform.tfstate
+terraform state rm module.stack.module.base.module.rds.aws_db_instance.rds_database
+terraform import module.stack.module.base.module.rds.aws_db_instance.rds_database my-restored-db-id
+```
+
+where my-restored-db-id is the database instance defined during the restore procedure.
+
+Now run the plan job pipeline in concourse for the respective database. If Concourse is not available (due to the DB being unavailable) you will need to run `terraform plan` from the command line.
```
jmcarp: Concourse might be unavailable for other reasons. WDYT about dropping the parenthetical here?

content/docs/ops/runbook/restoring-rds.md
```diff
@@ -0,0 +1,80 @@
...
+menu:
+  docs:
+    parent: runbook
+layout: ops
+
+title: Restoring RDS
+---
+
+## Introduction
+Many of the major services for the cloud.gov platform are backed by Postgres databases provided by RDS.  In the case of outages or corruption, it is imperative an operator be able to quickly find the affected database and restore it from a snapshot if possible.  These instructions are augment the documenation from the official RDS documentation.
```
jmcarp: "These instructions are augment" should maybe be "These instructions augment". And "documenation" should be "documentation".

content/docs/ops/runbook/restoring-rds.md
```diff
@@ -0,0 +1,80 @@
...
+```sh
+cat tmp/sample.file | grep staging_concourse_rds_host | less
+```
+
+The value displayed is the database to restore.  To restore the BOSH instnace grep for `bosh_rds_host_curr`.
+
+NOTE: These values can also be found in the AWS RDS console.  This can be faster.
+
+## Restore the database
+Refer to the [RDS documentation](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_RestoreFromSnapshot.html) for restoring a database.
```
jmcarp: We might want to add a note about instance types, networks, security groups, etc. We should make sure the new instance lives in the same subnet group, has the same security groups, etc., as the original.

content/docs/ops/runbook/restoring-rds.md
```diff
@@ -0,0 +1,80 @@
...
+
+For the step that specifies choosing the database to restore, this database will be the respective database we found in the steps above.
+
+### CloudFoundry Applications
+The database restore process will create a new instance with the same credentials.  Create a user provided service with the credentials:
+```sh
+cf create-user-provided-service database-restore -p '{"db_name": "DATABASE_NAME", "host": "DATABASE_HOST", "password":"DATABASE_PASSWORD", "port": "5432", "uri": "postgres://DATABASE_USERNAME:DATABASE_PASSWORD@DATABASE_HOST:5432/DATABASE_NAME", "username": "DATABASE_USERNAME"}'
```
jmcarp: If the new instance has the same credential as the old one, do we need to send the credentials to the user again?

content/docs/ops/runbook/restoring-rds.md
timothy-spencer: I think good things, myself.

content/docs/ops/runbook/restoring-rds.md
jontours: I didn't make this edit but can remove it.

sharms: This specific section was how we did it previously, it may not apply now

content/docs/ops/runbook/restoring-rds.md
```diff
@@ -0,0 +1,89 @@
...
+## Restore the database
+Refer to the [RDS documentation](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_RestoreFromSnapshot.html) for restoring a database.
+
+For the step that specifies choosing the database to restore, this database will be the respective database we found in the steps above.
+
+Take note to carefully set options matching the configuration of the target database.  Specifically pay attention to these options:
+
+1) Database publicly accessible from the internet (yes/no)
+2) Instance size (m4_large, etc.)
+3) Multi-zone (yes/no)
```
timothy-spencer: I really think that VPC is super important to call out here too.

content/docs/ops/runbook/restoring-rds.md
jontours: Well, it may actually.  At the very least a tenant will have to `update-user-provided-service`.  But if you run update, according to CF docs, any parameters not provided in the update will be deleted.  So I think these instructions are valid.

content/docs/ops/runbook/restoring-rds.md
jontours: Instructions added taking note of these settings.

content/docs/ops/runbook/restoring-rds.md
jontours: It is and it's noted in the following paragraph.  Remember we couldn't set it until after restoring the snapshot?

content/docs/ops/runbook/restoring-rds.md
jmcarp: Why will tenants need to update the service?

jontours: In the case that the host name changes.  For example I thought the Tock team had to do an update, but I think you could do the restore in a way that the host name doesn't change so I can remove these steps.

jmcarp: But the tenant won't/shouldn't be using the user-provided service to provide credentials to applications. AFAICT the user-provided service is just a secure way for us to send credentials to users so we don't have to send them over email or something.

content/docs/ops/runbook/restoring-rds.md
timothy-spencer: VPC could (and should) be set, but the security groups could not be set upon restore.  The security groups are what needed to be updated, not the VPC.

jontours: You're right!  added.

jontours:
If there are any further edits requested let me know, otherwise I believe this is done.  Also verified the formatting looks good by deploying locally.

timothy-spencer (dismissed):
Yay!

brittag (approved):
This looks sufficient to me, thanks everyone!

PR #1279, opened 2018-06-01
Update rds versions and backup retention period.

jmcarp:
As of https://github.com/18F/aws-broker/pull/46, newly created databases will keep two weeks of backups. We should probably update existing instances to increase the backup retention period for consistency.


mogul (approved):
Is the update of existing DBs to have the same retention period a white noise task or do we need to track/plan for that separately?

LinuxBozo (approved):

jmcarp:
It was white noise, and now it's done.

PR #1293, opened 2018-06-25
Link to example UAA client

pburkholder:


brittag:
Did we work out for sure whether `cg-fake-uaa` is part of our responsibility scope? I'm hesitating on reviewing this because I'm not sure about that.

PR #1295, opened 2018-06-29
Update docs for ending UAA user session

rogeruiz:
This came up during the penetration testing of Tock for its ATO. This is something that folks using the UAA server for authentication need to be made aware of in order to ensure that UAA sessions are cleared at the Login Server level and not just locally to their application.

See 18f/tock#810 for more information on what this documentation is trying to solve.
See 18F/tock#814 for more information on how this UAA endpoint is leveraged.


brittag:
Thanks Roger! This looks good to me from a content perspective - just put in a couple of minor edits. It should get a technical check from another platform person before merge.

sharms (approved):

PR #1302, opened 2018-07-12
Consistently use the default description instead of leaving empty meta tag empty

mogul:


abbeykos (approved):
Excellent! Thanks much, @mogul 🥇

PR #1315, opened 2018-08-06
Initial mfa migration proposal.

jmcarp:
This is basically a brain dump before I'm off the team so that we understand how to migrate from shibboleth to uaa once the necessary features are ready.


rogeruiz (approved):
:shipit:

PR #1316, opened 2018-08-07
Document reset-totp script.

jmcarp:
We can also script the cf-related pieces--this is just what I had time for between other interrupt tasks today.


rogeruiz (approved):
:chef-kissing-fingers: :shipit: less docs!

PR #1318, opened 2018-08-07
Set expectations for DB creation, and test for readiness

pburkholder:


pburkholder:
FWIW, the Oracle DB I used for this example took less than 20m to provision.

sharms (approved):

PR #1320, opened 2018-08-08
The link itself does not work and does not add value to this step

sharms:


jontours (approved):
Beat me to it!

timothy-spencer (commented):
How about we change the link to https://raw.githubusercontent.com/18F/cg-scripts/master/cloudtrail-check.sh?  That script saved me hours of paging through cloudtrail logs.

sharms:
Hi Tim,

I think automating this process is a great idea.  I went ahead and reviewed the script and there were quite a few items that I didn't feel comfortable with as a operator, is it possible you can review this further?

1. The link itself is broken
2. This if statement doesn't confirm it is `terraform`.  It just means that some system with a instance profile did perform these actions, which was not the intent: https://github.com/18F/cg-scripts/blob/master/cloudtrail-check.sh#L66
3. `EVENTNAMES` restricts the set of possible events monitored.  If a operator used the CloudWatch tooling to review, they would see additional events which are excluded from the output of this script
5. Script does not check errors and the `RETURN` value can indicate success when it really is a failure

timothy-spencer:
1)  The link is indeed broken.  I'm not sure how that slipped past the review.  I'd be happy to update the link if you like, but I was thinking that you might want a chance to say "thanks for letting me know where it was", plus replacing your removal with the original + the proper URL would clearly step on what you were trying to accomplish, so I wanted to be respectful of that.
2) The script implements the exact criteria that https://cloud.gov/docs/ops/maintenance-list/#review-aws-cloudtrail-events specifies for both what is/is not terraform and what eventnames need to be audited.  If you think that this should be changed, then that's where you should focus your energies.
3) See 2
4) The RETURN value was aimed at the future pipeline errand/task/whatever that was going to turn red and alert when stuff needed an operator to review it, not to indicate that there was an error.  

Truthfully, I think that the concourse integration which this is aiming at is extra effort that would not be as useful as people think.  I find it easier to run a script rather than clicky-clicky on the UI, and using fly has no advantage over just running the script.  That said, I planned on setting all that up on my next support rotation, and hope to get some time to do this for you folks after my struggle with ES index mapping and types is over.

sharms:
2, 3 and 4 we still need to address. 

I think I understand the miscommunication for 2/3. The process says: _**"All events in the following EventNames should be attributed to Terraform"**_ .  

It looks like maybe you were thinking that meant those were the only events to monitor.  Is this phrasing clearer? _**"Terraform generates the following event types, and if you can attribute them to a `cg-provision` change can be safely ignored"**_.  

We definitely need to monitor additional events if this was misunderstood or miscommunicated.

timothy-spencer:
Maybe you are misunderstanding the code, then, because I do pull down _all_ of the events for those terraform event types specified in https://cloud.gov/docs/ops/maintenance-list/#review-aws-cloudtrail-events and evaluate them all against the rules specified there too, and if they do not match the rules, then they cannot be attributed to cg-provision, and thus they get printed out for the operator to audit.

Believe me, it works.  It found Josh fiddling with domain-broker stuff by hand while doing development, and did not find all of the terraform related changes that the domain-broker stuff was doing.  If I reverse the logic, I get tons of terraform-related events to look at.  If you run it right now, you can see Adam and you doing manual changes to get the isolation segment VPN and so on going.  Or at least that's what it looks like.

sharms:
Could you explain how this code detects suspicious activity from any system (of which there are many) with a instance profile compared to Terraform?

**Current behavior (https://github.com/18F/cg-scripts/blob/master/cloudtrail-check.sh#L66):**
1. `ASIA` check (STS or Instance Profile) which means anyone who uses STS access keys will be classified as terraform, or anyone who leverages the system instance credentials
2. Checks IPs and if they are anywhere in `96.127` or `52.222` classify them as terraform

This means a attacker would not be detected if they use a system with instance profiles (again, there are many), of which the manual review accomplishes.

timothy-spencer:
Those are the specific checks that the auditing algorithm on https://cloud.gov/docs/ops/maintenance-list/#review-aws-cloudtrail-events tells us to use.  If they are incorrect, then that's something that perhaps you could clarify with @brittag, but this is supposed to be the canonical place where we document what our compliance folks require us to do.  

You can see it find activity which needs to be audited if you vault up and run it right now.

It is definitely true that if somebody logs into one of our terraform-enabled hosts and runs terraform, or an API client that claims it's terraform, this will not detect it.  But neither will somebody trying to read all those messages by hand.  And that's not what our audit is set up to find.  Our audit is set up to try to find out of the ordinary changes, like changes from non-temporary keys (not ASIA keys) or from non-govcloud IPs (96.127 and 52.222 hosts), to use your examples.

pburkholder:
Maybe :raccoon: this to over to the cg-scripts repository?

sharms:
It sounds like in an attempt to be operator friendly the process lost important details:

1. **"Terraform Enabled Hosts"** - Terraform runs from a concourse container image with a tag of `iaas`, which assigns the containers to `iaas-worker` instances.  There is no unique "terraform" identifier and terraform does not "claim to be terraform" to the API. 

There are **50** other IAM roles / instance profiles also in use, which the process excludes due to the way the check is implemented (the docs should explain this better)

2. **STS credentials**: STS does not indicate Terraform, and many processes or operators can use STS from GovCloud.  STS should not be confused with Terraform.

3. **GovCloud IPs**: All of the traffic when logged into our systems originates from GovCloud.  This includes customer traffic.

I think I am comfortable saying that you might run this script **in addition** to following the manual process, and we should update the manual process steps to avoid confusion along with adding additional helper scripts to reduce operator burden.

rogeruiz:
hey @sharms & @timothy-spencer,

Thanks for bringing this up @sharms. We shouldn't have bad/wrong documentation, especially for support/interrupt when we have folks new to it working on it for the first time. Let's move fast and change things by iterating on them properly. @timothy-spencer your work isn't being removed from [18F/cg-scripts](https://github.com/18f/cg-scripts) with this work. It's only the documentation that points to it. 

As the person who merged this in originally, along with the original script as well, I think it's a good idea to step back and refocus these efforts. @sharms has mentioned a few times here that the documentation is clearly lacking. Thank you @timothy-spencer for the effort of improving a process which surfaced these deficiencies in the current process. @sharms mentions that an operator should be able to see any issues _which the manual review accomplishes_ by having operators comb through logs. I see he also thinks that having your script alongside the manual process really is the best of both worlds here. 

Let's add the automation to Concourse here to make things easier and I propose we break apart your script into a group of scripts that do one thing really well. That way we can take the most value out of your work and fix the issues that came up here in the other parts.

I'm going to approve this PR, and we can bring this conversation over to the original PR adding the script to improve it. https://github.com/18F/cg-scripts/pull/73

rogeruiz (approved):
:shipit:

timothy-spencer:
The manual process on that page is exactly the same process as what the script does.

It's definitely true that we could look at more things.  There are lots of cloudtrail logs that we could be staring at.  But this is the set that our compliance people have read and approved.  And as a security person, I'd have to say that this is a pretty good set of things to look at too, because it's focused, well defined, and actionable.  It's not going to catch everything, but that's why we have many other controls.

If you want to spend the time to add to that list and make it better, great.  Until then, consider using the script (or even better, adding to it when you come up with these things).  It's certainly not required, but it will save you time if you use it, and it will keep us in compliance with our continuous monitoring duties.

timothy-spencer:
Apologies for responding there.  I did not see the updates to this until I hit submit.

rogeruiz:
No worries @timothy-spencer. In order to not lose context, I think the issue was that the dichotomy of `Automatic` versus `Manual` made it seem like these two processes were equivalent. But you're saying their not. I think in the future as a part of this documentation, we put something along the lines of:

--- 

## Scripts

These scripts will facilitate some of the manual steps below. Please ensure you're running both the scripts and the manual steps outlined below.

## Manual process

PR #1321, opened 2018-08-10
Add operator documentation for private egress feature.

jmcarp:
* Document current operator steps to configure private egress
* Document work left to go from POC to production


LinuxBozo (dismissed):
I think this is good for now. No need to put effort into remaining technical work until need be. Also don't want it getting lost in the backlog, so this is a great place for a reminder when the time comes.

LinuxBozo (approved):

PR #1322, opened 2018-08-14
Explicitly list supported IPsec endpoints

mogul:
The list comes from AWS' docs:
https://docs.aws.amazon.com/AmazonVPC/latest/NetworkAdminGuide/Introduction.html#DevicesTested

(I refrained from going into as much detail as is listed here: https://aws.amazon.com/vpc/faqs/#Connectivity)

### DIFF
--- a/content/docs/apps/experimental/private-egress.md
+++ b/content/docs/apps/experimental/private-egress.md
@@ -57,11 +57,9 @@ graph TB;
   customer-vpn-endpoint---private-service
 {{< /diagrams >}}
 
-**Note**: At this time, only IKEv1 key exchange is supported, so make sure that your VPN endpoint supports this.
-
 To use this service, you must have an internet-routable IP address to use as the IPsec endpoint for the connection. If you have a firewall in place between the internet and your endpoint, then you will have to open both ingress and egress on UDP port 500 and ESP (IP Protocol 50) to enable the connection. If you are also using NAT behind your firewall, you will also have to enable UDP port 4500.
 
-Your IPsec endpoint can be either a software or hardware device. We are prepared to support the following options (in addition to open source IPsec endpoints such as strongSWAN and OpenSwan):
+To be compatible with this feature, your IPsec endpoint must support IKEv1 key exchange along with AES128 encryption, SHA256 hashing, and Diffie Hellman Group 14 (MODP 2048) for Perfect Forward Secrecy (PFS). Your IPsec endpoint can be either a software or hardware device. Here is a list of compatible options we are prepared to support (in addition to open source IPsec endpoints such as strongSWAN and OpenSwan):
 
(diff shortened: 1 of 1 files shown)


content/docs/apps/experimental/private-egress.md
```diff
@@ -59,9 +59,30 @@ graph TB;
...
 
-To use this service, you must have an internet-routable IP address to use as the endpoint for the connection. If you have a firewall in place between the internet and your endpoint, then you will have to open both ingress and egress on UDP port 500 and ESP (IP Protocol 50) to enable the connection. If you are also using NAT behind your firewall, you will also have to enable UDP port 4500.
-
-### How to set up a dedicated pool of hosts connected to your organization
+To use this service, you must have an internet-routable IP address to use as the IPsec endpoint for the connection. If you have a firewall in place between the internet and your endpoint, then you will have to open both ingress and egress on UDP port 500 and ESP (IP Protocol 50) to enable the connection. If you are also using NAT behind your firewall, you will also have to enable UDP port 4500.
+
+Your IPsec endpoint can be either a software or hardware device. We are prepared to support the following options (in addition to open source IPsec endpoints such as strongSWAN and OpenSwan):
+
+<!-- List from https://docs.aws.amazon.com/AmazonVPC/latest/NetworkAdminGuide/Introduction.html#DevicesTested -->
```
jmcarp: Any reason not to include this link also, since it's maybe more authoritative than the list we're copying from it?

mogul: I don't want to expose any IaaS-specific details if we don't have to, so linking directly to AWS' docs, particularly in the midst of a very complex section, seems inadvisable.

LinuxBozo: Then there really is no reason to include this list. Better would be to say that we support IKEv1, with AES128 encryption, SHA256 hashing, and Diffie Hellman Group 14 (MODP 2048) for Perfect Forward Secrecy (PFS) and that your endpoint has to do the same.

mogul: Oh, I meant I don't want to expose IaaS-specific details on *our* side... I don't mind putting out a specific list so that customers can see a device that they identify with. 

I'll add those parameters, though, @LinuxBozo!

LinuxBozo: I still wouldn't include any information about the specific implementations so there is less upkeep to these docs as versions change, or products are retired.

mogul: How strongly do you feel about that? Hint: Phrase your answer in the form of a radio selection on the `Review Changes` UI. ;)

LinuxBozo: Not strongly at all, but much like "you broke it, you bought it", "you wrote it, you update it" ;-)

mogul: Still looking for that `Review changes` form submit...

LinuxBozo (approved):

PR #1324, opened 2018-08-14
Make unfurls page-specific

mogul:
This branch causes Hugo to use the auto-generated page summary for the meta description tag instead of the sitewide description. This makes unfurls in Slack, Twitter, etc. include information specific to the particular URL that's referenced in a post.

There's a trade-off here, in that if a page doesn't contain `<!--more-->` Hugo will slurp up to 70 words into the summary, which might go too far into the text. Let's try it and see how it feels! 

For references, see:
https://api.slack.com/docs/message-link-unfurling#classic_unfurling
https://gohugo.io/content-management/summaries/


eddietejeda:
I think having descriptive headers, even if it's the first 70 words of the page, is better than having the same information.

Lets merge.

bengerman13 (approved):
LGTM

PR #1334, opened 2018-09-07
Document s3 policy change

jontours:
This documents the change made to s3 broker.
 - essentially when ever `cf delete-service S3-SERVICE` is ran the bucket(s) are cleared and the service deletes automatically.  PR for broker can be found here: https://github.com/cloudfoundry-community/s3-broker/pull/33 .
 - has been tested on staging.


rogeruiz (dismissed):
:shipit:

jontours:
After talking with @rogeruiz we are going to create a new plan for this and associate the sandbox accounts with that plan.

sharms (approved):

PR #1339, opened 2018-10-02
Examples S3 and bound services and CUPS

pburkholder:


pburkholder:
Any interest still in these S3 examples:

- Example using bucket provisioned outside of cloud.gov
- Putting and getting an object with the cloud.gov managed S3 bucket:

If so I can rebase and try again

mogul:
@pburkholder I haven't seen anyone call for using AWS's CLI to copy files from an S3 bucket into an application, rather than having the application directly reference the S3 bucket content through APIs... What's the driver for adding this?

pburkholder:
@mogul and I discussed and we're okay with closing this unmerged.

PR #1341, opened 2018-10-09
+files

abbeykos:
adding new one-pager


mogul:
@abbeykos how does one access this from elsewhere on the site? Or is this just intended to be a URL we can link to in emailed comms?

abbeykos:
Sorry, @mogul, just now saw this (weird, I must have missed it...). For now this is just the file available at a URL that can be easily shared. We still need to figure out how to integrate it into the site.

mogul (approved):

PR #1348, opened 2018-10-30
Create 2018-10-30-release-notes.md

abbeykos:
Would someone mind taking a look to spot any errors in my Markdown?


mogul (approved):
Reviewed along with Britta. We had to fix the front matter and a few links; sorry I didn't do that as a separate commit before also working on content edits! We think this is good to go.

PR #1351, opened 2018-10-31
Reference CloudCheckr as a potential list of improvements

mogul:


jessykacastillo-zz (approved):
Britta and I reviewed this and agree this is a good addition.

PR #1357, opened 2018-11-29
NogoDB

mogul:
Removing explicit mention here. I kind of feel like I need to leave some pointer to a normative thing about what to do instead of Mongo, but we don't have such a page yet... Probably should! 🤷‍♂️


rogeruiz (approved):
:shipit:

PR #1363, opened 2018-12-13
Update link to FedRAMP conmon strategy guide

pburkholder:


mogul (dismissed):

mogul:
@sharms (since you're on interrupt): I'm looking in Concourse and at a loss as to how to figure out which pipeline and which build this was so that I understand why `concourse-ci/status` didn't report back. Can you show me how to find and fix this so I can handle problems like this when they happen future?

mogul:
Maybe @jontours can help now that he's on interrupt?

mogul (approved):

PR #1371, opened 2018-12-21
Remove hidden character

mogul:
There was some non-rendering unicode character here that was hard to see. Thanks for the report @wslack !


ccostino (approved):
Thank you!

PR #1386, opened 2019-02-04
IP filtering

pburkholder:
I've removed a lot of the explanation of how route services work to
  instead point to a demo that works with cloud.gov.

  I've worked with alphagov to get the code updates and example merged in
  so it works just great with cloud.gov.


mogul (approved):

pburkholder:
@mogul Thanks for the 👍 but I don't have permissions to merge, so that's on y'all.

PR #1387, opened 2019-02-08
Make ATO inheritance guidance more accurate

brittag:
In this change, I made clear that the diagram of control inheritance is an example instead of a guarantee, and I moved the Control Implementation Summary download right next to it.

Being very precise about this is important because people with a lot of compliance experience simply won't take us seriously if we appear to be exaggerating the control inheritance that we offer (even accidentally).

I also removed a few sentences that didn't seem necessary.

### DIFF
--- a/content/overview/security/fedramp-tracker.md
+++ b/content/overview/security/fedramp-tracker.md
@@ -27,7 +27,7 @@ For DoD teams: the Defense Information Systems Agency (DISA) categorizes [FedRAM
 
 Your agency still needs to grant your system an Authority to Operate, but FedRAMP has done the labor-intensive work of reviewing cloud.gov's security posture and endorsed it, which reduces the compliance work you need to do. Your agency's authorizing official can request the P-ATO documentation package from FedRAMP and accept that endorsement for your own system. See [ATO process]({{< relref "docs/compliance/ato-process.md" >}}) for the typical workflow.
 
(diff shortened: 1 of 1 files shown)


content/overview/security/fedramp-tracker.md
```diff
@@ -25,18 +25,16 @@ For DoD teams: the Defense Information Systems Agency (DISA) categorizes [FedRAM
...
 
-![Graph showing the breakdown of how many controls are fully covered by cloud.gov.](/img/fedramp-moderate-controls.png)
-
-Here's how it works: Every "moderate" impact federal system is required to account for a baseline of 325 controls before it can be granted an ATO. Your agency may choose to do more than that, but once cloud.gov's P-ATO is reviewed and accepted, 269 of those requirements are already implemented and documented. Of the remaining requirements, responsibility for 41 of them is shared between cloud.gov and your application, and 15 are fully yours.
+Here's how it works: Every "moderate" impact federal system is required to account for a baseline of about 325 controls before it can be granted an ATO. Once cloud.gov's P-ATO is reviewed and accepted, many of those requirements are already implemented and documented. Of the remaining requirements, responsibility for most of the rest are shared between cloud.gov and your application, and only a few are fully yours.
```
mogul: "only a few" is probably implying we reduce the number more than we do. How about "only some" which is a little more ambiguous?

mogul (changes requested):
Other than this, looks good!

content/overview/security/fedramp-tracker.md
brittag: Yes that makes sense, done!

mogul (approved):

PR #1392, opened 2019-02-19
[WIP] February release notes

mogul:
Needs an image fixed, and we want to move at least a couple paragraphs out into a separate doc on service brokers.


abbeykos:
👍 Thank you!

abbeykos (approved):

mogul:
Argh, @abbeykos this was still WIP because of this piece:
> Needs an image fixed, and we want to move at least a couple paragraphs out into a separate doc on service brokers. I will open a separate PR to fix that; keep your eyes open for that as you'll need to make a similar change in the mailed bulletin!

abbeykos:
Apologies -- I thought we were done! I'll look for the PR.

PR #1395, opened 2019-02-22
Docker tweak: restage the app, not the service

rosskarchner:
I ran into this while trying this example


mogul (approved):

mogul:
Thanks for catching that and following through with a submitted PR to fix it! 🥇

PR #1401, opened 2019-03-05
direct users to `routes` instead of `urls`

bengerman13:
`cf push` outputs `routes` but the quickstart calls out `urls`. This
updates the quickstart to match the command output.


rogeruiz (approved):
:shipit:

PR #1404, opened 2019-03-08
document troubleshooting bosh master

bengerman13:
add documentation for how to connect to bosh master when a bosh deployment fails.
Relies on 18f/cg-deploy-bosh#336


jontours (approved):
Thanks @bengerman13!

PR #1409, opened 2019-03-14
For private egress, give more direct instructions and remove confusing bit of diagram

brittag:
Our JAB folks need us to make clear that an isolated segment is required (not optional) for setting up the network connection. This tries to make that clear by giving more direct recommendations (that you should not connect your agency network to all of cloud.gov) and removing the shared segment from the diagram. We shouldn't merge without review from @mogul.


brittag:
This is the revised diagram:

Screen Shot 2019-03-14 at 2 26 51 PM

mogul (approved):

PR #1410, opened 2019-03-15
For private egress page, add information on customer responsibilities

brittag:
Our JAB folks would like us to be clear to customers about their responsibilities for managing the connection appropriately, so I added a section about this. I also made the support ticket instructions less redundant. This should get a review from @mogul.


mogul (approved):
:chef-kissing-fingers:

PR #1414, opened 2019-04-03
Xenial is 16.04, not 18.04!

mogul:
:doh:

### DIFF
--- a/content/updates/2019-02-26-quarterly-update.md
+++ b/content/updates/2019-02-26-quarterly-update.md
@@ -9,13 +9,13 @@ It’s a new year, and we’ve got a pile of changes and highlights to tell you 
 
 ## Announcements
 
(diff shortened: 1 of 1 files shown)


bengerman13 (changes requested):
`cflinuxfs3` is Bionic Beaver/18.04

mogul:
Oh whoa, you're right! Wow, now I'm so confused. So: cflinuxfs3 is based on 18.04, and the underlying VMs are based on 16.04... How does that work if something in a runC app instance (with cflinuxfs3/Bionic/18.04) tries to do something with a kernel ABI that doesn't exist in 16.04?? Maybe the Xenial/16.04 is keeping up with modern Bionic kernel releases...?

mogul:
Never mind, [I took the question upstream](https://cloudfoundry.slack.com/archives/C033RE5D6/p1554350553063400).

PR #1416, opened 2019-04-05
added asg reference.

siennathesane:
removed the custom route service documentation in favour of asgs.

Signed-off-by: Mike Lloyd <mike.lloyd@gsa.gov>

### DIFF
--- a/content/docs/apps/experimental/restricting-users-to-trusted-ip-ranges.md
+++ b/content/docs/apps/experimental/restricting-users-to-trusted-ip-ranges.md
@@ -29,9 +29,9 @@ Application Security Groups (ASGs) support two types of scoping:
 * Platform
 * Space
 
-These scopes allow both operators and developers to have more fine grained access control. Platform-scoped ASGs are applied at the platform level, and would apply to all tenants of cloud.gov. The cloud.gov team does not apply platform-scoped ASGs as they can interrupt tenant operations unless they are needed to meet platform compliance requirements.
+These scopes allow both operators and developers to have fine-grained access control. Platform-scoped ASGs are applied at the platform level, and would apply to all tenants of cloud.gov. The cloud.gov team does not apply platform-scoped ASGs as they can interrupt tenant operations unless they are needed to meet platform compliance requirements.
 
-Space-scoped ASGs are specific to a given tenant's space within their allotted organization. This allows for tenants to create Space-specific security groups that apply to all applications within a given space. To apply a space-scoped ASG to a single application, create a new space, assign the ASG, and then deploy the application.
+Space-scoped ASGs are specific to a given tenant's space within their own organization. Tenants can create space-specific security groups that apply to all applications within a given space. To apply a space-scoped ASG to a single application, create a new space, assign the ASG, and then deploy the application in that space.
 
 ### Defaults
 
--- a/content/docs/apps/experimental/restricting-users-to-trusted-ip-ranges.md
+++ b/content/docs/apps/experimental/restricting-users-to-trusted-ip-ranges.md
@@ -20,7 +20,7 @@ To fully limit access to your apps and services, you must ensure that all users 
 
 ## Restricting access to your own applications
 
-Limiting or restricting access for both ingress and egress traffic from your applications is fairly straightforward through a feature called [Application Security Groups](https://docs.cloudfoundry.org/concepts/asg.html), or ASGs. ASGs support white- and black-listing network traffic from given IPv4 ranges. For IPv6 and other more complex restriction cases, you can leverage a proxy application in a sidecar deployment pattern. 
+Limiting or restricting access for both ingress and egress traffic from your applications is fairly straightforward through a feature called [Application Security Groups](https://docs.cloudfoundry.org/concepts/asg.html), or ASGs. ASGs support white- and black-listing network traffic from given IPv4 ranges. For IPv6 and other more complex restriction cases, you can leverage a proxy application in a sidecar deployment pattern. 18F publishes an example proxy application with Nginx, which you can find [here](https://github.com/18f/boardproxy).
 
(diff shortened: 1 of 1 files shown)


content/docs/apps/experimental/restricting-users-to-trusted-ip-ranges.md
```diff
@@ -20,5 +20,60 @@ To fully limit access to your apps and services, you must ensure that all users
 
 ## Restricting access to your own applications
 
-To limit requests to your own applications, you can modify your application logic directly. Alternatively, you can [can follow this example](https://github.com/alphagov/re-paas-ip-safelist-service#quick-demo) within cloud.gov to [create a route service](https://docs.cloudfoundry.org/services/route-services.html#introduction) 
-to block all traffic that does not originate from a trusted IP address range.
+Limiting or restricting access for both ingress and egress traffic from your applications is fairly straightforward through a feature called [Application Security Groups](https://docs.cloudfoundry.org/concepts/asg.html), or ASGs. ASGs support white- and black-listing network traffic from given IPv4 ranges.
+
+### Scoping
```
mogul, line 25: I'd cut this section on scoping. It's not very relevant to customers, and they can read about it in the CF docs if they need to.

content/docs/apps/experimental/restricting-users-to-trusted-ip-ranges.md
```diff
@@ -20,5 +20,60 @@ To fully limit access to your apps and services, you must ensure that all users
...
+
+Application Security Groups (ASGs) support two types of scoping:
+
+* Platform
+* Space
+
+These scopes allow both operators and developers to have more fine grained access control. Platform-scoped ASGs are applied at the platform level, and would apply to all tenants of cloud.gov. The cloud.gov team does not apply platform-scoped ASGs as they can interrupt tenant operations unless they are needed to meet platform compliance requirements.
+
+Space-scoped ASGs are specific to a given tenant's Space within their allotted organization. This allows for tenants to create Space-specific security groups that apply to all applications within a given space. To apply a Space-scoped ASG to a single application, create a new Space, assign the ASG, and then deploy the application.
```
mogul: ...except this line about space-scoped ASGs, keep this one! Also s/Space/space/.

content/docs/apps/experimental/restricting-users-to-trusted-ip-ranges.md
```diff
@@ -20,5 +20,60 @@ To fully limit access to your apps and services, you must ensure that all users
 
 ## Restricting access to your own applications
 
-To limit requests to your own applications, you can modify your application logic directly. Alternatively, you can [can follow this example](https://github.com/alphagov/re-paas-ip-safelist-service#quick-demo) within cloud.gov to [create a route service](https://docs.cloudfoundry.org/services/route-services.html#introduction) 
-to block all traffic that does not originate from a trusted IP address range.
+Limiting or restricting access for both ingress and egress traffic from your applications is fairly straightforward through a feature called [Application Security Groups](https://docs.cloudfoundry.org/concepts/asg.html), or ASGs. ASGs support white- and black-listing network traffic from given IPv4 ranges.
```
mogul: I'd rework this paragraph to talk about the two cases... "If you only need to restrict to IPv4 ranges, use ASGs. For IPv6 and other more complex restriction cases (eg limiting particular kinds of requests), you can use a proxy app". Then you can add an anchored section heading at the bottom which gives them a pointer to making a front-end proxy app with the nginx-buildpack.

I believe Peter wrote an example of doing this, btw, so we might even have an something to point them to! See [Slack discussion](https://gsa-tts.slack.com/archives/C0DF6LAMQ/p1553803321013900).

content/docs/apps/experimental/restricting-users-to-trusted-ip-ranges.md
```diff
@@ -20,5 +20,60 @@ To fully limit access to your apps and services, you must ensure that all users
...
+These scopes allow both operators and developers to have more fine grained access control. Platform-scoped ASGs are applied at the platform level, and would apply to all tenants of cloud.gov. The cloud.gov team does not apply platform-scoped ASGs as they can interrupt tenant operations unless they are needed to meet platform compliance requirements.
+
+Space-scoped ASGs are specific to a given tenant's Space within their allotted organization. This allows for tenants to create Space-specific security groups that apply to all applications within a given space. To apply a Space-scoped ASG to a single application, create a new Space, assign the ASG, and then deploy the application.
+
+### Defaults
+
+cloud.gov is preconfigured with two ASGs: `public_networks` and `dns`. These ASGs are applied by default to all containers in your deployment.
+
+* `public_networks`: This group allows access to public networks, and blocks access to private networks and link-local addresses. Cloud Foundry blocks outgoing traffic to the following IP address ranges by specifically allowing traffic to all other addresses.
+
```
mogul: Believe it or not, compliance rules forbid us to publish internal network CIDRs in docs even when they're private ranges. 😞 So better to just say that private IP ranges are blocked and let our readers look up what private ranges are, removing the following lines.

mogul (changes requested):
Sorry for the hassle, want to be thorough and unfortunately less-thorough due to compliance here. @pburkholder is your Trello proxy published anywhere that we can link to?

content/docs/apps/experimental/restricting-users-to-trusted-ip-ranges.md
siennathesane: I only cover scoping so I can specify that we do not deploy platform-scoped ASGs - I think it's worth leaving in there. I did want to specify that we will not accept requests for platform-scoped policies but that seems a bit aggressive for the docs.

mogul (dismissed):

mogul (approved):

PR #1422, opened 2019-04-17
Remove docs around using HTTP-01 validation for LetsEncrypt

bengerman13:
HTTP-01 is currently broken, and not tested, so remove docs about using it

### DIFF
--- a/content/docs/services/cdn-route.md
+++ b/content/docs/services/cdn-route.md
@@ -88,7 +88,6 @@ The maximum number of domains that can be associated with a single cdn-route ser
 
 ### How to set up DNS
 
-**Note:** If you are creating a new site on cloud.gov or you are migrating an existing site to cloud.gov that can tolerate a small amount of downtime during the migration, you can skip the first step and proceed directly to [Create CNAME record(s)](#step-2-create-cname-record-s)
 
 #### Step 1: Create TXT record(s)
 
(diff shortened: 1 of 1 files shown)


siennathesane:
Let's reference that it's currently in a broken state and not actively tested instead of taking it down completely because we need to ensure we have incentive to fix it.

bengerman13:
@mxplusb sounds good. Updated with your recommendation.

content/docs/services/cdn-route.md
```diff
@@ -88,7 +88,10 @@ The maximum number of domains that can be associated with a single cdn-route ser
 
 ### How to set up DNS
 
-**Note:** If you are creating a new site on cloud.gov or you are migrating an existing site to cloud.gov that can tolerate a small amount of downtime during the migration, you can skip the first step and proceed directly to [Create CNAME record(s)](#step-2-create-cname-record-s)
+**Note:** Due to changes in how CloudFront processes requests, we currently only offer certificate provisioning via DNS challenges. This means that users can no longer skip step 1 below.
+We are investigating ways to bring this feature back.
```
siennathesane: Can you add something like this:

> If you have a use case where DNS configurations are not available for you, please reach out to us at support@cloud.gov.

It's a good opportunity for us to find out how many users really use the HTTP-01 challenges.

siennathesane (changes requested):
Sorry for being nitpicky.

content/docs/services/cdn-route.md
bengerman13: updated.

siennathesane (approved):
LGTM

PR #1429, opened 2019-05-17
bump mermaid dependency to make graphs render again

bengerman13:


mogul (approved):
Some diagrams lost their scrollability, but at least they're rendering again now!

PR #1433, opened 2019-05-30
Adding note about answering RFQ questions to contact info

brittag:
For review by @jessykacastillo as part of our response procedure for questions from companies.


jessykacastillo-zz (commented):
Looks good. Can we write in the word 'proposal' somewhere in here?

mogul:
We should tell them where to post their question. How about in https://github.com/18F/cg-product/issues ?

jessykacastillo-zz (approved):

PR #1439, opened 2019-06-25
Add some guidelines around leaking information during totp resets

bengerman13:
I think the way we're *actually* doing this currently is a potential weakpoint in our security, so I threw some a note in to hopefully remind people of this. 

This is equal parts suggested change in behavior and the corresponding documentation change.


brittag (approved):

brittag:
Thanks! Yeah, I believe it's a good practice to minimize unnecessary information during this process.

PR #1440, opened 2019-06-26
Update language about local governments

eddietejeda:
While we should continue to look into supporting local governments, this is not a priority and should not be advertised.


siennathesane (approved):
LGTM.

hillaryj (approved):

PR #1449, opened 2019-07-16
Fix version number mention in release notes

brittag:
We ended up with a missing sentence half instead of the right version number.


bengerman13 (approved):
🚀

PR #1451, opened 2019-07-31
Clean up out of date language and other messaging adjustments

eddietejeda:
This pull request updates the messaging 

- [x] Removed confusing and out of date pricing from the services page
- [x] Instead of hiding the sandbox accounts inside content pages, it is now a top level menu item
- [x] Mark elasticsearch 2.4 as deprecated
- [x] Split the Low/Moderate pricing options into their own columns to make it easier to talk about the differences
- [x] The "updates" section was not aligned properly and overflowed into the navigation. This center aligned the content and makes the layout more consistent


content/docs/services/cdn-route.md
```diff
@@ -20,9 +20,9 @@ If you don't need CDN caching, use the [custom domain service]({{< relref "docs/
 
 ## Plans
 
-Plan Name | Description | Price
+Plan Name | Description | 
 --------- | ----------- | -----
```
bengerman13, line 24: you need to remove the dashes on line 24 to finish removing the last column on all the tables you removed a column from.

RonWilliams (approved):

PR #1459, opened 2019-08-26
elasticsearch56 plan name example command fix

ondrae:
The elasticsearch56 plan name appears to have changed from `medium` to `medium-ha`. It was changed in one place on the page, but not everywhere. I was following instructions from a project that was still using `medium` and was receiving this error `The plan medium could not be found for service elasticsearch56`. I came here to find the official plan names and noticed the inconsistencies on the page.

I'm very new to cloud.gov, so please double check.

### DIFF
--- a/content/docs/services/elasticsearch56.md
+++ b/content/docs/services/elasticsearch56.md
@@ -24,7 +24,7 @@ Plan Name | Description |
 To create a service instance run the following command:
 
 ```sh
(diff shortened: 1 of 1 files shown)


eddietejeda:
This is a good catch. We just [announced changes](https://cloud.gov/updates/2019-08-26-changes-to-cloud-gov-services-and-prices/) and updated the site, but looks like we missed this.

eddietejeda (approved):

PR #1467, opened 2019-08-28
Some guidance on using Oracle DB

pburkholder:
Replaces #1319, as I don't have branch push access.


siennathesane (changes requested):
Looks great, I like how comprehensive it is. Can you please update the service access steps to include using Cloud Foundry Service Keys? That's the preferred way of accessing services.

https://docs.cloudfoundry.org/devguide/services/service-keys.html

pburkholder:
Paging @eddietejeda -- good to go?

pburkholder:
@mxplusb As suggested, switched to service keys.

siennathesane (approved):
LGTM.

PR #1471, opened 2019-09-05
use "memory quota" instead of "usage"

eddietejeda:
- to provide more consistent and clearer language, usage is now referred to as memory quota


RonWilliams (approved):
LGTM

PR #1473, opened 2019-09-09
add deprecation policy

bengerman13:
Kinda bare-bones.

things I specifically wonder about:
- is this the best place to put this? I couldn't find anywhere better
- would a concrete example be clearer?
- did/should we define how frequently we send out reminders?


siennathesane:
> would a concrete example be clearer?

I think so. I would use something fictional to prevent concern.

> did/should we define how frequently we send out reminders?

Yup!

siennathesane (approved):
lgtm

PR #1492, opened 2019-10-24
add Customer service objectives

eddietejeda:


RonWilliams (approved):
Matches agreed language. Ready for use.

PR #1504, opened 2019-12-11
update compute limit docs to remove references to purchase units and …

bengerman13:
…punt the hard work to cf docs

The current docs explain CPU shares incorrectly, and reference old pricing models no longer in use. 
This points users to the CF docs for CPU share info, and removes references to the old pricing model.


siennathesane (approved):
lgtm.

PR #1507, opened 2019-12-13
Direct readers to consider the OMB memo re DNSSEC

pburkholder:
The language in OMB 18-23 is squishy, so we shouldn't directly interpret the meaning, but leave that as an exercise for the reader.


eddietejeda (approved):

pburkholder:
This already got merged, it seems.

PR #1515, opened 2020-01-02
Tighten language, finesse "control", add ciphers 

pburkholder:
Since were getting questions about Netsparker and the
CBC ciphers beeing flagged as weak, want to make clear
they're support by NIST.

Also, tweaked the language so it doesn't sound like
we're not in control, but have delegated to a respected
and trusted 3rd party.


pburkholder:
Keep this open while I resolve the NetSparker Qs with our Federalist customers.

pburkholder:
Obsolete, replaced by #1620

PR #1523, opened 2020-01-22
Explain log service creation

LindsayYoung:
The docs for the log store in the cloud foundry docs are not the same as what we need for cloud.gov. Thank you Steve Greenberg, for explaining!


eddietejeda:
Merged in #1641

PR #1533, opened 2020-02-04
Add Federalist support

eddietejeda:
Ensure that cg-site builds in Federalist environment.

## Security Considerations
None... may improve security since federalist is actively managed and maintained.


content/docs/apps/managing-teammates.md
```diff
@@ -94,7 +94,7 @@ On the Organization page, scroll down to the “Organization users” section. T
 
 To remove a teammate from an organization, you must first remove them from any of the organization's spaces first.
 
-![App Overview dashboard](/img/remove-organization.png)
+![App Overview dashboard]({{ .Site.BaseURL }}/img/remove-organization.png)
```
bengerman13, line 97: why not relative links? (here and all the places below)

eddietejeda: Federalist requires baseurl to be set, since the URL structure for preview branches is different than production URLs.

More info: https://federalist.18f.gov/documentation/previews/

bengerman13: aha - I was guessing the opposite would be the case. TIL.

bengerman13 (approved):

PR #1574, opened 2020-02-28
updated command.

siennathesane:
- update the forced reallocation command to be copy & pasteable.
- and added instructions so it's understandable.

## security considerations

None, just updating a command.

Signed-off-by: Mike Lloyd <mike.lloyd@gsa.gov>


bengerman13 (approved):
Thanks, I hate it.

PR #1619, opened 2020-04-15
Notify FedRAMP for non-security outages

pburkholder:

- Notify FedRAMP for non-security outages

Hold pending confirmation from FedRAMP...

## security considerations

Nope


bengerman13:
@pburkholder is this ready to ship?

PR #1664, opened 2020-05-11
Share how we enforce MFA for AWS

pburkholder:

- Demonstrate to world (and auditors) that we use MFA for all interactions
- Document how so others can learn

## Security Considerations

Policy snippet is open-source, so that's okay. Saying how we have good standards it okay to share.


_docs/ops/secrets.md
```diff
@@ -52,7 +76,8 @@ echo 'region = us-gov-west-1' >> ~/.aws/config
 echo "mfa_serial = $mfa_serial" >> ~/.aws/config
 ```
```
ccostino, line 77: @pburkholder I realize this isn't a part of your changes and something I had modified with the previous update, but I wonder if we should modify it still, especially after seeing Jonathan Pirro's demonstration using `aws-vault` for login.gov.  This will only configure a profile for the operator's production GovCloud account, correct?

Does it make sense to call that out and leave it as an exercise for the operator to determine for their remaining accounts, or expand on this a bit more with more detailed instructions?

ccostino (commented):
Thank you for updating this, @pburkholder!  The only thing I wanted to check on isn't related to your changes.  However, given the restructuring and addition of new details here I wanted to confirm if it makes sense to adjust things a bit further based on what we've learned over the past week.

_docs/ops/secrets.md
ccostino: @pburkholder and I caught up about this, we'll address it separately from this work!

ccostino (approved):
All good now, thanks @pburkholder!

PR #1699, opened 2020-05-28
Suggest that forks don't go to the cloud-gov org

pburkholder:
See also https://gsa-tts.slack.com/archives/C0ENP71UG/p1590679069401100

-
-
-

## Security Considerations


onelittlebecca:
so are any of the patches reusable? would it be useful to have a repo or folder where we store patches or are majority standalone?

pburkholder:
@onelittlebecca The example I'm thinking of is stratos. We had a fork of it in our ~repo~ cloud-gov org to fix the SSO_WHITELIST feature, but we don't deploy that fork, we _always_ deploy the upstream fork.

Anyhow, when the auditors were checking for old credentials in repositories, there were some really old commits in that repo that got flagged.

So if we're forking a repo just for a minor fix, and we're never going to deploy that fork to our environment, then it shouldn't be in our GitHub org, it can just go to one of our personal accounts.  E.g. I fork stuff into @pburkholder when making a contribution upstream. 

How could I adjust my language to make that clearer?

bengerman13 (approved):

PR #1702, opened 2020-06-03
Update database backup and restoration guides

ccostino:
Resolves https://github.com/cloud-gov/product/issues/1381 - I'm flagging the team for review due to the multiple touch points on the site that are customer and internal facing, but I'd appreciate 👀 especially from @pburkholder and at least one operator.

This changeset updates both the customer documentation and our operator runbook for database backups and restores with AWS RDS.  It clearly lays out how the backup and restoration process works, what information is needed, and how to perform the steps required to complete the work.

Includes work from PR https://github.com/cloud-gov/cg-site/pull/1692 from @apburnes that details how to perform the steps in AWS RDS and verify the restore worked properly.

For customers, there is expanded information on how to connect directly to a database service and perform manual backups for local restores.

- Update customer documentation for database backups and restores
- Expand customer documentation for connecting to database services directly and performing manual backups
- Update cloud.gov operator runbook with clearer steps for performing customer database restores

:sunglasses: [PREVIEW URL - Customer Documentation](https://cg-88d42ca6-59d7-47e0-9500-4dd9251360b9.app.cloud.gov/preview/cloud-gov/cg-site/update-db-restore-documentation/docs/services/relational-database/)
:sunglasses: [PREVIEW URL - Operator Runbook](https://cg-88d42ca6-59d7-47e0-9500-4dd9251360b9.app.cloud.gov/preview/cloud-gov/cg-site/update-db-restore-documentation/docs/ops/runbook/restoring-rds/)

## Security Considerations
None; this is public knowledge of existing information that is now consolidated and easier to understand.

However, I did shy away from adding steps on how a customer can perform a database restoration across spaces (environments).  I feel that is a situation better suited for hands-on customer support as there are caveats with copying production data to non-production environments and the fact that there is downtime involved.

I'd rather we educate and walk customers through that process directly first, but I'm not opposed to adding the steps if others feel different (there is at least [one example](https://github.com/fecgov/fec-cms/wiki/Backup-or-Restore-Wagtail-PostgreSQL-database-in-dev,-staging,-or-production) of a customer knowing how to do this already).

### DIFF
--- a/_docs/ops/runbook/restoring-rds.md
+++ b/_docs/ops/runbook/restoring-rds.md
@@ -46,15 +46,47 @@ The JSON containing the environment variables contains the `database identifier`
 Refer to the [RDS documentation](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_RestoreFromSnapshot.html) for restoring a database.
 
 Prior to restoring, record the configuration settings of the existing instance:
+- DB identifier (instance name)
 - Instance size (m4.large, etc.)
 - Multi-zone (yes/no)
 - VPC (dev, staging, etc.)
+- Subnet
 - Security Groups
 - Master password (this is the `password` you saw above in the application environment variables)
 
 Copy all of these values and configure them in the restore form to mirror the configuration of the existing database instance.  For the `identifier` name, provide a name like `<instance name>-<date>-restore` to make the instance easy to find.  Note that the database should never be set to publicly accessible.
 
+Steps for launching the snapshot:
+- Log into the AWS Console and navigate to *RDS* > *Snapshots* and filter by <DB identifier>
+- Select the latest snapshot and click *Actions* > *Restore Snapshot*
+- In *Restore DB Instance*
+  - Under *Settings* > *DB Instance Identifier* enter a new DB identifier ie (`<instance name>-<date>-restore`)
+  - Using the configuration settings recorded prior to restore, fill in the relevant fields with the corresponding values.
+  - Finally, click *Restore DB Instance*
+- Navigate back to *RDS* > *Databases* and filter by the new DB identifier
+- Once *Status* is changed to *Available*, copy the `endpoint` in the database settings we can compare and verify the restored database
+
 Once the restore has finished, confirm the new instance matches the previous configuration.  To update a configuration setting, click 'Modify' from the RDS console instance view and make the required modifications.
+
+Steps for verifying the finished restore:
+- Log into the jumpbox
+- Grab the database credentials (`password`, `address`, `dbuser`...) from `bosh`
+- If possible, log into the database
+  - Connect `psql "postgres://${dbuser}:${password}@${address}:${port}/${name}"`
+  - List all the tables in the database: `psql> \dt+`
+  - Grab record counts for key tables
+    - ie *Cloud Controller DB*: `select count(*) from organizations;`
+    - ie *Cloud Controller DB*: `select count(*) from spaces;`
+    - ie *Cloud Controller DB*: `select count(*) from apps;`
+- Log into the new, restored database
+  - *Note:* Change the `address` from bosh to the `endpoint` copied from the database settings
+  - Connect `psql "postgres://${dbuser}:${password}@${endpoint}:${port}/${name}"`
+  - List all the tables in the database: `psql> \dt+`
+  - Grab record counts for key tables
+    - ie *Cloud Controller DB*: `select count(*) from organizations;`
+    - ie *Cloud Controller DB*: `select count(*) from spaces;`
+    - ie *Cloud Controller DB*: `select count(*) from apps;`
+- Compare tables and record counts between the original and restored databases
 
 When the settings and configuration of the new restored instance are confirmed, rename the original instance with a new qualifier, e.g, `<instance name>-<date>-old`.
 
--- a/_docs/ops/runbook/restoring-rds.md
+++ b/_docs/ops/runbook/restoring-rds.md
@@ -9,7 +9,7 @@ title: Restoring RDS
 
 cloud.gov structured data is stored using Amazon's Relational Database Service (RDS).  RDS provides multiple availability zones, automated backups and snapshots allowing operators to restore the database in the event of a failure.
 
-## Customer Databases
+## Tenant Databases
 
 Coordinate with the tenant to determine the point in time to restore the database from, and when the operation will take place:
 
@@ -18,13 +18,14 @@ Coordinate with the tenant to determine the point in time to restore the databas
 1. The restore will result in the database being restored to a point in time.
 1. Data written after the restore point in time will be lost.
 
-### Obtain information and confirmation from the customer
+### Obtain information and confirmation from the tenant
 
-In order to perform a restore, we need the following information from the customer:
+In order to perform a restore, we need the following information from the tenant:
 
 - The organization name
 - The space they are working within
 - The name of the application(s) connected to the database service they need a restoration performed on
+- Phone numbers and contact information if it's an urgent situation
 
 Be sure to confirm this information and remind them that a restoration may result in a brief period of downtime with database connectivity.  Ask if they have some way of shutting off access/enabling a maintenance mode of the website and if so, they should do so prior to the restoration process starting.
 
@@ -39,7 +40,7 @@ cf target -s SPACE -o ORGANIZATION
 cf env APP
 ```
 
-The JSON containing the environment variables contains the `database identifier` and `password` under the key `host` within the credentials map for the database service (look for `aws-rds`).  The `database identifier` is needed to find the database instance in the AWS RDS instance listing, and the `password` is required for setting the master password in the new instance.
+The JSON containing the environment variables includes the `database identifier` and `password` under the key `host` within the credentials map for the database service (look for `aws-rds`).  The `database identifier` is needed to find the database instance in the AWS RDS instance listing, and the `password` is required for setting the master password in the new instance.
 
 ### Performing the database restore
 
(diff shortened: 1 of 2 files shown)


_docs/services/relational-database.md
```diff
@@ -117,38 +117,44 @@ In short, `cf bind-service` will provide a `DATABASE_URL` environment variable f
...
 ```sh
-$ cf connect-to-service --no-client ${APP_NAME} ${SERVICE_NAME}
+# Find the aws-rds service you are targeting in the services list and make note of the service's name
+$ cf services
+
+# Check to see if a SERVICE_CONNECT service key already exists for the service.
+$ cf service-keys ${SERVICE_NAME}
+
+# If a key exists, remove it first before starting.
+$ cf delete-service-key ${SERVICE_NAME} SERVICE_CONNECT
```
mheadd: Just curious why this step is needed, to delete an existing `SERVICE_CONNECT` service key. Won't the `connect-to-service` plugin just reuse an existing key if it exists, or create. a new one automatically if it doesn't?

_docs/services/relational-database.md
```diff
@@ -117,38 +117,44 @@ In short, `cf bind-service` will provide a `DATABASE_URL` environment variable f
 
 The contents of the `DATABASE_URL` environment variable contain the credentials to access your database. Treat the contents of this and all other environment variables as sensitive.
 
-## Access the data in the database
+## Backups
+
+For shared plans (`shared-psql` and `shared-mysql`), RDS does not back up your data.
 
-There are currently two ways to access the database directly.
+For dedicated plans, RDS automatically retains daily backups for 14 days. These backups are AWS RDS storage volume snapshots, backing up the entire DB instance and not just individual databases. If you need to have a database restored using one of these backups, you can [email support](mailto:support@cloud.gov) with the following information:
```
mheadd: The runbook changes include this: `...coordinate with them on the date and time to perform the restore`. Should we specify in the customer docs a specific time period in advance to reach out for doing the restore? Maybe something like, `minimum of 48 hours in advance` reach out to support..?

_docs/services/relational-database.md
ccostino: Theoretically `connect-to-service` should reuse the existing key and I believe it does, but there's a situation where that key becomes "stale" so to speak.  I think it's if an app is restaged/redeployed, because the plugin requires an app that's bound to the service instance.  If the app isn't the same app instance as what was used before when first creating the connection, `connect-to-service` can't reconcile that and it won't create a new one automatically.

I'll investigate a bit further to see if that's still the case or if my memory isn't serving me correct here!

_docs/services/relational-database.md
ccostino: That's a good point!  Perhaps in normal circumstances we can suggest that, but what if the customer has an immediate issue (a migration gone wrong, a compromised system) that requires attention right away?  There's no other option, correct?

mheadd: That's fair. Just wondering if - outside of an emergency - we should encourage as much advance notice as possible.

ccostino: I think that's still good to note, I'll add that language to the guides!

ccostino: I've update the language here to encourage advance notice when possible; this may be starting to touch on broader policies for the overall role and coverage of support, too, that we may need to have @RonWilliams and @eddietejeda take a look at.

_docs/services/relational-database.md
ccostino: I was slightly mistaken, it isn't the app instance that causes the existing key to become stale, it's deleting the original service the key was bound to and binding a new instance in its place with the same service name.  In other words, if you follow this flow (in its simplest form):

- Connect to existing DB service named `mydb` with the plugin
- Unbind and delete the existing DB service once you're done with it
- Create and bind a new DB service named `mydb` again
- Try to connect to the new DB instance

The reason why this will fail is because the service key will still have the credentials and information of the original service information.  When you provision a new service, a whole new identifier is given and the credentials will be different (this is why we have to be careful in the operator runbook to match everything up when restoring an instance).

For the purposes of this customer documentation as it is written here, you're right, this should not be needed and if anything, only mentioned as a troubleshooting step if necessary.

Thanks for catching and asking, @mheadd!

mheadd (dismissed):
Looks good to me. :high-five:

_docs/ops/runbook/restoring-rds.md
```diff
@@ -10,35 +10,62 @@ title: Restoring RDS
...
+- The name of the application(s) connected to the database service they need a restoration performed on
+
+Be sure to confirm this information and remind them that a restoration may result in a brief period of downtime with database connectivity.  Ask if they have some way of shutting off access/enabling a maintenance mode of the website and if so, they should do so prior to the restoration process starting.
+
+If the customer agrees to proceed given this information, coordinate with them on the date and time to perform the restore and with which PITR (point in time restore) or snapshot to use.  At the agreed upon date and time, continue with the steps below.
+
 ### Identify RDS Hostname
-Once the tenant agreed to a database restore, identify the RDS instance attached to the application:
+
+Once the tenant agreed to a database restore, identify the RDS instance attached to the application using the information they provided:
```
pburkholder: s/agreed/agrees/
s/provided/provide/

_docs/services/relational-database.md
```diff
@@ -117,38 +117,38 @@ In short, `cf bind-service` will provide a `DATABASE_URL` environment variable f
...
+When you do [contact support](mailto:support@cloud.gov) with a database backup request, please include the following information:
 
-### cg-migrate-db plugin
-You can access the data in your database via the `cg-migrate-db`
-plugin. See the [repository](https://github.com/18F/cg-migrate-db)
-for instructions on how to install the plugin, backup data, import data,
-download a local copy of the data, and upload a local copy of the data.
+- Your organization name
+- The space you are working within
+- The name of the application(s) connected to the database service you need a restoration performed on
```
pburkholder, line 134: - phone numbers and contact information if it's an urgent situation.

Do _not_ share any passwords or details of any exploit or compromise. We'll call you if necessary, but we'll never ask you for a password over the phone.

_docs/ops/runbook/restoring-rds.md
```diff
@@ -10,35 +10,62 @@ title: Restoring RDS
...
 Refer to the [RDS documentation](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/USER_RestoreFromSnapshot.html) for restoring a database.
 
-Prior to restoring, record the configuration settings:
-1. Database publicly accessible from the internet (yes/no)
-1. Instance size (m4.large, etc.)
-1. Multi-zone (yes/no)
-1. VPC (dev, staging, etc.)
-1. Security Groups
+Prior to restoring, record the configuration settings of the existing instance:
+- Database publicly accessible from the internet (yes/no)
```
pburkholder: Let's remove this as none of our DBs are publicly accessible.

pburkholder:
I've made a few comments but one thing that concerns me is that we're making an implicit agreement that we'll be available for customer restores into the indefinite future. 

We have a 6-month deprecation policy. Should we add the caveat that our guidance will be rescinded/updated once a self-service option is available, regardless of that policy? Or would we need to keep ourselves available to help with restores for six months after making a self-service option available?

_docs/ops/runbook/restoring-rds.md
ccostino: That's true - I think it's still worth noting to make sure that the DB is not set to publicly accessible, but I removed it from this list and turned it into a note to double check the setting to make sure it's set properly, in case the defaults ever change.

_docs/services/relational-database.md
ccostino: Excellent call - added!

ccostino:
@pburkholder that's exactly why I wanted to check on this. :-)

If we're going to be offering a self-service option in the future, then I think it should be explicitly noted that our guidance will change in the future, and that the instructions will be updated to reflect new steps when the time comes.

I imagine that we would have to keep ourselves available for at least a bit of time once a self-service option is available.  There will be a transition period to account for, and I think we should make sure folks are able to make the switch easily and smoothly.

ccostino:
Thank you for adding in your pieces from https://github.com/cloud-gov/cg-site/pull/1692, @apburnes! :high-five:

ccostino:
@RonWilliams this should be ready for your full review at this point.  Please let us know if you have any questions or concerns!

_docs/ops/runbook/restoring-rds.md
```diff
@@ -9,54 +9,118 @@ title: Restoring RDS
...
+
+When the settings and configuration of the new restored instance are confirmed, rename the original instance with a new qualifier, e.g, `<instance name>-<date>-old`.
+
+Finally, rename the new instance to be just `<instance name>` so it matches what the original instance was, and what the application(s) is bound to.  This ensures that the application will still function properly and connect to the new database instance.
+
+### Follow up and confirm with the tenant
+
+When the restore is completely finished, notify the tenant and ask them to confirm that their application(s) is still functioning properly and that the data is properly restored.  Help them troubleshoot any other issues if there is anything still wrong.
+
+Once we receive confirmation that the restore has been completed successfully, coordinate with the tenant when it is appropriate to remove the old instance, particularly if it is needed for a security audit or forensic analysis.  This is important as we do not want old database instances hanging around, which contribute to extraneous overhead costs.
```
RonWilliams: Can we be more prescriptive here with requiring a specific shutdown date to be coordinated and agreed upon? Even if it's 30-60 days out, it would be helpful to set a reminder for removal.

_docs/services/relational-database.md
```diff
@@ -117,38 +117,43 @@ In short, `cf bind-service` will provide a `DATABASE_URL` environment variable f
 
 The contents of the `DATABASE_URL` environment variable contain the credentials to access your database. Treat the contents of this and all other environment variables as sensitive.
 
-## Access the data in the database
+## Backups
+
+*Please note that these instructions will change in the future as we expand our service offerings and provide more options for customers.*
 
-There are currently two ways to access the database directly.
+For shared plans (`shared-psql` and `shared-mysql`), RDS does not back up your data.
```
RonWilliams: Per conversation with Carlo, can we revise this to "For shared plans (`shared-psql` and `shared-mysql`), cloud.gov and RDS does not back up your data. These are only intended for limited use development and testing instances."

RonWilliams (commented):
A couple minor revisions to consider, but overall good.

_docs/ops/runbook/restoring-rds.md
ccostino: Sure thing!  I'll work this in.

RonWilliams (approved):

ccostino:
Thanks to everyone who took a look and/or worked on this PR! :high-five:

PR #1713, opened 2020-06-16
Standardize authenticator app language, remove Google authenticator

mheadd:
- Remove references to Google Authentication App, since it is less secure than other options
- Standardize language across different pages on use of authenticator apps.

## Security Considerations
- Remove references to references to Google Authentication App as it is less secure than other options.


ccostino (approved):
Thanks, @mheadd!

PR #1718, opened 2020-06-22
add necessary dots to DNS examples

bengerman13:
- add dots so examples look more like what users will see in their DNS controls

## Security Considerations
None


ccostino (approved):
Thanks, @bengerman13!

PR #1724, opened 2020-06-30
add docs for insecure_origin flag

bengerman13:
- add docs for insecure_origin flag

## Security Considerations
None

### DIFF
--- a/_docs/services/external-domain-service.md
+++ b/_docs/services/external-domain-service.md
@@ -53,15 +53,18 @@ Name      | Required   | Description                   | Example                
 
 ### `domain-with-cdn` plan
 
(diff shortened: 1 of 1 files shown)


mheadd:
This is nitpicky, but for the tables that have a `Required` column, should we change the values to `true` or `false`? Or `Yes` and `No`?

bengerman13:
I *think* I like required/optional better for this type of table, so you don't have to look at the table headers to see what it means. 
Not a big issue here, but if this table had many more rows, it's nice to not have to scroll to the top to see what each column means.

mheadd (approved):
Looks good.

PR #1732, opened 2020-07-10
Update to remove HTTP only comment

coobr01:
- Just a .md update because HTTPS is actually available when using the HTTP option to access an s3 bucket  (i.e. outside of using the s3 protocol itself).

## Security Considerations
None.

### DIFF
--- a/_docs/services/s3.md
+++ b/_docs/services/s3.md
@@ -130,9 +130,7 @@ If you plan to enable "website mode" for the bucket and use it that way, you wil
 
 * `http://${BUCKET_NAME}.s3-website-${AWS_DEFAULT_REGION}.amazonaws.com/`
 
-Note that "website mode" URLs don't support HTTPS, and they aren't appropriate for production use unless fronted by a CloudFront distribution.
-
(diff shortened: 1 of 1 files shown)


onelittlebecca (approved):

bengerman13:
@coobr01 thanks for the PR. Has this changed recently? 
 The [docs from AWS](https://docs.aws.amazon.com/AmazonS3/latest/dev/WebsiteHosting.html) indicate https is not available on bucket websites.

coobr01:
Not sure, all I can say is that its working in an s3 service instance that I recently created under my Cloud.gov org/space for my app.

image

bengerman13:
I see - that request is going to the API endpoint `<bucket>.s3-<region>.amazonaws.com`, not the website endpoint `<bucket>.s3-website-<region>.amazonaws.com`. With public objects, like in this bucket, the API will behave pretty similarly to a website, but [with some differences](https://docs.aws.amazon.com/AmazonS3/latest/dev/WebsiteEndpoints.html#WebsiteRestEndpointDiff)

I believe the bucket you're looking at has website hosting disabled, but I tested on a bucket with hosting enabled and I'm not able to connect via HTTPS.

coobr01:
Ok, i would close this PR without merging then, thanks.

bengerman13:
Thanks @coobr01

onelittlebecca:
Good catch -- I missed that <facepalm>

PR #1744, opened 2020-08-10
document external-domain-service error_responses option

bengerman13:
Depends on cloud-gov/external-domain-broker#162
- document external-domain-service error_responses option

:sunglasses:[PREVIEW URL](https://cg-88d42ca6-59d7-47e0-9500-4dd9251360b9.app.cloud.gov/preview/cloud-gov/cg-site/custom-error-page-docs/docs/services/external-domain-service/#options)

## Security Considerations
None

### DIFF
--- a/_docs/services/external-domain-service.md
+++ b/_docs/services/external-domain-service.md
@@ -53,13 +53,14 @@ Name      | Required   | Description                   | Example                
 
 ### `domain-with-cdn` plan
 
-Name              | Required   | Description                           | Example                           |
-------------------|------------|---------------------------------------|-----------------------------------|
-`domains`         | *Required* | Your custom domain or domains         | `my-domain.gov,www.my-domain.gov` |
-`origin`          | optional   | A custom origin to serve from         | `external-app.example.gov`        |
-`insecure_origin` | optional   | Is the custom origin HTTP (not HTTPS) | `true`                            |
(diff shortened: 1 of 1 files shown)


_docs/services/external-domain-service.md
```diff
@@ -81,6 +82,22 @@ and unless you are using a custom origin, we set the `Host` header.
...
+```
+cf create-service external-domain domain-with-cdn -c '{"domains": "example.gov", "error_responses": {"404": "/errors/404.html", "403": "/login.html"}}
+```
+Be careful when setting this for 5xx responses: 5xx responses indicate a server error, and setting a custom error response will increase the load on a potentially unhealthy application.
+
+The default for this setting is `{}`, so errors are passed to the client exactly as the CDN receives them, and you can use the same setting to reset to the default:
+```
+cf create-service external-domain domain-with-cdn -c '{"domains": "example.gov", "error_responses": {}}
+```
+Note that only these error codes can be customized: 400, 403, 404, 405, 414, 416, 500, 501, 502, 503, 504
```
ccostino, line 98: Just noting for anyone else looking that this limitation of which error codes can be customized [comes from CloudFront](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/custom-error-pages.html).

ccostino (approved):
Thanks, @bengerman13!

apburnes (approved):
👍

PR #1746, opened 2020-08-11
Stop referring to vestigial DB name field

bengerman13:
from @mogul :
> As a consequence of https://github.com/cloud-gov/aws-broker/pull/97 , the `db_name` field is vestigial and only present for > backward-compatibility. 
>
> This change makes the docs refer instead to `name`, which is the same attribute name used by other DB brokers out there in the Cloud Foundry community.

## Security Considerations
None


ccostino (approved):
Thanks, @mogul and @bengerman13!

PR #1747, opened 2020-08-11
Update to jekyllv3.9 to support kramdown v2 and add peer dependency

apburnes:
- Fixes kramdown security alert

:sunglasses:[PREVIEW URL](https://cg-88d42ca6-59d7-47e0-9500-4dd9251360b9.app.cloud.gov/preview/cloud-gov/cg-site/update-kramdown-deps)

## Security Considerations
- Addresses dependabot alert


bengerman13:
@apburnes is this ready for review?

apburnes:
@bengerman13 we've also got a security alert on the `a11y` dev dependency which is an archived project that hasn't been touched in two years and we actually do not even run with `npm test`. Thinking about ripping it out or adding `pa11y` for testing alternative.  Thoughts?

bengerman13:
I think it would be best to update them one at a time, so we can tell if one of them breaks everything

bengerman13 (approved):

PR #1776, opened 2020-09-21
Add new CISA directives page for reference

pburkholder:

(OUTDATED: See comments below):

- I've added both a News post and a Compliance docs post.
- I don't think we need to do both. In the past we've only done a StatusPage update. It might be good to have a documented update in our core site for reference.
- Thoughts on which approach is better? Now that I've done both, I (peter) lean to just doing the news post.

:sunglasses:[PREVIEW URL](https://cg-88d42ca6-59d7-47e0-9500-4dd9251360b9.app.cloud.gov/preview/cloud-gov/cg-site/peterb/cisa-eds)

## Security Considerations

Required public comms, not an issue.


ccostino (dismissed):
Thanks, @pburkholder!

I think have a page to capture these for reference is helpful.  Perhaps having this page serving as an archive and the StatusPage.io updates so folks receive a notice of the update is enough in this case?  That seems to make the most sense to me.

ccostino:
Hey @eddietejeda - I approved this PR, wanted to flag you just in case you wanted to see before it goes live.  Thanks!

pburkholder:
I don't think I'm doing this right as it doesn't show up in the News section. 

Anyhow, I'm going to pull the News post, and then it looks like we need to fix the dropdown menus since Compliance isn't listed there.  Stay tuned....

image

eddietejeda (dismissed):

ccostino (dismissed):
This looks good to me and all links are working, thanks @pburkholder!

ccostino (approved):
Good call - thanks @pburkholder!

PR #1793, opened 2020-10-01
update marketplace.fedramp.gov URL format

bengerman13:
- Correct fedramp URL format

## Security Considerations
None


mheadd (approved):
Nothing like a good hashbang URL. Thanks for making this change.

PR #1853, opened 2020-12-23
clarify service account role assignment

spgreenberg:

Clarifies roles assigned to service accounts can be manipulated if necessary.

## Security Considerations

This makes clear that roles can be assigned to service accounts (this is a feature of the platform and therefore it should be documented). On the surface, this may appear risky (for example assigning an OrgManager role), but given that a service account could be created for every space I do not believe there is any more risk calling out this capability.

### DIFF
--- a/_docs/services/cloud-gov-service-account.md
+++ b/_docs/services/cloud-gov-service-account.md
@@ -14,10 +14,12 @@ To set up your application to be deployed with an automated system, you need a d
 
 ## Plans
 
-Plan Name | Description | 
---------- | ----------- | -----
-`space-deployer` | A service account for continuous deployment, initially limited to a single space | 
-`space-auditor` | A service account for auditing configuration and monitoring events, initially limited to a single space | 
+Plan Name | Description | Cloud Foundry Role
+--------- | ----------- | -------------------|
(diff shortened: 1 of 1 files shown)


davemcorwin (commented):
thanks @spgreenberg !
If the `space-auditor` account is handled the same way as the `space-deployer` maybe the language should be the same OR just represented once, as in the "note".

Related but not the intent of this particular PR, the docs don't actually say what the difference is between to 2 types of accounts, it only says for "continuous deployment" or for "auditing configuration and monitoring events". It might be helpful to also explain WHY each one is appropriate for the corresponding situation.

spgreenberg:
@davemcorwin That is great feedback, thank you. I didn't notice that the role is `space-deployer` versus `space-developer`. SpaceDeveloper and SpaceAuditor are both roles in cloud foundry and it would likely be more clear if the nam ing was consistent.

spgreenberg:
> @davemcorwin That is great feedback, thank you. I didn't notice that the role is `space-deployer` versus `space-developer`. SpaceDeveloper and SpaceAuditor are both roles in cloud foundry and it would likely be more clear if the nam ing was consistent.

I have added a clarification. Thanks for the suggestion @davemcorwin!

bengerman13 (approved):

PR #1876, opened 2021-02-23
Update RDS service documentation with new broker details

ccostino:
This changeset updates our RDS service documentation with several new details.

- Updates the table of available offerings with proper formatting and version number updates
- Updates the shared database information to the current versions deployed
- Reformats the create and update sections slightly with improved documentation and instructions
- Adds information about the new version custom parameter
- Adds more details about updating existing instances and when to contact support for assistance

:sunglasses: [PREVIEW URL](https://cg-88d42ca6-59d7-47e0-9500-4dd9251360b9.app.cloud.gov/preview/cloud-gov/cg-site/update-rds-documentation/docs/services/relational-database/)

## Security considerations:
- Includes updates to version information that is already publicly available in AWS' documentation

### DIFF
--- a/_docs/services/relational-database.md
+++ b/_docs/services/relational-database.md
@@ -13,35 +13,33 @@ If your application uses relational databases for storage, you can use the AWS R
 
 ## Plans
 
-Plan Name                | Description                                                                  | Software Version |
----------                | -----------                                                                  | ---------------- |
-`shared-psql`            | Shared PostgreSQL database for prototyping (no sensitive or production data) | 9.5.15           |
-`micro-psql`             | Dedicated micro RDS PostgreSQL DB instance                                   | AWS RDS Latest   |
-`micro-psql-redundant`  | Dedicated redundant micro RDS PostgreSQL DB instance                        | AWS RDS Latest   |
-`small-psql`             | Dedicated small RDS PostgreSQL DB instance                                   | AWS RDS Latest   |
-`small-psql-redundant`  | Dedicated redundant small RDS PostgreSQL DB instance                        | AWS RDS Latest   |
-`medium-psql`            | Dedicated medium RDS PostgreSQL DB instance                                  | AWS RDS Latest   |
-`medium-psql-redundant`  | Dedicated redundant medium RDS PostgreSQL DB instance                        | AWS RDS Latest   |
-`medium-gp-psql`            | Dedicated higher workload medium RDS PostgreSQL DB instance                                  | AWS RDS Latest   |
-`medium-gp-psql-redundant`  | Dedicated higher workload redundant medium RDS PostgreSQL DB instance                        | AWS RDS Latest   |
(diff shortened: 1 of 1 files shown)


_docs/services/relational-database.md
```diff
@@ -13,35 +13,33 @@ If your application uses relational databases for storage, you can use the AWS R
 
 ## Plans
 
-Plan Name                | Description                                                                  | Software Version |
----------                | -----------                                                                  | ---------------- |
-`shared-psql`            | Shared PostgreSQL database for prototyping (no sensitive or production data) | 9.5.15           |
```
bengerman13, line 18: Where did shared go? are we no longer allowing it?

_docs/services/relational-database.md
```diff
@@ -84,6 +73,35 @@ For example, if you wanted to create a new micro PostgreSQL instance called `my-
...
 
+### Setting optional parameters
+
+There are several optional parameters that you can specify when creating a new instance:
+
+Name               | Description                                                    | Default |
+---                | ---                                                            | ---     |
+`storage`          | Number of GB available to the database instance                | 10      |
+`enable_functions` | Boolean to enable functions on supported databases             | false   |
+`version`          | Specifies a supported major version in RDS (must be in quotes) | ""      |
```
bengerman13: maybe instead of `""` we should refer them to the table above? or say aws latest?

_docs/services/relational-database.md
ccostino: Nothing has been formalized yet, but the plan is to phase them out.  I figured this would be a first step in steering folks away from them.  Do you feel we should hold off on removing them from the documentation until that's ironed out some more?

If so, then I'd like to just verify the version number of each in production as a part of the update, because I think the shared DB instances have been updated in recent months.

_docs/services/relational-database.md
ccostino: Good call!

_docs/services/relational-database.md
bengerman13: if we're going to phase them out, we should do so loudly rather than quietly
If we're doing this simply to discourage their use, then I have mixed feelings about this


Maybe leave them in and have a larger discussion tomorrow after standup?

ccostino: Sounds good. :-)

_docs/services/relational-database.md
ccostino: Updated this to match the table above.

bengerman13 (dismissed):

_docs/services/relational-database.md
ccostino: Come to think of it, this wouldn't follow our deprecation policy anyway, so I'll add these back in for now.

onelittlebecca (approved):
This looks really good; you cleaned up a lot of sentences and de-confused them / added clarity.

PR #1923, opened 2021-05-17
Upgrade transitive dependency on outdated version of trim module

mheadd:
- Fixed outdated packages. Bumped Netlify to `v2.10.126`
- Fix transitive dependency on trim module, which has a known vulnerability.
- Modified package.json on to add resolution key
- Use `npm-force-resolutions` 

:sunglasses: [PREVIEW URL](https://cg-88d42ca6-59d7-47e0-9500-4dd9251360b9.app.cloud.gov/preview/cloud-gov/cg-site/fix-trim)

## Security Considerations
Fix outdated dependency

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,8 +5,7 @@
     "build": "bundle exec jekyll build",
     "clean": "bundle exec jekyll clean",
     "reset": "npx rimraf Gemfile.lock package-lock.json LICENSE.md CONTRIBUTING.md degit.json .git",
-    "start": "bundle exec jekyll serve",
-    "preinstall": "npx npm-force-resolutions"
+    "start": "bundle exec jekyll serve"
   },
   "dependencies": {
     "anchor-js": "^4.3.1",
@@ -16,8 +15,5 @@
   },
   "devDependencies": {
     "rimraf": "^3.0.2"
-  },
-  "resolutions": {
-    "trim": ">=0.0.3"
   }
 }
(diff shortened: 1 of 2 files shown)


mheadd:
I lightly tested Netlify, which this update affects. All seems well, but a more thorough eval might be warranted.

mheadd:
Reupping this for @apburnes review. This addresses a vulnerability and I'm addressing it with some changes to `package.json` that I've not tried before, Would appreciate more 👁️ 's on this one.

package.json
```diff
@@ -5,15 +5,19 @@
     "build": "bundle exec jekyll build",
     "clean": "bundle exec jekyll clean",
     "reset": "npx rimraf Gemfile.lock package-lock.json LICENSE.md CONTRIBUTING.md degit.json .git",
-    "start": "bundle exec jekyll serve"
+    "start": "bundle exec jekyll serve",
+    "preinstall": "npx npm-force-resolutions"
```
apburnes: @mheadd I'm not familiar with this package but looks cool.  The only thing that I would be concerned about is the repository dependencies being out of sync from the actual deploy. From their docs: `Then add npm-force-resolutions to the preinstall script so that it patches the package-lock file before every npm install you run` and we run `npm ci` to install the JS dependencies so this may cause the discrepancy between source control and deployed. Can this package be run just once to update the package-lock.json and not use it in a preinstall script?

mheadd: Worth a try

mheadd: OK, that seemed to work. I might kill this PR and rerun the steps so that the only thing that is committed is the modified package-lock file.

mheadd:
Close in favor of #1928

PR #2024, opened 2021-11-22
add security considerations action

bengerman13:
## Changes proposed in this pull request

- add security considerations check.

## security considerations

Adds security considerations check, which should be considered secure.


apburnes:
Closing since security considerations check is already implemented.

PR #2043, opened 2021-12-21
Specify specific region that cloud.gov is deployed to

folksgl:
In order to prevent having to field the same "What region is Cloud.gov deployed to?" question multiple times, specify the region in on the website. If specifying here is undesirable (because there are plans to deploy to multiple regions, or cloud.gov may switch regions) then feel free to comment/close.

- Specify the region that Cloud.gov is deployed to.

## Security Considerations
None - Wording change to website only.

### DIFF
--- a/_docs/technology/iaas.md
+++ b/_docs/technology/iaas.md
@@ -12,7 +12,7 @@ redirect_from:
 
 ## The infrastructure underlying cloud.gov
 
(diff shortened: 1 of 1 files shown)


bengerman13:
Thanks @folksgl. I pulled this into a branch on the cloud-gov fork so the federalist build will run. You can track it at #2044

PR #2052, opened 2022-01-12
Fix S3 service-key script

mogul:
- Fix the example script so that it actually works.
- Note that the script must be "sourced" (run with a '. ' in front of it for maximum cross-shell compatibility).

## Security Considerations

No impact to the system. Corrects a non-functional script in documentation which customers might copy-and-paste.

### DIFF
--- a/_docs/services/s3.md
+++ b/_docs/services/s3.md
@@ -100,17 +100,24 @@ For example, you might want to use the [AWS Command Line Interface](https://aws.
 ```sh
 #!/bin/bash
 
+Run this script with a . in order to set environment variables in your shell
+For example:
+  . ./getcreds.sh
+
+set -e 
+
 SERVICE_INSTANCE_NAME=your-s3-service-instance-name-here
 KEY_NAME=your-service-key-name-here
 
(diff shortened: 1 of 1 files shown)


_docs/services/s3.md
```diff
@@ -100,18 +100,25 @@ For example, you might want to use the [AWS Command Line Interface](https://aws.
 ```sh
 #!/bin/bash
 
+Run this script with a . in order to set environment variables in your shell
```
pburkholder, line 103: For readability, it might be better to use `source` instead of `.`, but OK to ship as far as I'm concerned.

pburkholder (approved):

_docs/services/s3.md
mogul: I was going to do that, but I discovered that only the "." syntax is portable across shells per POSIX spec.

mogul: https://stackoverflow.com/a/20097303/17138235

mogul: https://ss64.com/bash/source.html#:~:text=source%20is%20a%20synonym%20for,available%20after%20the%20script%20completes.

mheadd:
Changes incorporated via #2062

PR #2090, opened 2022-04-12
Update jekyll and uswds deps

apburnes:
## NOTICE:
This deployment has switched from using 18F-Franklin font families to the current USWDS font-families. This changes the site appearance slightly. Getting jekyll-assets, USWDS SCSS font mixins, and custom fonts to work in harmony will take some additional time.

- Updates Ruby to `>=2.7`
- Updates to Jekyll `v4`
- Updates jekyll-assets to latest github commit
- Updates USWDS to `v2.13.x`
- Switches font families to default USWDS (Should we do this?)

:sunglasses:[PREVIEW URL](https://cg-88d42ca6-59d7-47e0-9500-4dd9251360b9.app.cloud.gov/preview/cloud-gov/cg-site/apb/update-jekyll-uswds-deps/)

## Security Considerations
Updates build dependencies to latest


Gemfile.lock
```diff
@@ -1,146 +1,155 @@
+GIT
+  remote: https://github.com/envygeeks/jekyll-assets
+  revision: 056d2c88719ef3b1f90967a606dd1441581dd832
+  specs:
+    jekyll-assets (4.0.0.alpha)
```
bengerman13, line 5: alpha?

apburnes: We have to use the latest possible version to get it to work with the Jekyll 4 updates.

bengerman13:
Looks good to me, except the tree on the docs page is gray now. Not the worst, but it looks kinda funky:
before:
before
after:
after

The font all over is also a little smaller (I think it looks better, but I know I tolerate smaller fonts more than most folks)

I don't think these are show-stoppers, but I think they're worth calling out

bengerman13 (approved):

PR #2155, opened 2022-07-13
Fix jq commands within Elasticsearch backup docs

folksgl:
For both instances of jq usage, the value of the JSON element we
actually want is wrapped in a "credentials" object, so an additional
layer of filtering is needed in the jq command for the returned value to
be anything other than "null".

## Changes proposed in this pull request:
- Update cli commands for Elasticsearch backup process

:sunglasses:[PREVIEW URL](https://cg-88d42ca6-59d7-47e0-9500-4dd9251360b9.app.cloud.gov/preview/cloud-gov/cg-site/correct-es-backup-commands)

## Security Considerations
Backup/Restore process for tenant-brokered services needs to be up-to-date.

PR #2163, opened 2022-08-05
Fix jq commands within Elasticsearch backup docs

markdboyd:
Merge in user contribution from #2155 

## Security considerations

None


rbogle (approved):
Thanks @markdboyd !!

PR #2215, opened 2022-12-07
linking to egress rules

LindsayYoung:
Adding a bread crumb so that people don't waste time wondering why their database won't connect to their app

Add a link to egress rules because, realistically it is part of setting up the database. 

(I don't think this works with forks?)

:sunglasses:[PREVIEW URL](https://cg-88d42ca6-59d7-47e0-9500-4dd9251360b9.app.cloud.gov/preview/cloud-gov/cg-site/LindsayYoung:patch-5)

## Security Considerations
doc update

### DIFF
--- a/_docs/services/relational-database.md
+++ b/_docs/services/relational-database.md
@@ -194,6 +194,8 @@ name                 service   plan                bound apps   last operation
 test-oracle          aws-rds   medium-oracle-se2                create succeeded
 ```
 
+By default, when new spaces are created in your organization an application security group (ASG) is applied that restricts access to only the internal cloud.gov network. You will need to [update egress traffic](/docs/management/space-egress/) rules to allow for your app to reach the database. 
+
 ## Update an instance
 
 To update an existing service instance to a different plan run the following command:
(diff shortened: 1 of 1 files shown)


markdboyd:
Hey @LindsayYoung - I'm seeing that public egress is being applied when creating a new space.

Let me dig deeper to see what is going on though, because that doesn't match our documentation or my expectations.

_docs/services/relational-database.md
```diff
@@ -194,6 +194,8 @@ name                 service   plan                bound apps   last operation
 test-oracle          aws-rds   medium-oracle-se2                create succeeded
 ```
 
+By default, when new spaces are created in your organization an application security group (ASG) is applied that restricts access to only the internal cloud.gov network. You will need to [update egress traffic](/docs/management/space-egress/) rules to allow for your app to reach the database.
```
markdboyd:
```suggestion
By default, when new spaces are created in your organization, an application security group (ASG) is applied that doesn't allow any outgoing traffic. You will need to [update egress traffic]({{ site.baseurl }}{% link _docs/management/space-egress.md %}) rules to allow for your app to reach the database. 
```

@LindsayYoung We updated our settings so that the default ASG actually allows no egress. You could be seeing different behavior in your org if it pre-dates when we made this change.

LindsayYoung: Thanks for the updated text! Sorry I did not see your comment earlier.

markdboyd (approved):

PR #2369, opened 2023-07-05
Vulnerable JS library updates

drewbo:
- Updates to the latest version of mermaid.js (10.2.3) and reworks some of the diagrams for better layout/readability
- Removes netlify-cms

## Security Considerations
Removes libraries with noted or unpatched vulnerabilities


_assets/css/styles.scss
```diff
@@ -437,20 +437,20 @@ code.language-plaintext {
 }
 
 .cg-diagrams-stage {
-  background: #f1f1f1;
+  background: #f1f1f1 !important;
```
markdboyd, line 440: It's been a while since CSS, but is there any reason we **have** to use `!important` for these rules? If the goal is just to make sure this CSS rule is used, we could try increasing the selector specificity here instead. I'm happy to pair on this

pburkholder:
The two mermaid docs I viewed look better:

-  docs/ops/design/secrets/
- docs/ops/design/secrets/

so 👍 from me.

svenaas (commented):
Terraform moved two docs out from under their linked URLs and they've failed the link checker. 

`_docs/ops/dns.md` links to

1. https://www.terraform.io/docs/configuration/resources.html
2. https://www.terraform.io/docs/configuration/syntax.html

These should probably be 

1. https://developer.hashicorp.com/terraform/language/resources

and

2. https://developer.hashicorp.com/terraform/language/syntax 

respectively.

_assets/css/styles.scss
drewbo: The older versions of mermaid had the ability to apply a `null` theme which didn't add CSS rules. This version requires a theme to be set, so now we're setting the theme to `default`. The updated way to customize the theme is to set [theme variables](https://mermaid.js.org/config/theming.html#theme-variables), some of which sets inline styles on elements. We need to use `!important` to override the inline styles. We also had a lot of existing CSS, much of which already used `!important` to override mermaid CSS. Ideally we'd rewrite this all using theme variables in the future but this was a quick fix matching the current style for now

svenaas:
@drewbo I'll push a commit provided I'm not stepping on your toes.

_assets/css/styles.scss
markdboyd: Would this help?

https://mermaid.js.org/config/theming.html#customizing-themes-with-themevariables

drewbo: Yeah that's roughly what I linked but I mostly didn't want to migrate all the legacy CSS into that format

markdboyd (approved):

PR #2430, opened 2023-11-03
Add malicious traffic protections doc

markdboyd:
Closes https://github.com/cloud-gov/private/issues/882

- Add document on the malicious traffic protections on cloud.gov

## Security Considerations

This document outlines the protections against malicious traffic built-in to the platform


_docs/technology/malicious-traffic.md
```diff
@@ -0,0 +1,53 @@
...
+sidenav: true
+title: Protections against malicious traffic
+weight: 30
+---
+
+Like any hosting provider for web applications, cloud.gov is subject to constant malicious traffic from actors seeking to exploit vulnerabilities, to compromise infrastructure, or to deny service altogether.
+
+Furthermore, since cloud.gov is a multi-tenant platform, it experiences a variety of malicious traffic as attackers target specific customers hosted on the platform.
+
+In order to mitigate these ongoing attacks to keep our customers' applications secure and online, cloud.gov includes a number of protections built-in to the platform.
```
jameshochadel: I like this sentence because it focuses on the customer impact. What do you think about re-working the previous two paragraphs to also be from the customer point of view, instead of the platform point of view? We could explain that as a provider, we observe more diverse and frequent attacks than any single application team could see themselves, and because of this, we build strong tools to mitigate them. We could also mention that our mitigations are intended to not only protect individual tenants from attack, but also from being affected by high volumes of traffic going to other customers on the platform.

_docs/technology/malicious-traffic.md
```diff
@@ -0,0 +1,53 @@
...
+- Traffic matching [path traversal attack patterns](https://owasp.org/www-community/attacks/Path_Traversal)
+
+## Protections against traffic surges
+
+Occasionally, cloud.gov sees significant spikes in traffic that may be attempting to overwhelm platform infrastructure via a [DDoS attack](https://www.cloudflare.com/learning/ddos/what-is-a-ddos-attack/) or may simply be a very large scale probing attack.
+
+In order to mitigate the effect of traffic surges on the platform, cloud.gov includes the following rate limits for requests:
+
+- Traffic coming through CloudFront is rate limited with a [CHALLENGE action][challenge action] to **2000 requests** per **forwarded IP address** per 5 minutes
+- Traffic not coming through CloudFront is rate limited with a [CHALLENGE action][challenge action] to **2000 requests** per **source IP address** per 5 minutes
```
pburkholder: Do we have the a note in the relevant Terraform to update this doc if we change our protections?

pburkholder:
I like this. I don't want the perfect to be the enemy of the good, so I think we can publish as-is. If we have capacity, it would be good to match this up with

- traffic flow diagrams (which exist at diagrams.fr.cloud.gov, but should be moved)
- a news post that we now have this documentation.
- perhaps a strong nudge to use CloudFront

_docs/technology/malicious-traffic.md
markdboyd: How about 9f379ca8d0540f24f475d9c8de599321c7b9926e ?

_docs/technology/malicious-traffic.md
markdboyd: We can add that if you think we should, but I'm wary of whether people will notice it even if we add it.

markdboyd:
@pburkholder 

> traffic flow diagrams (which exist at diagrams.fr.cloud.gov, but should be moved)

I'm not sure how this would link up with those. But personally, I also find the diagrams at diagrams.fr.cloud.gov completely unintuitive and not something I'd generally want to surface for customers. If we did need/want some diagram for this article, I'd rather implement them in Mermaid for maintainability and intelligibility

> a news post that we now have this documentation.

I tend to agree with this and it could be a follow-up PR

> perhaps a strong nudge to use CloudFront

We can nudge all we want, but I think the better approach here will be updating the external-domain-service to drop the non-CDN option altogether and giving people no choice

jameshochadel:
FWIW I think the diagrams.fr.cloud.gov diagrams *are* written in Mermaid — but I agree that they're unintuitive, and I think it's because they're either too high-level for the amount of detail they contain, or too detailed for a high-level view of the system. (I think it's the latter.) (But I digress.)

pburkholder:
All my concerns are addressed. @jameshochadel ?

_docs/technology/platform-protections.md
```diff
@@ -0,0 +1,53 @@
...
+sidenav: true
+title: Protections against malicious activity
+weight: 30
+---
+
+As a multi-tenant platform hosting web applications, cloud.gov experiences frequent attacks and ongoing malicious traffic. The goal of these attacks may be to exploit vulnerabilities, to compromise infrastructure, or to deny service altogether.
+
+Given the multitude of applications and application types running on cloud.gov, there is a much greater diversity of attacks experienced by the overall platform than those faced by any single application operating on cloud.gov. Thus, in order to keep all of our customers' applications safe from a variety of attacks, cloud.gov includes multiple layers of defense against different types of attacks.
+
+These protections are not just intended to protect individual tenants from attack, but also to preserve multi-tenancy by preventing large traffic volumes to one customer application affecting other customer applications.
```
jameshochadel: Sorry, I was thinking of something more like this:

```suggestion
Applications and services on the internet experience frequent attacks, probes, and other malicious traffic. Threat actors making malicious requests may aim to exploit vulnerabilities, compromise infrastructure, or deny service to legitimate clients of your applications.

As a multi-tenant platform, cloud.gov observes a diverse and frequent number of attacks. We are constantly improving our defenses to keep your applications online and unaffected by high-traffic-volume attacks against the platform and our customers. To achieve this, cloud.gov includes multiple layers of defense against different types of attacks.
```

Let me know what you think.

jameshochadel (approved):

PR #2431, opened 2023-11-06
fix RDS doc to specify which parameters require an instance reboot

markdboyd:

- see title

## Security Considerations

None just providing clearer, more accurate instructions for our users


cweibel (dismissed):
Adds column about whether a change to the parameter requires a reboot

Changes for email markdown and clarification to ask for assistance for activities requiring a reboot.

Looks good to me

cweibel (approved):

PR #2465, opened 2024-02-16
Create SECURITY.md

JJediny:
-
-
-

## Security Considerations


drewbo:
@JJediny is this required as part of GSA policy or only as part of the TTS Pages ATO? If it's the latter, my understanding was that the cloud.gov site would remain in the cloud.gov boundary/ATO rather than under TTS Pages (e.g. not having to switch github orgs). Happy to take a look and merge if it's required per GSA policy

markdboyd:
Superseded by #2490

PR #2486, opened 2024-04-03
Update identity provider list

mogul:
NSF is no longer registered as an IdP, but many other agencies are.

- Remove NSF
- Add many others

## Security Considerations

None, this is just a documentation update; the list of IdPs is already on the login page.

### DIFF
--- a/_docs/management/leveraging-authentication.md
+++ b/_docs/management/leveraging-authentication.md
@@ -7,11 +7,26 @@ redirect_from:
 title: Leveraging cloud.gov authentication
 ---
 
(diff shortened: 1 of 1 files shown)


mogul:
Note that the directions [here](https://github.com/mogul/cg-site/blob/patch-15/_docs/getting-started/accounts.md) should also remove NSF, and account (hah!) for the other IdPs in the list somehow. I don't know what restrictions there are/aren't, so I didn't try to do that here.

bengerman13 (approved):

markdboyd:
@mogul I'm not sure if we should explicitly list all the integrated providers. The list will always be changing and it seems like it will lead to perpetual documentation rot

mogul:
Up to you, I was just trying to unrot it.

markdboyd:
@mogul I'm going to opt to remove the explicit agency references instead. See #2512. Thanks for bringing this to our attention!

PR #2519, opened 2024-07-03
update article with vite specific instances in the app

Ephraim-G:
- update the current pages KB article with relevant code fences from the updated app and language 

## Security Considerations
N/A


markdboyd (approved):
I don't know anything about vite, but it looks good to me

PR #2593, opened 2025-02-03
BREACH is likely a False Positive

pburkholder:

- BREACH is better considered an FP when application-level mitigations are in place 
- Carry this as an OR is only appropriate if those mitigations are not in place
- CSRF should be in place anyhow for apps exchanging secrets

## Security Considerations
Yes. This better encapsulates the risk from BREACH


wz-gsa (approved):
This is the recommendation as I recall it.

PR #2595, opened 2025-02-14
add global table uswds component + pages documentation

Ephraim-G:
- Add a global `usa-table--borderless` USWDS component which can be referenced throughout the site
- Add Pages documentation

## Security Considerations
N/A

### DIFF
--- a/_includes/content-table.html
+++ b/_includes/content-table.html
@@ -1,0 +1,18 @@
+<table class="usa-table usa-table--borderless">
+  {% if include.caption %}
+  <caption>
+    {{ include.caption }}
+  </caption>
+  {% endif %}
+  <thead>
+    <tr>
+      <th scope="col">{{ include.header1 }}</th>
+      <th scope="col">{{ include.header2 }}</th>
+      <th scope="col">{{ include.header3 }}</th>
+      <th scope="col">{{ include.header4 }}</th>
+    </tr>
+  </thead>
+  <tbody>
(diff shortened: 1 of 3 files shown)


_includes/content-table.html
```diff
@@ -0,0 +1,18 @@
...
+  <thead>
+    <tr>
+      <th scope="col">{{ include.header1 }}</th>
+      <th scope="col">{{ include.header2 }}</th>
+      <th scope="col">{{ include.header3 }}</th>
+      <th scope="col">{{ include.header4 }}</th>
+    </tr>
+  </thead>
+  <tbody>
+    {{ include.content }}
```
sknep, line 16: Not a big deal, but it's kind of a pain to always have to write the table row HTML manually. If you wanted to make it more structured (like if there's always a row title, a description, and an example code block), you might be able to pass in an array through the yaml front matter. I'd defer to Brian on what yall want to maintain going forward

Ephraim-G: I'll confer with him but that may be a solid possibility. I'm all for whatever is easiest for folks to replicate down the line...ultimately we'll convert all existing tables to use the borderless table component

Ephraim-G: @sknep we'll leave it as is for now but implement this in the 11ty version of the site

hursey013 (approved):
This works for now, will revisit on `site` repo 👍

PR #2596, opened 2025-02-26
Update 2021-05-17-troubleshooting-ssh-connections.md

ArsHaider:
Fixed a misspelling

-Fixed a misspelling
-
-

## Security Considerations
None, fixed a misspelling.


pburkholder (dismissed):

pburkholder:
Approved, but if you could also remove references to v6 behavior, that would be a bonus (since ppl are supposed to migrate to support CF clis)

ArsHaider:
> Approved, but if you could also remove references to v6 behavior, that would be a bonus (since ppl are supposed to migrate to support CF clis)

Peter, good call, I removed references to v6, thanks!

pburkholder (approved):

